PR #1, opened 2019-02-22
Add folders for diagrams and initialise ADR tooling

mattmachell:


.circleci/config.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
```
andrewpearce-digital:
```suggestion
version: 2.1
```

PR #3, opened 2019-03-15
UML-46 Local development setup for Viewer

nsmithuk:
Covers:
- Setting up the Docker containers for the Viewer (front) service
- Setup of the initial docker-compose
- A bit of PHP to demo things are working


README.md
```diff
@@ -1,2 +1,10 @@
 # opg-use-my-lpa
 OPG Use My LPA: Managed by opg-org-infra &amp; Terraform
+
+## Setup
+
+``bash
+cd opg-use-my-lpa
+docker-compose up
+docker run --rm --interactive --tty --volume $PWD/service-viewer/app:/app composer install
```
damacus: you might be better off here using something like https://github.com/ministryofjustice/serve-opg/blob/master/docker-compose.yml#L144

Which is simpler to remember :)

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    volumes:
+      - ./service-viewer/web:/web
+    ports:
+      - 9001:80
+    environment:
+      APP_HOST: viewer-app
+      APP_PORT: 9000
+      NGINX_LOG_LEVEL: warn
+
+  viewer-app:
```
damacus: Why not just call it, app and web? (I'm also not apposed to these names!)

README.md
nsmithuk: Thanks Dan. I agree. Definitely on the roadmap.

docker-compose.yml
nsmithuk: Because there will be 3 of each in the end.
- viewer-app & viewer-web
- actor-app & actor-web
- api-app & api-web

service-viewer/docker/web/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Add Waitforit to wait on app starting
+ENV WAITFORIT_VERSION="v2.4.1"
+RUN wget -q -O /usr/local/bin/waitforit https://github.com/maxcnunes/waitforit/releases/download/$WAITFORIT_VERSION/waitforit-linux_amd64 \
+  && chmod +x /usr/local/bin/waitforit
+
+COPY service-viewer/web /web
+COPY service-viewer/docker/web/etc /etc
+
+CMD confd -onetime -backend env \
+  && nginx -g "daemon off;"
```
damacus: Do you actually need confd here yet? If not I'd drop it until you actually need to configure something in the container at runtime. 
It'll make the container a bit easier to reason with between dev and prod.

nsmithuk: This was taken from Serve OPG, where it's used to set the hostname and port for PHP-FPM in the NGINX config. What was the original reasoning for adding it in there?

docker-compose.yml
hansnqyr: this namespacing here is indicative of the wider need to namespace - it might be that 3 folders (viewer, actor, api) with a compose file each is a better fit as this would directly mirror the AWS tasks configurations needed.

nsmithuk: Thanks @hansnqyr. I'll give that a try and see what it looks like.

README.md
nsmithuk: This has been done.

docs/application/decisions/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
+# 1. Record architecture decisions
```
andrewpearce-digital:
```suggestion
# 1. Record application decisions
```

docs/application/decisions/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
...
+
+Date: 2019-02-22
+
+## Status
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made on this project.
```
andrewpearce-digital:
```suggestion
We need to record the application decisions made on this project.
```

PR #4, opened 2019-03-19
UML-49 Added Route53 DNS record.

TomTucka:


terraform/locals.tf
```diff
@@ -0,0 +1,37 @@
+locals {
+  target_accounts = {
+    "development"   = "367815980639"
+    "preproduction" = "888228022356"
+    "production"    = "690083044361"
```
hansnqyr: dropping the quotes here is more idiomatic tf imho and would be consistent with the tags block further down

terraform/credentials.tf
```diff
@@ -0,0 +1,29 @@
...
+    dynamodb_table = "remote_lock"
+  }
+}
+
+provider "aws" {
+  region = "eu-west-2"
+
+  assume_role {
+    role_arn     = "arn:aws:iam::${local.target_account}:role/${var.default_role}"
+    session_name = "terraform-session"
```
hansnqyr: in current versions of terraform using a local here is broken for imports. A variable is fine though and possible preferable, see later comments

terraform/locals.tf
```diff
@@ -0,0 +1,37 @@
...
+    "production"    = "use-an-lpa"
+  }
+
+  dns_prefix = "${lookup(local.dns_prefixes, terraform.workspace)}"
+
+  is_production = {
+    "development"   = "false"
+    "preproduction" = "false"
+    "production"    = "true"
+  }
```
hansnqyr: As a rule I prefer to put env specific settings into a separate, json formatted vars file. This allows for cleaner separation of environment config which can be reused by scripts and other processes. If you use terraform.tfvars it'll get automatically loaded by terraform too.

terraform/locals.tf
```diff
@@ -0,0 +1,37 @@
+locals {
+  target_accounts = {
+    "development"   = "367815980639"
+    "preproduction" = "888228022356"
+    "production"    = "690083044361"
+  }
+
+  target_account = "${lookup(local.target_accounts, terraform.workspace)}"
```
hansnqyr: nothing wrong here, but i'm curious - why target_account and not just account?

terraform/load_balancer.tf
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+
+resource "aws_security_group_rule" "allow_all" {
+  type              = "ingress"
+  from_port         = 443
+  to_port           = 443
+  protocol          = "tcp"
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = "${aws_security_group.loadbalancer.id}"
```
hansnqyr: I suspect this is a copy of something I did :) - the rule should be namespaced to loadbalancer.

TomTucka: renamed the loadbalancer resources to `view_loadbalancer` becuase there will be one for each frontend.

terraform/locals.tf
TomTucka: I mean this is just what @andrewpearce-digital said 😄 renamed to `account` and `accounts`

terraform/locals.tf
TomTucka: done this

terraform/credentials.tf
TomTucka: @andrewpearce-digital said he wants to talk to you about this 👀

terraform/locals.tf
TomTucka: Not sure how to do this... could you show me?

PR #6, opened 2019-03-21
UML-44 build containers in pipeline

andrewpearce-digital:


andrewpearce-digital:
This pipeline is working. However the app itself returns a 200 and an error. Fixing that can be a separate tiket as it's just a scaffold of an app at present.

TomTucka:
I'm Happy with the changes, I would definitely squash the commits though

docker-compose.yml
```diff
@@ -17,6 +18,7 @@ services:
 
   viewer-app:
     container_name: viewer-app
+    image: 311462405659.dkr.ecr.eu-west-1.amazonaws.com/use_my_lpa/view_lpa_front:latest
```
nsmithuk: @andrewpearce-digital Is it easy to change this to `viewer` also?
And what are you thoughts on using `use_an_lpa` everywhere now? Other than the github repo perhaps?

andrewpearce-digital: It's straightforward, naming of github and ecr managed from opg-org-infra.

I think changing the name of an ECR destroys the old and creates a new one, losing the images.
Changing the repo name is an edit.

I want to make another ticket after we have a follow up about naming things, then rename the things.

PR #10, opened 2019-03-23
Multi-Stage Docker build

TomTucka:
Added so that composer can be installed when building the `Dockerfile` in CircleCI.


TomTucka:
This will be merged in as part of @andrewpearce-digital pull request (#6)

PR #15, opened 2019-03-25
Tag all builds with branch and sha sum

andrewpearce-digital:


andrewpearce-digital:
CIRCLECI_SHA1 is not becoming part of the tag. Need to find a way to trim and attach

andrewpearce-digital:
> CIRCLECI_SHA1 is not becoming part of the tag. Need to find a way to trim and attach

Typo, missed the 1 in SHA1

PR #16, opened 2019-03-26
Add C4 main architectural decisions

ramchale:


docs/architecture/decisions/0002-persistent-storage-uses-dynamodb.md
```diff
@@ -0,0 +1,20 @@
...
+
+Date: 2019-02-21
+
+## Status
+
+Accepted
+
+## Context
+
+* he application will require persistent storage for storing LPA ownership, granted access, and possibly user credentials
```
mattmachell: The?

docs/architecture/decisions/0005-automated-testing.md
```diff
@@ -0,0 +1,25 @@
...
+
+Date: 2019-02-21
+
+## Status
+
+Accepted
+
+## Context
+
+The team did not agree on a set of tools to use, but have agreed to research and review options in the coming days. The team have decided to include and implement automated code coverage tests, accessibility tests, and security and penetration testing and have suggested some tools and services to do that.
```
mattmachell: Not sure about this one. If we didn't agree, is it yet an ADR...

docs/architecture/decisions/0006-continuous-delivery.md
```diff
@@ -0,0 +1,21 @@
...
+
+Date: 2019-02-21
+
+## Status
+
+Accepted
+
+## Context
+
+Frequent small releases are preferred strategy to prevent accumulating risk and deliver benefits to users more quickly
```
mattmachell: 👍

docs/architecture/decisions/0005-automated-testing.md
ramchale: Meant to delete that. Just a copy paste of the original note 🤦‍♂️

ramchale: I've removed the extra information. The decision is that we'd test the things rather than picking a tool. Although I think we know now for unit testing and code coverage. Not so much for security and penetration. If it's not worth putting in at this point we can delete or set it to a different status, while we make the other choices

PR #17, opened 2019-04-02
Install behave in python container

andrewpearce-digital:


.circleci/config.yml
```diff
@@ -161,3 +170,13 @@ orbs:
           - run:
               name: check response from viewer service
               command: curl https://viewer.use-an-lpa.opg.service.justice.gov.uk --fail
+      run_behave_suite:
+        executor: python
```
andrewpearce-digital: Needs a checkout. There's nothing in here to run

features/viewer.feature
```diff
@@ -0,0 +1,10 @@
+@web @duckduckgo
```
andrewpearce-digital: Fix these tags

.circleci/config.yml
```diff
@@ -161,3 +170,13 @@ orbs:
           - run:
               name: check response from viewer service
               command: curl https://viewer.use-an-lpa.opg.service.justice.gov.uk --fail
```
andrewpearce-digital: Use jq here to get the right domain and reduce to one test. Use workspace env var

features/steps/viewer.py
```diff
@@ -0,0 +1,29 @@
...
+
+VIEWER_URL = get_viewer_url()
+
+@given('I go to the viewer page on the internet')
+def step_impl(context):
+  context.browser.get(VIEWER_URL)
+ 
+@when('I click Start Now')
+def step_impl(context):
+  start_button= context.browser.find_element_by_css_selector('a.govuk-button')
```
mattmachell: Its worth having some patterns in place for hooking tests off in a consistent way outside of presentational CSS. Rather than a.govuk-button, which will break if you add another button with that class or style it differently later on, it's worth using data attributes to identify things data-action of similar, then using [data-action=""] on the selector.

.circleci/config.yml
```diff
@@ -143,21 +165,38 @@ orbs:
...
+          TF_WORKSPACE: "<<parameters.workspace>>"
         steps:
+          - checkout
           - run:
-              name: check response from viewer service
-              command: curl https://viewer.dev.use-an-lpa.opg.service.justice.gov.uk --fail
-      test_viewer_service_preproduction:
-        executor: terraform
-        steps:
+              command: sudo apt-get install jq -qq
```
nsmithuk: Is `jq` not installed in `circleci/python:3.6-stretch-browsers`? It was in `circleci/python`.

mattmachell:
How are you planning to scaffold fixture data for these tests later on down the line? My concern is that you'll end up having to duplicate work in Python and PHP for data models etc.

nsmithuk:
Do you think we should namespace the `/features` directory to the service? Or perhaps better, create a _features_ directory under the service directory. i.e. `/service-viewer/features`

features/steps/viewer.py
ramchale: Something like context.browser.find_element(By.XPATH, '//a[text()="Start now"]') would be less brittle for now. And we could make the test step generic later on

nsmithuk:
I think we should also start and try to add more documentation comments into `.circleci/config.yml`. Probably not worth doing it retrospectively for this PR, but ideally anything that's new.

Something long the lines on what's done for `test_sirius_access` in the other repo:
https://github.com/ministryofjustice/opg-sirius-api-gateway/blob/master/.circleci/config.yml#L184

features/steps/viewer.py
nsmithuk: How about on the link? (Is that what Matt is saying?)

i.e. Functionally we're expecting a link to `href="/enter-code"`. But we don't overly care what classes it has, or its copy.

andrewpearce-digital:
> Do you think we should namespace the `/features` directory to the service? Or perhaps better, create a _features_ directory under the service directory. i.e. `/service-viewer/features`

I'm not sure. Behave documentation talks about it's opinionated structuring and what it wants to look for.

I was taking an approach of a .feature per "service" (viewer, actor), with scenarios inside for the journey through that service.

It's possible to specify features and tags to run if we want to isolate a group of tests and/or run them in parallel.

Would that acheive what you want without namespacing using paths?

features/steps/viewer.py
andrewpearce-digital: I'd like to do that, I'll need a hand to find the specific thing to look for.

nsmithuk:
> > Do you think we should namespace the `/features` directory to the service? Or perhaps better, create a _features_ directory under the service directory. i.e. `/service-viewer/features`
> 
> I'm not sure. Behave documentation talks about it's opinionated structuring and what it wants to look for.
> 
> I was taking an approach of a .feature per "service" (viewer, actor), with scenarios inside for the journey through that service.
> 
> It's possible to specify features and tags to run if we want to isolate a group of tests and/or run them in parallel.
> 
> Would that acheive what you want without namespacing using paths?

I think I need to understand more what idiomatic tests look like. Happy to defer this for now.

.circleci/config.yml
andrewpearce-digital: I was using the terraform image when I added that. I'll see if it needs removing

andrewpearce-digital: sorted

PR #19, opened 2019-04-04
UML 75 Added session middleware

nsmithuk:


service-viewer/app/src/App/templates/app/enter-code.phtml
```diff
@@ -25,11 +25,9 @@
                             <label class="govuk-label" for="share-code">
                                 LPA code
                             </label>
-                            <input class="govuk-input govuk-input--width-10" id="share-code-one" name="share-code-one"
+                            <input class="govuk-input" id="share-code-one" name="share-code"
```
taylorsj1980: Change id attribute to `share-code` ?

service-viewer/app/src/App/src/Handler/HomePageHandler.php
```diff
@@ -38,65 +36,6 @@ public function __construct(
 
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if ($this->template === null) {
```
taylorsj1980: `$this->containerName` and `$this->router` aren't required anymore. Can we remove them from the constructor and adjust the factory?

nsmithuk: Yes - but I'd expect that to fall under your handlers ticket.

service-viewer/app/src/App/templates/app/enter-code.phtml
nsmithuk: Fixed.

PR #24, opened 2019-04-10
UML-96: Renamed Zend module

nsmithuk:
Refactor Zend Application such that Viewer and Actor are two modules in the same application.


nsmithuk:
> One commit changes 'App' to 'Viewer' and the other renames the apps folder from 'service-viewer' to 'service-front'?
@cooperaj 
Yes, we've changed where the name-spacing happens. It was (going to be) done at the top level. i.e. 
- /service-actor
- /service-viewer

And a Zend app installed under each. We've changed it so there will be a single Zend app, with the names-pacing done at the module level. Thus the above are combined into `/service-front`. 

And then in the modules we'll have `src/Viewer` and `src/Actor`.

We changed once it became apparent how much code the two components would need to share.

cooperaj:
> We changed once it became apparent how much code the two components would need to share.

👍

PR #27, opened 2019-04-16
Added XDebug

nsmithuk:


service-front/docker/app/Dockerfile
```diff
@@ -7,12 +7,17 @@ RUN composer install --prefer-dist --no-suggest --no-interaction --no-scripts --
...
+        && pecl clear-cache \
         && apk del .build-dependencies
 
 COPY service-front/app /app
 COPY --from=composer /app/vendor /app/vendor
 COPY service-front/docker/app/app-php.ini /usr/local/etc/php/conf.d/
 
 WORKDIR /app
+
+CMD ([[ -z "${ENABLE_XDEBUG}" ]] || docker-php-ext-enable xdebug) \
```
cooperaj: Would personally be tempted to break this out into an entrypoint script. This works a script is supposed to be the defined extension point for this type of thing.

PR #37, opened 2019-04-24
Uml 98 show valid lpa details

taylorsj1980:


service-front/app/src/Viewer/templates/app/view-lpa.html.twig
```diff
@@ -0,0 +1,175 @@
+{% extends '@layout/default.html.twig' %}
+
+{% block title %}View LPA{% endblock %}
+
+{% block content %}
+    <style>
```
cooperaj: I'm guessing the inline styles are temporary?

service-front/app/src/Viewer/templates/app/view-lpa.html.twig
```diff
@@ -0,0 +1,175 @@
...
+
+            </div>
+
+            <div class="govuk-grid-row">
+
+                <div class="govuk-grid-column-full">
+
+                    <div class="govuk-panel govuk-panel--confirmation">
+
+                        <h1 class="govuk-panel__title">This LPA is {{  lpa.isValid ? 'valid' : 'invalid' }}</h1>
```
cooperaj: Extra space "{{  lpa.isValid..."

service-front/app/test/ViewerTest/Handler/EnterCodeHandlerTest.php
```diff
@@ -0,0 +1,167 @@
...
+                ->willReturn($returnData);
+        }
+
+        $symfonyFormProphecy->createView()
+            ->willReturn($formProphecy->reveal());
+
+        $formFactoryProphecy = $this->prophesize(FormFactoryInterface::class);
+
+        $formFactoryProphecy->create(ShareCode::class, null, [
+                'csrf_token_manager' => self::CSRF_TOKEN,
```
cooperaj: The token manager is an object retrieved from the request attributes - an instance of TokenManager.

service-front/app/src/Viewer/templates/app/view-lpa.html.twig
taylorsj1980: Yes definitely. I wanted to match the prototype exactly as requested in this ticket so needed to create custom styles but I think there might be some rework to the prototype to make it adopt GDS styles. I'll pick this up with Katie and Darren.

service-front/app/src/Viewer/templates/app/view-lpa.html.twig
taylorsj1980: Will be fixed.

service-front/app/test/ViewerTest/Handler/EnterCodeHandlerTest.php
taylorsj1980: Good spot. Will change.

PR #39, opened 2019-04-29
Write container version into container environment variables using co…

andrewpearce-digital:
…ntainer definition

And make behave test work for localhost to.


andrewpearce-digital:
image
image

PR #42, opened 2019-05-01
ADR Updates

nsmithuk:


docs/architecture/decisions/0008-session-storage-using-an-encrypted-cookie.md
```diff
@@ -0,0 +1,31 @@
...
+To use a cookie who's payload is encrypted with AES GCM. This provides secrecy and authentication.
+
+Not to use JWT, because:
+* To ensure message secrecy, additional libraries are needed.
+* The resulting cookie value is significantly larger.
+* Concerns over the general suitability around using JWT for client side sessions.
+
+## Consequences
+
+* Cookie size will be smaller.
```
mattmachell: Will this still be true with the KMS option?

nsmithuk: Yes, in comparison to JWT with the same key manager.
i.e. Assuming using KMS is a given, the cookie would be smaller with Authenticated Encryption than it would be if we were to use JWT.

PR #44, opened 2019-05-02
UML-95 Moved to KMS for Sessions

nsmithuk:
- [x] Added KMS for local development
- [x] Added KMS based key manager
- [x] Added unit testz
- [x] Added Terraform to create new KMS CMK
- [x] Behave tests still pass

This also beings back the concept of Config objects which are autowired in.

I've removed the Secrets Manager stuff. Rather than strip out all of localstack, only to bring it back when we need DynamoDB, I've left it in and repurposed it ready.


service-front/app/src/Viewer/src/Service/Session/EncryptedCookie.php
```diff
@@ -5,7 +5,7 @@
 namespace Viewer\Service\Session;
 
 use Viewer\Service\Session\KeyManager\KeyNotFoundException;
-use Viewer\Service\Session\KeyManager\Manager;
+use Viewer\Service\Session\KeyManager\KeyManagerInterface;
```
mattmachell: Noticed this (not just this line, much of the work) is all in the Viewer namespace, but it's probably generic enough to be used by both sides of the app. Worth considering how much (if any) reuse it will get.

nsmithuk: @mattmachell Much of this will be put into the `Common` namespace once we start the `Actor` namespace.

service-front/app/test/ViewerTest/Service/Session/KeyManager/KmsManagerTest.php
```diff
@@ -0,0 +1,287 @@
...
+use Viewer\Service\Session\KeyManager\KeyCache;
+use Viewer\Service\Session\KeyManager\Config;
+use Viewer\Service\Session\KeyManager\KmsManager;
+use Viewer\Service\Session\KeyManager\KeyNotFoundException;
+
+class KmsManagerTest extends TestCase
+{
+    const TEST_KMS_CMK_ALIAS = 'test-alias-name';
+
+    private $cacheProphercy;
```
mattmachell: Should this/these be Prophecy? i.e. the Prophecy mock library?

service-front/app/src/Viewer/src/Service/Session/EncryptedCookie.php
```diff
@@ -89,17 +87,15 @@ protected function decode(string $data) : array
...
 
-            $key = $this->keyManager->getKeyId($keyId);
+            $key = $this->keyManager->getDecryptionKey($keyId);
 
             $ciphertext = Base64UrlSafe::decode($payload);
 
             $plaintext = $this->getBlockCipher()
-                ->setKey($key->getKeyMaterial())
-                ->decrypt($ciphertext);
+                            ->setKey($key->getKeyMaterial())
```
cooperaj: Picky me is not a fan of this indentation. It's neither the default (what it was) or aligned at a tab interval or aligned with the arrow of the line above.

cooperaj:
I have a query about the config objects. 

Why are they objects?

Generally an advantage of being able to inject configuration at construction is that it reduces boilerplate code, yet we're adding some in?

service-front/app/src/Viewer/src/Service/Session/EncryptedCookie.php
nsmithuk: This is weird; it doesn't show like this at all in my IDE. I'll give it a poke.

service-front/app/src/Viewer/src/Service/Aws/KmsFactory.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Builds a configured instance of the AWS KMS Client
+ *
+ * Class KmsFactory
+ * @package App\Service\Aws
+ */
+class KmsFactory
+{
+    public function __invoke(ContainerInterface $container)
+    {
```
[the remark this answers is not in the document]
cooperaj: I quite like this idea.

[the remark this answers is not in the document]
nsmithuk: The reasoning behind this is that I want to share the same instance of `Sdk` between all AWS end AWS client. It ensures that we only need to acquire AWS permissions once.

This may not be an issue / other ways around it; but I've just gone with the safe option initially.

PR #45, opened 2019-05-02
Uml 72 zend forms as an alternative

cooperaj:
See what Zend forms could look like.


service-front/app/config/config.php
```diff
@@ -14,6 +14,13 @@
 
 $aggregator = new ConfigAggregator([
     \Zend\Expressive\Session\ConfigProvider::class,
+    \Zend\Expressive\Csrf\ConfigProvider::class,
+
+    \Zend\Form\ConfigProvider::class,
```
nsmithuk: Are we making use of these 5 config imports?

cooperaj: Yes. They're the ones brought in by the CSRF component and the forms themselves.

cooperaj:
> I appreciate that this code stems from a spike rather than a story, so if you'd rather do them as part of a different PR, please can you create a task ticket for them for this or next sprint.

Will get them done now. Shouldn't take too long.

service-front/app/config/config.php
nsmithuk: I can't see where these are used:
(Expressive's install tool can be a bit overzealous when bring things in)
```
    \Zend\Form\ConfigProvider::class,
    \Zend\InputFilter\ConfigProvider::class,
    \Zend\Filter\ConfigProvider::class,
    \Zend\Validator\ConfigProvider::class,
    \Zend\Hydrator\ConfigProvider::class,
```
If I delete these lines locally, everything _seems_ to keep working fine.

nsmithuk: I think it's because we're not putting Form components in via the Container.

cooperaj: Ah. My assumption was that there were interdependencies but you're right that they don't rely on a DI container so I'll get rid.

I guess the question is then why aren't we using the container as they've intended?

nsmithuk: I'm not against it, but I haven't seen any good examples of where forms have used DI with Expressive. i.e. I'm not sure what benefit it would give, whilst I can see the extra layer of complexity it would add.

PR #46, opened 2019-05-03
UML-95: Part 2: Cookie Persistence 

nsmithuk:
A big chunk of the 'complex' stuff in this PR comes from
https://github.com/zendframework/zend-expressive-session-cache

- [x] Refactored concrete Session Persistence class
- [x] Added unit tests
- [x] Added Session Timeout page, and  Middleware to catch it
- [x] Finish wiring up Config object

As part of this I've merged the `CookiePersistence` into `EncryptedCookiePersistence`, as opposed to extending it. This is because on its own `CookiePersistence` would appear to function, but the resulting session would be insecure (un-authenticated). I could have made `CookiePersistence` abstract, but there seems no benefits to doing so.


service-front/app/config/autoload/envs.global.php
```diff
@@ -14,9 +14,67 @@
...
+        // Leave this null to use browser default (current hostname).
+        'cookie_domain' => null,
+
+        // The path prefix of the cookie domain to which it applies.
+        'cookie_path' => '/',
+
+        // Indicates that the cookie should only be transmitted over a
+        // secure HTTPS connection from the client. When set to TRUE, the
+        // cookie will only be set if a secure connection exists.
+        'cookie_secure' => false,
```
mattmachell: Not sure I like false as a default here.

nsmithuk: That's fair. The issue is that we don't have TLS setup for local dev yet, so setting that to `true` breaks everything.

I've created a ticket for next sprint to resolve this. I've included changing this to `true` as part of that, and adding a Behave to ensure the value is always `true`.
https://opgtransform.atlassian.net/browse/UML-125

mattmachell: Ah, if that's all in near future pipeline, then good.

PR #47, opened 2019-05-07
Create Codes dynamodb table

andrewpearce-digital:


terraform_environment/dynamodb.tf
```diff
@@ -0,0 +1,20 @@
+resource "aws_dynamodb_table" "viewer-codes-table" {
```
nsmithuk:
```suggestion
resource "aws_dynamodb_table" "viewer_codes_table" {
```

nsmithuk: We use underscored everywhere else.

PR #65, opened 2019-05-20
Uml 122 service dynamo layers in application

taylorsj1980:


service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -0,0 +1,190 @@
...
+        $this->dynamoDbClient = $dynamoDbClient;
+    }
+
+    /**
+     * Get an LPA using the ID value
+     *
+     * @param int $lpaId
+     * @return array
+     */
+    public function getById(int $lpaId) : array
```
nsmithuk: Please can we make `$lpaId` a string. Although it's likely it'll always be an int, it's not an assumption we need to make.

service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -0,0 +1,190 @@
...
+        throw new NotFoundException('LPA not found');
+    }
+
+    /**
+     * Get an LPA using the share code
+     *
+     * @param int $shareCode
+     * @return array
+     */
+    public function getByCode(int $shareCode) : array
```
nsmithuk: Please can you make `$shareCode` a string. i.e. the code is alphanumeric.

service-api/app/src/App/src/Service/Lpa/LpaService.php
taylorsj1980: Changed

service-api/app/src/App/src/Service/Lpa/LpaService.php
taylorsj1980: Changed

PR #66, opened 2019-05-20
Uml 118 add data model for lpa

cooperaj:
Implement data entities for the Lpa service. 

Provide a factory class to aid with building the entities from Sirius sourced JSON (as per the Swagger docs)


service-front/app/src/Viewer/src/Service/Lpa/LpaFactory.php
```diff
@@ -0,0 +1,124 @@
...
+
+use Viewer\Entity\Lpa;
+use Viewer\Entity\CaseActor;
+use Viewer\Entity\Address;
+use \DateTime;
+use Zend\Stdlib\Exception\InvalidArgumentException;
+
+use function filter_var;
+
+final class LpaFactory
```
nsmithuk: Why make it final?

cooperaj: I don't foresee any reason we'll ever need to subclass it. I've always considered that that being the case you should mark things as final.

Heres a good read https://ocramius.github.io/blog/when-to-declare-classes-final/

cooperaj: As per that article I've clarified the choice by creating an interface for the factory.

nsmithuk: Cool. Approved.

PR #68, opened 2019-05-21
Uml 79 all behave tests to date

taylorsj1980:


features/steps/viewer.py
```diff
@@ -3,17 +3,74 @@
...
+
+@given('the share code input is populated with "{share_code}"')
+def step_impl(context, share_code):
+  share_code_input = context.browser.find_element_by_css_selector('form[name="share_code"] input[name="lpa_code"]')
+  share_code_input.clear()  # Be sure the input is empty
+  share_code_input.send_keys(share_code)
+
+
+# WHENS
+@when('I click the "{button_text}" button')
```
mattmachell: {button_text} labelling and yet the xpath is for a link element.

features/steps/viewer.py
```diff
@@ -3,17 +3,74 @@
...
+def step_impl(context, button_text):
+  button_element = context.browser.find_element_by_xpath('//a[contains(concat(" ", @class, " "), " govuk-button ")][contains(text(), "' + button_text + '")]')
```
mattmachell: Text as a hook isn't great and makes your tests brittle to translation changes.

features/steps/viewer.py
```diff
@@ -3,17 +3,74 @@
...
+  button_element.click()
 
-@then('The enter code page is displayed')
+@when('I click on the "{help_link}" help section')
+def step_impl(context, help_link):
+  help_summary = context.browser.find_element_by_xpath('//summary/span[contains(text(),"' + help_link + '")]')
```
mattmachell: {help_link} and yet no link in the xpath, just summary/span

features/steps/viewer.py
```diff
@@ -3,17 +3,74 @@
...
-  assert context.browser.find_element_by_xpath("//form[@name='share_code']")
\ No newline at end of file
+  share_code_form = context.browser.find_element_by_css_selector('form[name="share_code"]')
+  share_code_form.submit()
+
+
+# THENS
+@then('the "{page_title}" page is displayed')
+def step_impl(context, page_title):
+  page_title_element = context.browser.find_element_by_xpath('//h1[contains(text(),"' + page_title + '")]')
```
mattmachell: Some ambiguity in page_title then h1, rather than title element

features/steps/viewer.py
```diff
@@ -3,17 +3,74 @@
...
+  assert page_title_element.text == page_title
+
+@then('error message "{error_message}" is displayed at the top of the screen')
```
mattmachell: Is this actually testing "at top of page", seems to be just on the page.

features/steps/viewer.py
taylorsj1980: The buttons are actually links with the `govuk-button` class applied

features/steps/viewer.py
taylorsj1980: The help link is styled as a link on the page but is actually just text (in a span) in the summary element. There is no hyperlink element for the help sections.

features/steps/viewer.py
taylorsj1980: I will update the xpath query to look for links with the correct href.

features/steps/viewer.py
taylorsj1980: I will rename to `page_heading`

features/steps/viewer.py
taylorsj1980: It's not strictly speaking the top of the page. I have updated the test step to more accurately convey that the check is for the error summary section near the top of the screen.

features/steps/viewer.py
mattmachell: Yeah, sorry my point was about the semantics of the html and the expectation the when definition sets up. i.e. link elements are not button elements, just because they are styled that way. Style can change independently of semantics, so don't write the test tied to the purely presentational aspects unless that's explicitly what you are testing. It's a good habit for larger test bases with lots of people working on them.

PR #71, opened 2019-05-22
Use the correct API hostname

taylorsj1980:


docker-compose.yml
```diff
@@ -34,6 +34,7 @@ services:
       KMS_SESSION_CMK_ALIAS: 'alias/viewer-sessions-cmk-alias'
 
       # Local only
+      API_SERVICE_HOSTNAME: http://api-web
```
nsmithuk:
```suggestion
      API_SERVICE_HOSTNAME: api-web
```
The AWS variable won't have the protocol, se we need to remove it from here. And we'll need to add it somewhere in the PHP. Perhaps in the `ApiClient\Client`?

taylorsj1980: As discussed, we have moved the protocol to the terraform variable and renamed to URL

PR #76, opened 2019-05-26
Uml 123 api to dynamo connectivity2

taylorsj1980:
- Adds connectivity to DynamoDB
- Coes are looked up via DynamoDB, but actual LPA details are still mocked.
- Viewer handles the responses from API appropriately.
- Includes a new container that seeds dev DynamoDB tables, both locally in AWS (except production).


service-api/app/src/App/src/Service/Lpa/LpaServiceFactory.php
```diff
@@ -0,0 +1,24 @@
...
+use Aws;
+use Psr\Container\ContainerInterface;
+
+class LpaServiceFactory
+{
+    public function __invoke(ContainerInterface $container)
+    {
+        $config = $container->get('config');
+
+        if ( !isset($config['repositories']['dynamodb']['viewer-codes-table'])) {
```
cooperaj: This will fail if either 'dynamodb' or 'repositories' is not set. IIRC.

nsmithuk: Yes, that's the expected behaviour. It need everything, or fail.

nsmithuk: Index warnings are not thrown within an `isset()`; it's a safe way of checking for keys deep within an array.

It's a pattern we use in most factories, on most projects. E.g.
https://github.com/ministryofjustice/opg-refunds/blob/develop/public-front/src/App/src/Service/Session/SessionManagerFactory.php#L17

cooperaj: Guess you really do learn something every day!

terraform_environment/api_ecs.tf
```diff
@@ -93,13 +93,14 @@ resource "aws_security_group_rule" "api_ecs_service_egress" {
...
 // Api ECS Service Task level config
+
 resource "aws_ecs_task_definition" "api" {
   family                   = "${terraform.workspace}-api"
   requires_compatibilities = ["FARGATE"]
   network_mode             = "awsvpc"
   cpu                      = 512
   memory                   = 1024
-  container_definitions    = "[${local.api_web}, ${local.api_app}]"
+  container_definitions    = "[${terraform.workspace == "production" ? local.api_container_definitions_production : local.api_container_definitions }]"
```
andrewpearce-digital: (this comment comes after I've had a look at the aws console, and the other naming comments)

I think we should define a separate ecs service for one-off tasks. If we include a one-off task in the long-running api service, we end up with stopped tasks, that on pre-prod would accumulate.

I think of it in this way, the "one-off "service configures the app with tasks (schema migrations for example), and the api service is the ongoing running off the app.

image

PR #77, opened 2019-05-27
Uml 139 add actor start page

taylorsj1980:
- Added Start Page for Actors.
- Refactored namespace for views.
- Made the layout context aware. i.e. Actor vs Viewer.


service-front/app/src/Viewer/src/Handler/EnterCodeHandler.php
```diff
@@ -39,11 +39,8 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
             }
         }
 
-        return new HtmlResponse(
-            $this->renderer->render(
-                'app::enter-code',
-                [ 'form' => $form ]
-            )
-        );
+        return new HtmlResponse($this->renderer->render('viewer::enter-code',[
```
cooperaj: See, I was always pointed in the direction that if you're going to multiline a piece of code it's either all of it, or none of it. Do we have a coding guidelines on that?

PR #78, opened 2019-05-28
UML 140: Add healthcheck to api and check from viewer

nsmithuk:
And include new OPG Gateway permissions.


nsmithuk:
Will close once environment is deleted.

PR #82, opened 2019-05-29
Allow underscores and pipes in branch names

nsmithuk:
Fixes bug where AWS cannot create resources with underscores in the name.


nsmithuk:
Pipes cause other issues. Leaving for now.

PR #87, opened 2019-06-06
UML 123: Codes in DynamoDB & Seeding from the pipeline

nsmithuk:
- Adds connectivity to DynamoDB
- Codes are looked up via DynamoDB, but actual LPA details are still mocked.
- Viewer handles the responses from API appropriately.
- Includes a new pipeline task that seeds the dev database.


service-api/seeding/dynamodb.py
```diff
@@ -0,0 +1,65 @@
...
+        'ViewerCode': "222222222222",
+        'SiriusId': "22222222222",
+        'Expires': "2019-01-01 12:34:56",
+    }
+)
+
+# Scan and output the table, so we can see what we've got in the logs.
+response = table.scan()
+for i in response['Items']:
+    print(json.dumps(i, indent=4, separators=(',', ': ')))
```
andrewpearce-digital: If this were to accidentally get run on the pipeline against production would this expose customer data? Perhaps we can test that we have specific things we added instead and return true.

.circleci/config.yml
```diff
@@ -34,9 +33,16 @@ workflows:
             branches:
               ignore: master
 
+      - use-my-lpa/seed_database:
+          name: dev_seed_database
+          requires: [dev_apply_environment_terraform]
+          filters:
+            branches:
+              ignore: master
```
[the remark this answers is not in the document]
nsmithuk: The scope of this PR thus far is just seeding dev. Extending it to cover preprod also makes sense. I'm not really sure we ever want fixture data in production.

.circleci/config.yml
```diff
@@ -416,6 +422,31 @@ orbs:
...
+          - attach_workspace:
+              at: /tmp
+          - run:
+              name: Seed DynamoDB
+              command: |
+                sudo pip3 install boto3
+                export AWS_ACCOUNT_ID=$(cat /tmp/cluster_config.json | jq .account_id | xargs)
+                export DYNAMODB_TABLE_VIEWER_CODES=$(cat /tmp/cluster_config.json | jq .viewer_codes_table | xargs)
+                python service-api/seeding/dynamodb.py
+
```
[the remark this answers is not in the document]
nsmithuk: I started from copying one of your scripts, thus using the config file. The issue with that is it unnecessarily tightly couples the script to a complex data type (JSON), and file system access. Keeping config to key/value, either via parameters or env vars, draws a cleaner boundary and (for example) makes the script easier to use from different contexts (i.e. pipeline and local).

service-api/seeding/dynamodb.py
nsmithuk: Yeah; I'm happy to narrow that down.

.circleci/config.yml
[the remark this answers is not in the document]
nsmithuk: I've added preprod. We still need to decide what to do about prod.

service-api/seeding/dynamodb.py
```diff
@@ -0,0 +1,63 @@
...
+]
+
+for i in items:
+    table.put_item(
+       Item=i,
+    )
+    response = table.get_item(
+        Key={'ViewerCode': i['ViewerCode']}
+    )
+    print(json.dumps(response['Item'], indent=4, separators=(',', ': ')))
```
andrewpearce-digital: I thought I made this comment before, but it's not gone through. If we were to run this on production, it would dump out all the data in the table, in our pipeline, in public.

Can we instead test that the data seeded is available and return true for each entry?

.circleci/config.yml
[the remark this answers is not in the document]
nsmithuk: Amended as per call. No seeding runs on production, and the functional test have been replaced with a smoke test.

.circleci/config.yml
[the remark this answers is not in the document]
andrewpearce-digital: 👍

features/view-an-lpa.feature
```diff
@@ -32,12 +32,19 @@ Feature: View a lasting power of attorney
...
     And error message "Enter an LPA share code in the correct format." is displayed next to the LPA code input
 
-  Scenario: Show a try again page when I search for an LPA with an LPA code that does not exist
+  Scenario: Show an error page when I search for an LPA with an LPA code that does not exist
     Given I go to the enter code page on the viewer service
     And the share code input is populated with "1111-1111-1111"
     When the share code form is submitted
     Then the "Invalid Code" page is displayed
 
+  @wip
```
andrewpearce-digital: is this a helper tag used during dev? is it still needed or well described?

nsmithuk: @taylorsj1980 ?

.circleci/config.yml
```diff
@@ -466,3 +504,51 @@ orbs:
...
+          - install_webdriver
+          - attach_workspace:
+              at: /tmp
+          - run:
+              name: Wait for new tasks in services to be running
+              command: |
+                sudo pip3 install boto3
+                python features/aws_ecs_task.py /tmp/cluster_config.json
+          - run:
+              name: Allow CircleCI ingress to environment
```
andrewpearce-digital: allow ingress isn't required if this job (run_heatlhcheck_test) is only running on production

.circleci/config.yml
```diff
@@ -466,3 +504,51 @@ orbs:
...
+              name: Run test
+              command: |
+                viewer_fqdn=$(cat /tmp/cluster_config.json | jq .viewer_fqdn | xargs)
+                viewer_response=$(curl --write-out %{http_code} --silent --output /dev/null https://$viewer_fqdn/healthcheck)
+                [[ $viewer_response == 200 ]] || (echo "Error with viewer health check. HTTP status: ${viewer_response}" && exit 1)
+
+                actor_fqdn=$(cat /tmp/cluster_config.json | jq .actor_fqdn | xargs)
+                # TODO Add for actor when endpoint exists.
+          - run:
+              name: Remove CircleCI ingress to environment
```
andrewpearce-digital: remove ingress isn't required on production

.circleci/config.yml
```diff
@@ -466,3 +504,51 @@ orbs:
...
                 python features/ci_ingress.py /tmp/cluster_config.json
               when: always
+
+
+      run_healthcheck_test:
+        #
+        # Checks the health check is returning 200
+        #
+        executor: python
+        parameters:
```
andrewpearce-digital: we don't need this parameter as the details needed are in the config file

.circleci/config.yml
nsmithuk: I'm not sure I'm confident enough to say this is 'only' production yet.

.circleci/config.yml
nsmithuk: I'm not sure I'm confident enough to say this is 'only' production yet.

.circleci/config.yml
nsmithuk: I'm not sure I'm confident enough to say this is 'only' production yet.

.circleci/config.yml
nsmithuk: Workspace is used. e.g.
```
              name: Allow CircleCI ingress to environment
              command: |
                ENV_NAME=<<parameters.workspace>>
                export TF_WORKSPACE=${ENV_NAME:0:13} >> $BASH_ENV
                echo $TF_WORKSPACE
                if [ "${TF_WORKSPACE}" != "production" ]; then
                  python features/ci_ingress.py /tmp/cluster_config.json --add
                fi
```

.circleci/config.yml
andrewpearce-digital: if we want to run it elsewhere, I'd be inclined to drop the health-check test into the middle of the behave job because ingress is already being opened and closed (saves doing that twice)

PR #92, opened 2019-06-17
Uml 152 sign up form validation errors

taylorsj1980:


service-front/app/src/Actor/src/Form/CreateAccount.php
```diff
@@ -19,57 +24,154 @@ public function __construct(CsrfGuardInterface $csrfGuard)
...
+                        'name'                   => NotEmpty::class,
+                        'break_chain_on_failure' => true,
+                        'options'                => [
+                            'messages'           => [
+                                NotEmpty::IS_EMPTY => 'Confirm your email address',
+                            ],
+                        ],
+                    ],
+                    [
+                        'name'                   => Identical::class,
```
cooperaj: You've put the Identical filter on the email_confirm for email but on the password field for the password. Is this intentional?

service-front/app/src/Common/templates/partials/govuk_form.html.twig
```diff
@@ -0,0 +1,49 @@
...
+
+        <input class="govuk-input {% if element.getMessages() is not empty %}govuk-input--error{% endif %}" id="{{ element.getName() }}"
+               name="{{ element.getName() }}" type="text" value="{{ element.getValue() }}" />
+    </div>
+
+{% endblock form_input_text %}
+
+{% block form_input_password %}
+
+    <div class="govuk-form-group {% if element.getMessages() is not empty %}govuk-form-group--error{% endif %}">
```
cooperaj: Looks good. Could be worth putting a ticket in to DRY this stuff. i.e. the password and text fields are basically the same. Twig supports inheritance type stuff and the Twig Forms code uses it pretty extensively (which we've essentially built here for Zend forms)

service-front/app/src/Actor/src/Form/CreateAccount.php
taylorsj1980: Yeah this is intentional. For the email addresses we want the error to show against the email_confirm input so users know that they have failed to match it.

But the difference on the password fields is that the values will clear when the page is reloaded with the errors. When that happens it looks strange if the now empty password_confirm field is highlighted rather than the first password field. We've done something similar with the LPA and refunds projects too.

cooperaj: Oh yeah. Makes sense

service-front/app/src/Common/templates/partials/govuk_form.html.twig
taylorsj1980: Ok. I'll add this to the error extension along with a similar function for the form group.

PR #94, opened 2019-06-19
Uml 141 connect to gateway

cooperaj:


nsmithuk:
Relevant bits will be moved to https://github.com/ministryofjustice/opg-use-an-lpa/pull/122

PR #95, opened 2019-06-19
Uml 155 account exists error message

taylorsj1980:


service-front/app/src/Actor/src/Handler/CreateAccountHandler.php
```diff
@@ -57,10 +59,17 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
...
+                    $userData = $this->userService->create($data['email'], $data['password']);
 
-                //  TODO - For now just redirect to create account page
-                return $this->redirectToRoute('create-account');
+                    //  TODO - For now just redirect to create account page
+
+                    return $this->redirectToRoute('create-account');
+                } catch (ApiException $ex) {
+                    if ($ex->getCode() == StatusCodeInterface::STATUS_CONFLICT) {
+                        $form->addErrorMessage('email', 'Email address is already registered');
```
mattmachell: So we need to be careful about messages like this on a public service as it gives a method for verifying the existence of users within the system by external parties who aren't that user. This is both a privacy and security concern.

taylorsj1980: Yes I agree, but I think the only alternative here is to appear to accept the new account creation but then not add the details to our database. Then we can email the user to inform them that they already have an account associated with the email address. Does that sound ok?

mattmachell: Yes, that's a pretty standard pattern for this kind of thing, so happy to adopt it here.

taylorsj1980: Ok. I have implemented that method on the latest commit. Thanks

service-front/app/src/Common/src/Service/ApiClient/ClientFactory.php
```diff
@@ -28,7 +28,7 @@ public function __invoke(ContainerInterface $container, RequestedEntry $entityCl
         }
 
         return new Client(
-            $container->get(ClientInterface::class),
```
cooperaj: Interface injection is what we should be using.

service-front/app/composer.json
```diff
@@ -28,11 +28,12 @@
     },
     "require": {
         "php": "^7.1",
+        "alphagov/notifications-php-client": "^1.6",
         "aws/aws-sdk-php": "^3.91",
         "dflydev/fig-cookies": "^2.0",
         "elie29/zend-phpdi-config": "^3.0",
         "paragonie/halite": "^4.5",
-        "php-http/guzzle6-adapter": "^2.0",
```
cooperaj: Is this downgrade a result of bringing in the alphagov library? Is it avoidable?

service-front/app/src/Common/src/Service/Email/EmailClient.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Class EmailClient
+ * @package Common\Service\Email
+ */
+class EmailClient
+{
+    /**
+     * Template IDs for the notify client
+     */
+    const TEMPLATE_ID_EMAIL_ADDRESS_ALREADY_REGISTERED = '4af9acf0-f2c1-4ecc-8441-0e2173890463';
```
cooperaj: Is this likely to change? If so might be worth pulling it from config.

service-front/app/src/Common/src/View/Twig/GovUKZendFormErrorsExtension.php
```diff
@@ -53,18 +55,19 @@ public function errorMessage(Environment $twigEnv, ElementInterface $element) :
...
+    {
         $template = $twigEnv->load(self::THEME_FILE);
 
-        $errors = [];
-        $invalidInput = $form->getInputFilter()->getInvalidInput();
+        $messages = $form->getMessages();
 
-        foreach ($invalidInput as $name => $input) {
-            $errors[$name] = $this->flattenMessages($input->getMessages());
+        //  Flattern each set of messages for each input
```
cooperaj: typo "Flattern"

service-front/app/src/Common/src/Service/Email/EmailClient.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
cooperaj: Needs tests

service-front/app/src/Common/src/Service/Email/EmailClientFactory.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
cooperaj: Needs tests

service-api/app/src/App/src/Exception/ConflictException.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
cooperaj: Needs tests
As part of the work I'll be doing on the login form I'll be hooking up the stuff I did on code coverage and proper test reporting into Circle.

docker-compose.yml
```diff
@@ -73,6 +73,7 @@ services:
       # Always required
       CONTEXT: 'actor'
       KMS_SESSION_CMK_ALIAS: 'alias/viewer-sessions-cmk-alias'
+      NOTIFY_API_KEY: initial_key_nonproduction-9133fd26-4205-4b52-b26a-27de8f5eb7c2-e749db62-33be-4137-9553-da765c4dc5c0
```
andrewpearce-digital: Can we avoid writing a key into the repo, even if it's a non-production key?

We'll need to also put the keys in parameter store or secrets manager in aws and provide them to the containers in ECS.

service-api/app/src/App/src/Exception/ConflictException.php
```diff
@@ -0,0 +1,31 @@
...
+     * ConflictException constructor.
+     *
+     * @param string $message
+     * @param string $title
+     * @param array $additionalData
+     * @param Throwable|null $previous
+     */
+    public function __construct(string $message = null, string $title = 'Conflict', array $additionalData = [], Throwable $previous = null)
+    {
+        parent::__construct($message, $title, $additionalData, $previous);
```
cooperaj: I did some work on AbstractApiException (and all it's dependants) in UML-141. 
The method signature is now ($title, $message, $additional, $previous). It's bad practice to have optional arguments (message) in front of required (title). 
Consequently it's probably better to at least have line 27 replicate that and change this line (29) later.

https://github.com/ministryofjustice/opg-use-an-lpa/blob/33ad6142876aaf239fcd11bd3a9e363a390efb7c/service-api/app/src/App/src/Exception/AbstractApiException.php#L38

service-api/app/src/App/src/Service/User/UserService.php
```diff
@@ -30,6 +32,16 @@ public function __construct(Repository\ActorUsersInterface $usersRepository)
      */
     public function add(array $data) : array
     {
+        //  First try to get any existing user
+        try {
+            $userData = $this->usersRepository->get($data['email']);
+
+            //  There is a user already a user with this email address so throw an exception
+            throw new ConflictException('User already exists with email address ' . $data['email']);
+        } catch (NotFoundException $nfe) {
```
cooperaj: It's the simple way to do it but probably shouldn't rely on exceptions to test for a value. 
I'd suggest adding an 'exists' method to the repository and making it a bool check.

service-front/app/src/Common/src/ConfigProvider.php
```diff
@@ -39,7 +38,8 @@ public function getDependencies() : array
         return [
 
             'aliases' => [
-                Psr\Http\Client\ClientInterface::class => Http\Adapter\Guzzle6\Client::class,
+
+                Http\Client\HttpClient::class => Http\Adapter\Guzzle6\Client::class,
```
cooperaj: HttpClient is a part of the Httplug package which is now deprecated as per it's own readme.

I actually removed all references to it and used ClientInterface in a previous commit. This is the correct PSR interface (and the one that Httplug, in an interim fashion) simply implements in the HttpClient interface.

service-front/app/src/Common/src/Service/ApiClient/Client.php
```diff
@@ -5,8 +5,8 @@
 use GuzzleHttp\Psr7\Uri;
 use GuzzleHttp\Psr7\Request;
 use Http\Client\Exception as HttpException;
+use Http\Client\HttpClient;
 use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Client\ClientInterface;
```
cooperaj: See the ConfigProvider comment.

service-front/app/src/Common/src/Service/ApiClient/ClientFactory.php
cooperaj: Ah, forgot to change this one. I realise now that HTTPClient is an interface. It is however a deprecated one. The original ClientInterface is the correct one. (see comment attached to ConfigProvider).

service-front/app/src/Common/src/Service/Email/EmailClient.php
taylorsj1980: The ID itself is very unlikely to change. The template contents in Notify will change but the ID value will remain the same.

service-front/app/src/Common/src/Service/ApiClient/ClientFactory.php
taylorsj1980: We have a slight problem here in that the Notify client code user a lower version of the Guzzle client where ClientInterface doesn't exist yet. I think we have to use HTTPClient.

service-front/app/composer.json
taylorsj1980: That is correct. The alphagov notify client requires a lower version of Guzzle.

service-front/app/src/Common/src/View/Twig/GovUKZendFormErrorsExtension.php
taylorsj1980: Good spot. Will update.

service-front/app/composer.json
mattmachell: Might be worth creating a pull request for that library? Imagine it'll benefit other government departments.

cooperaj: Was where I was going to go next :) Having looked through the library it looks like it would be a super easy swap.
We can fork it into the MoJ account until it gets accepted. Who has the rights for that?

service-api/app/src/App/src/Exception/ConflictException.php
taylorsj1980: As discussed earlier I think I'm going to remove the setting of the title in the specific exceptions. Taking the class above I can't think of a sensible situation where a ConflictException would be thrown with anything other than a title of 'Conflict'.
That will mean that the changes to the Abstract class that you made can remain without conflict.

service-api/app/src/App/src/Service/User/UserService.php
taylorsj1980: Ok. I'll add an exists function to the repository interface.

service-front/app/src/Common/src/Service/Email/EmailClient.php
taylorsj1980: I have added tests now.

service-front/app/src/Common/src/Service/Email/EmailClientFactory.php
taylorsj1980: I have added tests now.

docker-compose.yml
taylorsj1980: We've found another solution to this using Zend Expressive local files so that the value can be kept out of source control. We can coordinate the sharing of this value manually.

terraform_shared/secretsmanager.tf
```diff
@@ -2,4 +2,8 @@ resource "aws_secretsmanager_secret" "notify_api_key" {
   name = "notify-api-key"
 
   tags   = "${local.default_tags}"
+}
+
```
andrewpearce-digital: because this has moved out of shared, you need another data source to pull in the secret id you're using.
So,

Get the secret id 
use secret id to get secret value


You can insert this above aws_secretsmanager_secret_version
```suggestion
data "aws_secretsmanager_secret" "notify_api_key" {
  name = "notify-api-key"
}

```

terraform_shared/secretsmanager.tf
```diff
@@ -2,4 +2,8 @@ resource "aws_secretsmanager_secret" "notify_api_key" {
...
+data "aws_secretsmanager_secret_version" "notify_api_key" {
+  secret_id = "${aws_secretsmanager_secret.notify_api_key.id}"
 }
```
andrewpearce-digital: should have a new line at the end of the file here

terraform_shared/secretsmanager.tf
```diff
@@ -2,4 +2,8 @@ resource "aws_secretsmanager_secret" "notify_api_key" {
   name = "notify-api-key"
 
   tags   = "${local.default_tags}"
+}
+
+data "aws_secretsmanager_secret_version" "notify_api_key" {
+  secret_id = "${aws_secretsmanager_secret.notify_api_key.id}"
```
andrewpearce-digital: We're now reading this from a data source rather than the resource

```suggestion
  secret_id = "${data.aws_secretsmanager_secret.notify_api_key.id}"
```

docker-compose.yml
andrewpearce-digital: Doesn't that introduce 2 mechanisms for getting the value in? One way for local and another for ECS?

Does the ZE local file write a value into the container's environment variable? If it's not doing that, I would suggest we do things consistently between local and deployed so there are fewer moving parts and places to look if things go wrong.

service-api/app/src/App/src/Exception/ConflictException.php
taylorsj1980: Tests added

docker-compose.yml
taylorsj1980: All our environment variables will end up in the Zend Expressive config files so we can use them in the application code. But the config files are split into global and local counterparts. The contents of the latter are not kept in source control, and supersede the contents of the former.

Essentially just inserting our local copy of the config item into a lower down stage in the same config journey.

terraform_shared/secretsmanager.tf
taylorsj1980: added

service-front/app/src/Common/src/Service/ApiClient/ClientFactory.php
taylorsj1980: As previously discussed I have branched the Notify repo to make the required upgrades for this work, and raised a PR with them to get the changes properly integrated into their code.

This means that the HttpClient interface is no longer used.

service-front/app/src/Common/src/ConfigProvider.php
taylorsj1980: Changes made to utilise the correct psr/http-client package instead.

service-front/app/src/Common/src/Service/ApiClient/Client.php
taylorsj1980: This has been added back in

service-api/app/src/App/src/DataAccess/Repository/ActorUsersInterface.php
```diff
@@ -22,4 +22,12 @@ public function add(string $email, string $password) : array;
...
      */
     public function get(string $email) : array;
+
+    /**
+     * Check for the existence of an actor user
+     *
+     * @param $email
+     * @return bool
+     */
+    public function exists($email) : bool;
```
cooperaj: Type hint the parameters?

service-api/app/src/App/src/Exception/AbstractApiException.php
```diff
@@ -28,12 +28,12 @@ abstract class AbstractApiException extends RuntimeException
...
      * Following the sprint of https://framework.zend.com/blog/2017-03-23-expressive-error-handling.html
      *
-     * @param string $message
      * @param string $title
+     * @param string $message
      * @param array $additionalData
      * @param Throwable|null $previous
      */
-    public function __construct(string $message = null, string $title, array $additionalData = [], Throwable $previous = null)
+    public function __construct(string $title, string $message = null, array $additionalData = [], Throwable $previous = null)
```
cooperaj: For $message to ever be passed a null it'll need to be '?string'

service-api/app/src/App/src/Exception/ConflictException.php
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * ConflictException constructor.
+     *
+     * @param string $message
+     * @param array $additionalData
+     * @param Throwable|null $previous
+     */
+    public function __construct(string $message = null, array $additionalData = [], Throwable $previous = null)
+    {
+        parent::__construct('Conflict', $message, $additionalData, $previous);
```
cooperaj: Do we test the abstract with a null message?

service-api/app/src/App/src/Service/User/UserService.php
```diff
@@ -30,6 +31,10 @@ public function __construct(Repository\ActorUsersInterface $usersRepository)
      */
     public function add(array $data) : array
     {
+        if ($this->usersRepository->exists($data['email'])) {
```
cooperaj: Does the method docbloc detail this \@throw.

service-api/app/src/App/src/Exception/ConflictException.php
cooperaj: Title should be a const?

service-api/app/src/App/src/DataAccess/Repository/ActorUsersInterface.php
taylorsj1980: Done

service-api/app/src/App/src/Exception/AbstractApiException.php
taylorsj1980: done

service-api/app/src/App/src/Service/User/UserService.php
taylorsj1980: Changed

service-api/app/src/App/src/Exception/ConflictException.php
taylorsj1980: I don't think so. There's no need for it to be a constant.

PR #96, opened 2019-06-21
UML-158 Sign into account (Actor)

cooperaj:
As a user (actor)
I want to enter my email and password
So that I can access my account

### DIFF
--- a/service-front/app/src/Actor/src/Handler/LoginPageHandler.php
+++ b/service-front/app/src/Actor/src/Handler/LoginPageHandler.php
@@ -57,10 +57,23 @@ class LoginPageHandler extends AbstractHandler
             if ($form->isValid()) {
                 $data = $form->getData();
 
-                $userData = $this->userService->create($data['email'], $data['password']);
+                // TODO do actual login with something like Zend_Authentication
+                $loggedIn = false;
+                if ($data['email'] === 'test@example.com') {
+                    $loggedIn = true;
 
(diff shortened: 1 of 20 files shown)


service-front/app/src/Actor/src/Handler/LoginPageHandler.php
```diff
@@ -0,0 +1,84 @@
...
+                    return $this->redirectToRoute('home');
+                }
+
+                if ( ! $loggedIn) {
+                    // adding an element name allows the form to link the error message to a field. In this case we'll
+                    // link to the email field to allow the user to correct their mistake.
+                    $form->addErrorMessage(Login::INVALID_LOGIN, 'email');
+
+                    // blank the password field
+                    $form->get('password')->setValue('');
```
nsmithuk: Can we move this line to right before `return new HtmlResponse(...`?
I can't think of any scenario where we'd ever not want that value blank. By putting it right at the end, it removes the risk of it not being call.
For example, isn't the value left in if `$form->isValid() === false` at the moment?

nsmithuk: This being `$form->get('password')->setValue('');`

cooperaj: Ah yes. This was already raised. I think what we should be doing is generally not outputting any passwords at all. Which is probably better done somewhere else. I'll investigate.

PR #98, opened 2019-06-27
Reduce service discovery ttl

andrewpearce-digital:


terraform_environment/api_ecs.tf
```diff
@@ -133,13 +131,14 @@ data "aws_iam_policy_document" "api_permissions_role" {
...
-    resources = ["${aws_dynamodb_table.actor_users_table.arn}","${aws_dynamodb_table.viewer_codes_table.arn}","${aws_dynamodb_table.viewer_activity_table.arn}"]
+    resources = ["${aws_dynamodb_table.actor_users_table.arn}", "${aws_dynamodb_table.viewer_codes_table.arn}", "${aws_dynamodb_table.viewer_activity_table.arn}"]
   }
+
   statement {
     effect = "Allow"
 
     actions = [
-      "execute-api:Invoke"
+      "execute-api:Invoke",
```
mattmachell: Are trailing commas expected in TF?

andrewpearce-digital: They are. I use `terraform fmt` and `tflint` which both have put this in. (it was an unintended consequence of saving the file in this case.

When it's next convenient, I'll have a chat with the team about tool setup to make sure linting and formatting are applied before pushing commits.

PR #104, opened 2019-07-15
UML-189 Add authentication endpoint to the API

cooperaj:


terraform_environment/dynamodb.tf
```diff
@@ -3,10 +3,14 @@ resource "aws_dynamodb_table" "actor_users_table" {
...
-  attribute {
+  attribute = [{
     name = "Email"
     type = "S"
-  }
+  },
+  {
+    name = "Password"
+    type = "S"
+  }]
```
andrewpearce-digital: attributes need to be described in their own blocks I believe rather than a list

See the example here https://www.terraform.io/docs/providers/aws/r/dynamodb_table.html

andrewpearce-digital: attribute {
    name = "Email"
    type = "S"
  }

  attribute {
    name = "Password"
    type = "S"
  }

cooperaj: Was just copying the block below :P But also, I've reverted this now as it's not necessary and I was just misunderstanding things.

service-api/app/src/App/src/DataAccess/Repository/ActorUsersInterface.php
```diff
@@ -28,6 +32,7 @@ public function get(string $email) : array;
      *
      * @param string $email
      * @return bool
+     * @throws NotFoundException
```
taylorsj1980: I don't believe this function should be able to throw a NotFoundException.

If a NotFoundException occurs then this should be caught in the function and result in a false being returned.

service-api/app/src/App/src/DataAccess/Repository/ActorUsersInterface.php
```diff
@@ -12,6 +14,7 @@ interface ActorUsersInterface
      * @param string $email
      * @param string $password
      * @return array
+     * @throws NotFoundException
```
taylorsj1980: Why would the add function throw a NotFoundException? I think this should be a ConflictException.

service-api/app/config/routes.php
```diff
@@ -37,4 +37,5 @@
     $app->get('/v1/lpa/{uid}', App\Handler\LpaHandler::class, 'lpa.by.uid');
     $app->get('/v1/lpa-by-code/{shareCode}', App\Handler\LpaHandler::class, 'lpa.by.share-code');
     $app->route('/v1/user', App\Handler\UserHandler::class, ['GET', 'POST'], 'user.get');
+    $app->get('/v1/auth', App\Handler\AuthHandler::class, 'user.auth');
```
taylorsj1980: I would expect this to be a POST or PATCH rather than a get request, since the action will probably result in a change being made to the DB. i.e. a token value being added.

service-api/app/src/App/src/Service/User/UserService.php
```diff
@@ -44,10 +51,33 @@ public function add(array $data) : array
...
+
+    /**
+     * Attempts authentication of a user
+     *
+     * @param string $email
+     * @param string $password
+     * @return array
+     * @throws NotFoundException
+     */
+    public function authenticate(string $email, string $password) : array
```
taylorsj1980: Please add this function to `DataAccess\Repository\ActorUsersInterface`

service-api/app/src/App/src/Service/User/UserService.php
```diff
@@ -66,4 +74,27 @@ public function activate(string $activationToken) : array
...
         return $this->usersRepository->activate($activationToken);
     }
+
+    /**
+     * Attempts authentication of a user
+     *
+     * @param string $email
+     * @param string $password
+     * @return array
+     * @throws NotFoundException
```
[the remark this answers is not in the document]
cooperaj: As discussed I believe there is benefit to exposing this type of exception out of the API. We may not choose to do anything special with it on the frontend mind, but it gives us the option to.

PR #107, opened 2019-07-23
Uml 190 store user in session

cooperaj:
Depends on 189.

### DIFF
--- a/service-api/app/src/App/src/DataAccess/Repository/ActorUsersInterface.php
+++ b/service-api/app/src/App/src/DataAccess/Repository/ActorUsersInterface.php
@@ -45,7 +45,6 @@ interface ActorUsersInterface
      *
      * @param string $email
      * @return bool
-     * @throws NotFoundException
      */
     public function exists(string $email) : bool;
 }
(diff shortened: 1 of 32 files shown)


service-api/app/src/App/src/DataAccess/Repository/ActorUsersInterface.php
```diff
@@ -45,6 +45,7 @@ public function activate(string $activationToken) : array;
      *
      * @param string $email
      * @return bool
+     * @throws NotFoundException
```
taylorsj1980: This exception will not be thrown from here. It's caught in the function and handled.

service-front/app/src/Common/src/Service/ApiClient/Client.php
```diff
@@ -69,6 +69,10 @@ public function httpGet(string $path, array $query = []) : ?array
             switch ($response->getStatusCode()) {
                 case 200:
                     return $this->handleResponse($response);
+                case 401:
```
taylorsj1980: 4xx codes shouldn't be handled here. They should fall to the `default` block so that an exception can be created and thrown.

PR #110, opened 2019-07-26
UML-169 - create PR status based on code coverage

andrewpearce-digital:
Moves upload of test results to codecov to a single job that runs in parallel to terraform applies
Some uploads have taken longer so this aims to reduce overall workflow time.


.circleci/config.yml
```diff
@@ -193,9 +198,14 @@ orbs:
...
               path: build/service-front/
-          - run:
-              name: Upload to Codecov
-              command: bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
+          - persist_to_workspace:
+              root: build
+              paths:
+                - service-front/
+              when: always
+          # - run:
```
cooperaj: Can probably ditch this

.circleci/config.yml
```diff
@@ -286,9 +296,14 @@ orbs:
...
               path: build/service-api/
-          - run:
-              name: Upload to Codecov
-              command: bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
+          - persist_to_workspace:
+              root: build
+              paths:
+                - service-api/
+              when: always
+          # - run:
```
cooperaj: and this

codecov.yml
```diff
@@ -0,0 +1,31 @@
+coverage:
```
cooperaj: We need to have a bit of a think about the setting for this

cooperaj: CodeCov provides two views of your testing metrics.

cooperaj: A project level view, and this should be fairly obvious for most. It defines what the coverage is for the project as a whole. We might decide we want something sensible like 80% coverage in the end, but sadly we're not their yet so will have to bump up to that number. For any given PR we want our  project coverage to stay the same, or go up.

cooperaj: A patch level view. This took me a while to figure. This is the percentage of code in your PR that is covered by tests. Us, being good developers should want this as close to 100% as we can get it. i.e. all things we do are tested. Had this been in from the beginning we wouldn't have an issue meeting the 80% requirement for the project.

But this makes the refactoring of older, untested, code a pain as it means you *need* to write tests for that code as well as make whatever changes you need. It's good practise to do it anyway...

andrewpearce-digital: Set this for project
target: auto
threshold: 5%

will allow a 5% drop on the current master value.

.circleci/config.yml
andrewpearce-digital: done

.circleci/config.yml
andrewpearce-digital: done

codecov.yml
andrewpearce-digital: I've added this for each service on the project

PR #111, opened 2019-08-07
Uml 178 field validation

taylorsj1980:


service-front/app/src/Actor/src/Handler/CreateAccountHandler.php
```diff
@@ -25,6 +24,8 @@
  */
 class CreateAccountHandler extends AbstractHandler
 {
+    use CsrfGuard;
```
cooperaj: don't forget to add the CsrfGuardAware interface.

service-front/app/src/Viewer/src/Handler/EnterCodeHandler.php
```diff
@@ -19,6 +18,7 @@
  */
 class EnterCodeHandler extends AbstractHandler
 {
+    use CsrfGuard;
```
cooperaj: interface

service-front/app/src/Actor/src/Handler/CreateAccountHandler.php
taylorsj1980: Good spot. Will add. Thanks

service-front/app/src/Viewer/src/Handler/EnterCodeHandler.php
taylorsj1980: Good spot. Will add. Thanks

PR #113, opened 2019-08-12
Uml 193 logout

cooperaj:
Depends on UML-201 #112 

Add logout handler and functionality around sessions.

### DIFF
--- a/service-front/app/src/Actor/src/Handler/LogoutPageHandler.php
+++ b/service-front/app/src/Actor/src/Handler/LogoutPageHandler.php
@@ -25,18 +25,6 @@ class LogoutPageHandler extends AbstractHandler implements SessionAware
     use Session;
 
     /**
-     * CreateAccountHandler constructor.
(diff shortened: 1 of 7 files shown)


service-front/app/src/Actor/src/Handler/LogoutPageHandler.php
```diff
@@ -0,0 +1,51 @@
...
+class LogoutPageHandler extends AbstractHandler implements SessionAware
+{
+    use Session;
+
+    /**
+     * CreateAccountHandler constructor.
+     * @param TemplateRendererInterface $renderer
+     * @param UrlHelper $urlHelper
+     */
+    public function __construct(
```
taylorsj1980: I think this constructor can be removed. The AbstractHandler with deal with this.

cooperaj: ah yes. will fix

PR #115, opened 2019-08-27
Uml 197 add lpa cross validation

taylorsj1980:


service-api/app/test/AppTest/DataAccess/DynamoDb/ActorUsersTest.php
```diff
@@ -15,26 +16,369 @@
...
+    const TABLE_NAME = 'test-table-name';
+
+    private $dynamoDbClientProphecy;
+
+    protected function setUp()
+    {
+        $this->dynamoDbClientProphecy = $this->prophesize(DynamoDbClient::class);
+    }
+
+    public function testAdd()
```
cooperaj: Probably worth sticking to a single test naming format per file. 

/** @test */
this_is_a_test

vs

TestThisIsATest

service-api/app/test/AppTest/DataAccess/DynamoDb/GenerateAwsResultTrait.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace AppTest\DataAccess\DynamoDb;
+
+use Aws\Result;
+use DateTime;
+
+trait GenerateAwsResultTrait
```
cooperaj: There is a private function I created in a test (ill dig out where) that does this using plain iterators. Namely because dependant on the method called (get vs update) you'll get back different array keys (Item vs Attributes)

cooperaj: I only mention it because this is a nice solution but that test (and this trait) will need rework to be used together.

cooperaj: https://github.com/ministryofjustice/opg-use-an-lpa/blob/master/service-api/app/test/AppTest/DataAccess/DynamoDb/ActorUsersTest.php

taylorsj1980: Changed the test files to use the new create AWS result in a trait

PR #124, opened 2019-10-07
Content changes & updated tests

GemTay:


service-front/app/test/ViewerTest/Handler/CheckCodeHandlerTest.php
```diff
@@ -7,6 +7,7 @@
 use Common\Exception\ApiException;
 use Common\Middleware\Session\SessionTimeoutException;
 use Common\Service\Lpa\LpaService;
+use phpDocumentor\Reflection\Types\String_;
```
cooperaj: Don't suppose you know where this import came from?

PR #130, opened 2019-10-11
Content changes & added todays date functionality

GemTay:


service-front/app/src/Common/src/View/Twig/LpaExtension.php
```diff
@@ -89,11 +89,16 @@ private function filterData(iterable $data, array $filterFields)
     /**
      * @param string|null $date
      * @return string
+     * @throws \Exception
      */
     public function lpaDate(?string $date)
     {
         if (!empty($date)) {
-            $date = DateTime::createFromFormat('Y-m-d', $date);
+            if ($date == "today"){
```
cooperaj: I'm not quite sure we should be checking "today" or any variant. I'm guessing it's due to the usage of `The information on this LPA summary was correct on {{ lpa_date("today") }}` below. I think we may need to hash out a different way of doing this.

service-front/app/src/Viewer/templates/viewer/view-lpa.html.twig
```diff
@@ -4,203 +4,264 @@
 
 {% block content %}
     <style>
-        .logo {
```
cooperaj: Now that all these have been removed what has replaced them?

service-front/app/src/Viewer/templates/viewer/view-lpa.html.twig
```diff
@@ -4,203 +4,264 @@
...
-                        <p class="govuk-body-l">Office of the<br/>Public Guardian</p>
-                    </div>
+                    <span class="govuk-caption-l" style="margin-top: 0px">
+                        {% if lpa.caseSubtype == "pfa" %}
+                            Property and Finance
+                        {% else %}
+                            Health and Welfare
+                        {% endif %}
+                    </span>
+                    <h1 class="govuk-heading-l" style="color:#005ea5">Nicholas Brown</h1>
```
cooperaj: Should this name use `{{lpa.donor.firstname|surname}}` like the code replaced

service-front/app/src/Viewer/templates/viewer/view-lpa.html.twig
```diff
@@ -4,203 +4,264 @@
...
+                    <span class="govuk-caption-l" style="margin-top: 0px"></span>
+                    <h1 class="govuk-heading-l" style="margin-bottom: 0px; color: #005ea5"></h1>
+                </div>
 
                 <div class="govuk-grid-column-full">
 
                     {% if lpa.status == 'Cancelled' %}
 
-                        <div class="govuk-panel govuk-panel--failed">
+                        <div class="govuk-panel govuk-panel--failed" style="margin-bottom: 15px; padding: 35px; border: 5px">
```
cooperaj: Inline styles aren't the best. We'll need to liaise with John now that new build stuff is in place and figure out something that involves scss or similar.

PR #131, opened 2019-10-14
WIP: Uml 273 web webpack

johnnolan:
Added webpack into the web project. This now pulls in the JS, SCSS and assets from the govuk-frontend npm package instead of the copied content from github.

We can now update the npm package in package.json to bump the version we use.

I have also updated the packages to the latest 3.3.0.

We cannot merge this until we implement a build step in docker to `npm i` the web folder and build the assets out. I will be talking to Andrew today to hopefully sort this.

Once a PR in the MOJ Frontend is approved we will be able to npm install that library as well.

To test/use go to the `service-front/web` folder, run `npm i`.

After this has installed all dependencies, run `npm run build` and this will fetch the latest assets and put the files accordingly into the dist folder to be copied out via Docker.


service-front/web/src/main.scss
```diff
@@ -0,0 +1 @@
+@import "../node_modules/govuk-frontend/govuk/all";
```
cooperaj: Do node_modules import paths work here?

i.e. `@import "~govuk-frontend/govuk/all";

Might tidy things up a little.

johnnolan: Good catch, you are right, a lot nice! Just pushed the change

service-front/web/webpack.config.js
```diff
@@ -0,0 +1,67 @@
+// Webpack uses this to work with directories
```
cooperaj: Is this file indented with 2 spaces or 4?

Might be worth getting an editorconfig in to fix at one or the other, maybe a jslint configuration?

johnnolan: Have aligned the code correctly now

PR #132, opened 2019-10-14
Uml 199 move to id for users

cooperaj:
In order to get dashboards working we need to attach individual lpas to actor accounts. It is not possible to do this with email so this PR ensures unique id's for users instead. We'll be able to use them for assigning LPA's to accounts.

### DIFF
--- a/service-api/app/src/App/src/DataAccess/DynamoDb/ActorUsers.php
+++ b/service-api/app/src/App/src/DataAccess/DynamoDb/ActorUsers.php
@@ -38,11 +38,12 @@ class ActorUsers implements ActorUsersInterface
     /**
      * @inheritDoc
      */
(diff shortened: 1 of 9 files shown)


gregtyler:
With the caveat that I've not worked with this application or DynamoDB before:

Change seems sensible, and cleanly implemented. Tests are clear to read (I'm starting to see why Alex raves about Prophercy).

My only concern would be whether you can change the hash key of a DynamoDB table on the fly? I'd suspect it might need downtime and/or a full recreate cycle.

cooperaj:
> My only concern would be whether you can change the hash key of a DynamoDB table on the fly? I'd suspect it might need downtime and/or a full recreate cycle.

It's early days at this point with no "live" data. It would be a concern later, but no right now.

PR #133, opened 2019-10-15
Configured Dockerfile to build front end assets

johnnolan:
This branch will build out the assets as defined in the `src` directory in `web` on a production build. Locally you run a watch command which will automatically build the assets for you.

I have updated the readme with instructions of the extra step required to run the webpack command.

Ideally I wanted Docker when run locally to automatically do this for us but I don't have the knowledge one whether this would be possible at the moment.

### DIFF
--- a/service-front/web/package.json
+++ b/service-front/web/package.json
@@ -5,7 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "build": "webpack -p --progress --config webpack.config.js"
+    "build": "webpack -p --progress --config webpack.config.js",
+    "watch": "webpack -p --progress --watch --config webpack.config.js"
   },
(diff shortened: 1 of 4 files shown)


service-front/web/package.json
```diff
@@ -5,7 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "build": "webpack -p --progress --config webpack.config.js"
+    "build": "webpack -p --progress --config webpack.config.js",
+    "watch": "webpack -p --progress --watch --config webpack.config.js"
   },
   "author": "John Nolan <john.nolan@digital.justice.gov.uk>",
   "license": "ISC",
```
cooperaj: I think our repos are MIT licensed.

carinadigital:
LGTM :shipit:

PR #134, opened 2019-10-15
Uml 214 send appropriate user id to api

cooperaj:
Ensure that we're passing through appropriate user context when attempting some actions that require it.


cooperaj:
Currently working on tests.

PR #139, opened 2019-10-17
Uml 283 updates to viewer and tests

GemTay:


service-front/app/src/Viewer/templates/viewer/home-page.html.twig
```diff
@@ -24,10 +24,13 @@
 
                 <p class="govuk-body">To use this service, you’ll need an ‘LPA access code’ from the donor or an attorney. Codes expire 30 days after they are made.</p>
 
-                <div class="govuk-inset-text">Use a different service if you want to <a href="#" class="govuk-link">make a code to use your own LPA</a> .</div>
+                <div class="govuk-inset-text">Use a different service if you want to <a href="localhost:9002" class="govuk-link">make a code to use your own LPA</a> .</div>
```
johnnolan: Can this URL just be `/`?

service-front/app/src/Viewer/templates/viewer/view-lpa.html.twig
```diff
@@ -37,7 +37,7 @@
                             Health and Welfare
                         {% endif %}
                     </span>
-                    <h1 class="govuk-heading-l" style="color:#005ea5">{{ lpa.donor.firstname }} {{ lpa.donor.surname }}</h1>
+                    <h1 class="govuk-heading-l">{{ lpa.donor.firstname }} {{ lpa.donor.surname }}</h1>
```
johnnolan: This should be sorted when you merge the latest master branch, it now has its own class to style it properly

service-front/app/src/Viewer/templates/viewer/home-page.html.twig
cooperaj: It is a link to the actor side of the app from the viewer. They will be on different domains at some point and will be driven by configuration.

service-front/app/src/Viewer/templates/viewer/check-code-found.html.twig
```diff
@@ -6,7 +6,7 @@
     <div class="govuk-width-container">
         {{ include('@partials/new-service.html.twig') }}
 
-        <a href="/" class="govuk-back-link">Back</a>
+        <a href="/enter-code" class="govuk-back-link">Back</a>
```
cooperaj: You should be able to use a route twig function here. Something like `{{ route('enter-code') }}` (assuming the route is a named on in the routes file).

PR #144, opened 2019-10-29
Uml 264 creation of actor codes from lpa uids

cooperaj:
A big and horrible single class cli command function that can create actor codes for lpa's that are passed in.

Takes a comma separated list of lpa uids and spits out a bunch of JSON telling you of codes created for actors. 

It *does not* create any codes for actors that exist already. But *may* create codes for other 'new' actors on an LPA.

### DIFF
--- a/scripts/generate_actor_codes.py
+++ b/scripts/generate_actor_codes.py
@@ -5,20 +5,21 @@ import os
 import sys
 
 class CodeGeneration:
-    lpauids = ''
     aws_account_id = ''
     aws_iam_session = ''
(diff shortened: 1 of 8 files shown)


cooperaj:
For local dev spin up the environment as normal e.g. `docker-compose -f ... up`. When running the command can be run in the api container.

Something like 

```
$ docker-compose -f ... -f ... -f ... run api-app php console.php actorcode:create 700000000047,700000000138,700000000252,700000000344,700000000435
```

Should do it.

PR #145, opened 2019-10-30
Uml 238 actor view lpa summary

GemTay:


service-front/app/src/Actor/src/Handler/ViewLpaSummaryHandler.php
```diff
@@ -0,0 +1,73 @@
...
+        $actorLpaToken = $request->getQueryParams()['lpa'];
+
+        if (is_null($actorLpaToken)) {
+            throw new InvalidRequestException('No actor-lpa token specified');
+        }
+
+        $user = $this->getUser($request);
+        $identity = (!is_null($user)) ? $user->getIdentity() : null;
+
+        $lpa = $this->lpaService->getLpaById($identity, $actorLpaToken);
```
mattmachell: Does this service call not throw an exception if an LPA is not found?

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -54,27 +55,18 @@ public function getLpas(string $userToken) : ?ArrayObject
...
-     * @return ArrayObject|null
+     * @return Lpa|null
      * @throws Exception
      */
-    public function getLpaById(string $userToken, string $actorLpaToken) : ?ArrayObject
+    public function getLpaById(string $userToken, string $actorLpaToken) : ?Lpa
     {
         $this->apiClient->setUserTokenHeader($userToken);
 
         $lpaData = $this->apiClient->httpGet('/v1/lpas/' . $actorLpaToken);
```
mattmachell: Looks like there should be some logic here to handle main HTTP result options other than a happy 200 OK path here. What if I get a 504 or 500 from the call? Client seems to throw an APIException and that should be handled.

service-front/app/src/Actor/templates/actor/view-lpa-summary.html.twig
```diff
@@ -0,0 +1,222 @@
...
+            </div>
+        </div>
+
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-full">
+                <h1 class="govuk-heading-xl">LPA summary</h1>
+            </div>
+        </div>
+
+        <div class="govuk-grid-row" style="border: 5px solid #006435; margin: 0px 0px 30px 0px; padding: 5px">
```
mattmachell: Why the use of inline styles? Would be better to extract to a local (S)CSS file if they cover customisations for the service. Maintaining inline styles is a pain.

cooperaj:
I'm quite confused by all my commits being a part of this PRs history. These have already been merged into master so I think some sort of merging/rebasing wonkiness has gone on.

service-front/app/src/Actor/src/Handler/ViewLpaSummaryHandler.php
cooperaj: It does, but to get to this handler the user will have clicked a link that used this particular LPA's code. If an error occurs at this point we would want to see an application error on the frontend and (when logging is in place) log it somewhere appropriate.

service-front/app/src/Common/src/Service/Lpa/LpaService.php
cooperaj: I think this probably does need visiting but as a general rule the APIExceptions encapsulate some business responses that get dealt with at a higher level i.e. showing a not found message (404). 

In cases of 5xx errors we'll want to be showing an Oops page to the user and logging it in cloudwatch (not yet in place)

service-front/app/src/Actor/src/Handler/ViewLpaSummaryHandler.php
mattmachell: Yeah, my thinking was the next line where $lpa is checked for null then throws and exception seems out of place if the exception handling for the API call is correct.

cooperaj: That... is a good point.

service-front/app/src/Actor/src/Handler/ViewLpaSummaryHandler.php
```diff
@@ -0,0 +1,74 @@
...
+
+        if ($lpa == null){
+            throw new Exception('Lpa not found');
```
cooperaj: $lpa shouldn't be null unless an API exception is thrown so this **shouldn't** be necessary

service-front/app/src/Actor/templates/actor/partials/lpa-details.html.twig
```diff
@@ -10,9 +10,9 @@
         </dt>
         <dd class="govuk-summary-list__value govuk-!-padding-top-4 govuk-!-padding-bottom-4 govuk-!-margin-bottom-0">
             <p class="govuk-!-margin-top-0">
-                <a href="{{ path('lpa.create-code', {}, {'lpa': actorToken }) }}">Give an organisation access</a><br>
-                <a href="{{ path('lpa.create-code', {}, {'lpa': actorToken }) }}">Check access codes</a><br>
-                <a href="{{ path('lpa.create-code', {}, {'lpa': actorToken }) }}">View LPA summary</a>
```
GemTay: @cooperaj @SeemaMenon I've had to rename the actor token from 'lpa' to 'lpa-actor-token' as it clashed with the 'lpa' (actual lpa object) passed to the renderer in the handler - if you could both use this from now on too :)

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -54,27 +55,18 @@ public function getLpas(string $userToken) : ?ArrayObject
...
 
-        if (is_array($lpaData)) {
-            $lpaData = $this->parseLpaData($lpaData);
-        }
-
-        return $lpaData;
+        return isset($lpaData['lpa']) ? $this->lpaFactory->createLpaFromData($lpaData['lpa']) : null;
```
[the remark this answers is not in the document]
GemTay: @cooperaj Haven't made changes here as I was a bit unsure, could this be something to revisit at a later stage?

service-front/app/src/Actor/templates/actor/partials/lpa-details.html.twig
cooperaj: The 'lpa' value is merely what the query string parameter gets named in the generated link and should not affect any rendering of anything.

service-front/app/src/Actor/templates/actor/partials/lpa-sub-navigation.html.twig
```diff
@@ -16,7 +16,7 @@
                     </a>
                 </li>
                 <li class="moj-sub-navigation__item">
-                    <a class="moj-sub-navigation__link"  href="/use-lpa//lpa/lpa-detail">
+                    <a class="moj-sub-navigation__link" {% if path() == '/lpa/view-lpa' %} aria-current="page" {% endif %} href="{{ path('lpa.view', {}, {'lpa-actor-token': actorLpaToken }) }}">
```
cooperaj: The url is subject to change at this point so it might be better to do 

`path() == path('lpa.view-lpa')`

(Assuming it's actually 'lpa.view-lpa'. Thats from memory

service-front/app/src/Actor/templates/actor/view-lpa-summary.html.twig
```diff
@@ -0,0 +1,216 @@
...
+                        {% endif %}
+                    {% endif %}
+                </p>
+            </div>
+        </div>
+
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <h2 class="govuk-heading-l">The donor</h2>
+                <dl class="govuk-summary-list govuk-summary-list--no-border" style="margin-bottom: 30px">
```
cooperaj: There are classes that can be used to do simple margin/padding changes. Something like 

`govuk-!-margin-bottom-6`

(They work in multiples of 5 I think?)

service-front/app/src/Common/src/Service/Lpa/LpaService.php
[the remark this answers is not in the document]
cooperaj: Yep. As long as it works.

PR #146, opened 2019-10-30
Uml 237 show access code

SeemaMenon:


service-front/app/test/ActorTest/Handler/CreateViewerCodeHandlerTest.php
```diff
@@ -20,12 +20,16 @@
...
 use Zend\Expressive\Template\TemplateRendererInterface;
+use Common\Entity\User;
 
 class CreateViewerCodeHandlerTest extends TestCase
 {
-    const CSRF_CODE      = '123456';
+    const CSRF_CODE = '123456';
     const IDENTITY_TOKEN = '01234567-01234-01234-01234-012345678901';
-    const LPA_ID         = '01234567-01234-01234-01234-012345678901';
+    const LPA_ID = '01234567-01234-01234-01234-012345678901';
```
cooperaj: There are a number of consts here that are identical in intended function. Namely IDENTITY_TOKEN == IDENTITY and LPA_ID == LPA_TOKEN. Would be best to refactor these out to remove duplicates.

service-front/app/test/ActorTest/Handler/CreateViewerCodeHandlerTest.php
```diff
@@ -73,7 +90,7 @@ public function setUp()
         $this->requestProphecy->getAttribute(CsrfMiddleware::GUARD_ATTRIBUTE)
             ->willReturn($csrfProphecy->reveal());
 
-        $this->rendererProphecy->render('actor::lpa-create-viewercode', Argument::that(function($options) {
+        $this->rendererProphecy->render('actor::lpa-create-viewercode', Argument::that(function ($options) {
```
cooperaj: Since this 'view' is specific to a particular test it's probably better to set this prophecy expectation in that test.

service-front/app/test/ActorTest/Handler/CreateViewerCodeHandlerTest.php
SeemaMenon: DONE

service-front/app/test/ActorTest/Handler/CreateViewerCodeHandlerTest.php
SeemaMenon: AMENDED

PR #147, opened 2019-10-31
fix condition for certificate on preprod

andrewpearce-digital:


terraform/account/locals.tf
```diff
@@ -7,7 +7,7 @@ locals {
   account = var.accounts[local.account_name]
 
   dns_namespace_acc = terraform.workspace == "production" ? "" : "${local.account_name}."
-  dev_wildcard      = local.account_name != "development" ? "" : "*."
+  dev_wildcard      = local.account_name == "production" ? "" : "*."
```
carinadigital: should `dev_wildcard` be rename `certificate_prefix` ?

PR #149, opened 2019-11-01
refactor: Improved compilation speed of SCSS in Docker

johnnolan:
Added ignore files for docker, removed package-lock. Split custom and npm scss into seperate calls.

20 seconds down to 120ms compilation


service-front/web/.dockerignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
+dist/
+package-lock.json
```
cooperaj: My understanding of the lock file is that it allows us reproducible builds in a CI environment. Why would we exclude it?

johnnolan: The web project is only ever built within Docker on local and CI. The contents copied in and npm installed from there.

As it is ran in the Docker container and not outside of it, in my mind I can't think of a reason to keep track as it should never change.

However! Please let me know otherwise if you think there is a reason to track in. It may be I am doing things the wrong way.

cooperaj: IIRC correctly one of the best practice ways to ensure you're doing NPM build in docker right is to;

copy in package.json and package-lock.json,
run npm i
copy in rest of build requirements.
run build.

This ensures you've got your reproducible builds and caching of dependencies is kept. 

The danger with not using a lock file is that any particular build in CI my fail due to upstream changes.

johnnolan: You are dead on, right, leave it with me, I am going to have another step through the flow and see how it can be tracked nicely, will let you know how I get on

johnnolan: I have brought it back in, from no on people will have to commit this should it change.

PR #154, opened 2019-11-05
Uml 257 check existing access codes

GemTay:


cooperaj:
Mmm. I think this one is going to need some thought.

There are a number of architecture issues that will need changing on this one.

service-front/web/src/index.js
```diff
@@ -2,3 +2,5 @@ import './main.scss';
 import { initAll } from 'govuk-frontend';
 
 initAll();
+
+document.body.classList.add('js-enabled');
```
mattmachell: If you're adding this to handle a set of style toggles on pageload, you should consider adding it to the base template as an inline script, otherwise you'll get flash of un-styled content issues.

service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
```diff
@@ -0,0 +1,94 @@
...
+        $identity = (!is_null($user)) ? $user->getIdentity() : null;
+
+        $lpa = $this->lpaService->getLpaById($identity, $actorLpaToken);
+
+        $shareCodes = $this->viewerCodeService->getShareCodes(
+            $identity,
+            $actorLpaToken
+        );
+
+        foreach($shareCodes as $key => $code){
```
cooperaj: Until we get some automatic formating in place you'll need to ensure spacing  around your ifs/elses/foreachs is consistant.

Personally I'd go with 

`if ($this) {`
`} elseif ($that) {}`

or 

`foreach ($this as $that) {`

service-front/app/src/Actor/templates/actor/check-access-codes.html.twig
```diff
@@ -0,0 +1,79 @@
...
+
+                    {% if check_if_code_has_expired(code.Expires|date("Y/m/d", false)) == true %}
+                        {% set allActive = false %}
+                    {% endif %}
+                    {% endfor %}
+
+                    <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-default">
+
+                        {% for code in shareCodes %}
+                            {% if check_if_code_has_expired(code.Expires|date("Y/m/d", false)) == false %}
```
cooperaj: This is a neat solution though I'd prefer to see the `code.Expires` to be passed in as is to the check function. Also simplify the name and remove the boolean check since it's not necessary.

e.g. `{% if code_has_expired(code.Expires) %} .. {% else %} ... {% endif %}`

you can also (I think) use a 'not' keyword in twig so 

`{% if not code_has_expired(code.Expires) %} .. {% else %} ... {% endif %}`

(best double check that though.

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -77,12 +77,18 @@ public function getLpaById(string $userToken, string $actorLpaToken) : ?Lpa
...
      * @throws Exception
      */
-    public function getLpaByCode(string $shareCode, string $donorSurname) : ?ArrayObject
+    public function getLpaByCode(string $shareCode, string $donorSurname, bool $track) : ?ArrayObject
     {
         //  Filter dashes out of the share code
         $shareCode = str_replace('-', '', $shareCode);
 
-        $lpaData = $this->apiClient->httpPost('/v1/viewer-codes/summary', [
+        if($track){
```
cooperaj: Consistency on spacing please. We'll want to follow PSR-12 ideally (https://www.php-fig.org/psr/psr-12/)

service-front/app/src/Common/src/View/Twig/LpaExtension.php
```diff
@@ -119,6 +120,30 @@ public function daysRemaining(?string $expiryDate) : string
...
+     * @param string|null $expiryDate
+     * @return bool
+     * @throws \Exception
+     */
+    public function hasCodeExpired(?string $expiryDate) : bool
+    {
+        $hasExpired = false;
+
+        if (!empty($expiryDate)) {
+            $expires = new DateTime($expiryDate);
```
cooperaj: It's possible to directly compare DateTime objects (see https://www.php.net/manual/en/datetime.diff.php#example-2787)

service-front/app/src/Actor/src/Handler/LpaDashboardHandler.php
```diff
@@ -51,6 +66,33 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             return new RedirectResponse($this->urlHelper->generate('lpa.add'));
         }
 
+        foreach ($lpas as $lpaKey => $lpaData) {
```
cooperaj: I think this handle function is going to need refactoring. It seems like there is a lot of conditionals and loops that would be nice to avoid if possible.

service-front/app/src/Actor/src/Handler/LpaDashboardHandler.php
```diff
@@ -51,6 +66,33 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+            $shareCodes = $this->viewerCodeService->getShareCodes(
+                $identity,
+                $actorToken
+            );
+
+            if (!empty($shareCodes[0])) {
+                foreach ($shareCodes as $codeKey => $code) {
+                    $expiryDate = new DateTime($code['Expires']);
+                    $expiryDate = $expiryDate->format('Y-m-d');
+                    if ($this->lpaExtension->hasCodeExpired($expiryDate)){
```
cooperaj: The lpaExtension is a twig view only type of thing and should not be used in a handler (which is only meant to martial the data needed to make a response and give that to a view).  Perhaps this logic needs moving to the viewerCodeService? You could add a boolean parameter to the getShareCodes function, maybe named `$withActiveCount` that would direct that.

service-front/app/src/Common/src/View/Twig/LpaExtension.php
cooperaj: This code can be reduced to 

```
if ($date = new DateTime($expiryDate)) {
    return $date < (new DateTime('now'));
}

return null;
```

(I think, thats from memory)

service-front/app/test/ActorTest/Handler/CheckAccessCodesHandlerTest.php
```diff
@@ -127,7 +127,7 @@ public function check_access_codes_page_is_displayed()
             ->willReturn($lpa);
 
         $this->viewerCodeServiceProphecy
-            ->getShareCodes(self::IDENTITY_TOKEN,self::LPA_ID)
+            ->getShareCodes(self::IDENTITY_TOKEN,self::LPA_ID, false)
```
cooperaj: Just a formatting tweak - parameters in a function should be separated by a space.

service-front/app/src/Common/src/Service/Lpa/ViewerCodeService.php
```diff
@@ -63,6 +65,33 @@ public function getShareCodes(string $userToken, string $lpaId) :?ArrayObject
...
 
+        if ($withActiveCount) {
+            $shareCodes = $this->getNumberOfActiveCodes($shareCodes);
+        }
+
         return $shareCodes;
     }
+
+
+    public function getNumberOfActiveCodes(ArrayObject $shareCodes) :?ArrayObject
```
cooperaj: Unless this is likely to be used elsewhere making it private is probably better.

service-front/app/src/Actor/src/Handler/LpaDashboardHandler.php
```diff
@@ -68,29 +65,15 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
-                    } else {
-                        $counter += 1;
-                        $lpas[$lpaKey]['activeCodeCount'] = $counter;
-                    }
-                }
-            } else {
-                $lpas[$lpaKey]['activeCodeCount'] = $counter;
-            }
-
+            $lpas[$lpaKey]['activeCodeCount'] = $shareCodes['activeCodeCount'];
```
cooperaj: This is really nice refactoring

service-front/app/src/Actor/src/Handler/LpaDashboardHandler.php
```diff
@@ -65,16 +65,16 @@ public function handle(ServerRequestInterface $request): ResponseInterface
 
         foreach ($lpas as $lpaKey => $lpaData) {
 
-            $actorToken = $lpaData['user-lpa-actor-token'];
```
cooperaj: This indentation change doesn't look right?

service-front/app/src/Actor/templates/actor/partials/lpa-details.html.twig
```diff
@@ -16,13 +16,17 @@
...
-            <span class="moj-badge moj-badge--grey">
-                {% if activeCodeCount == 1 %}
-                {{ activeCodeCount }} active code
-                {% else %}
-                {{ activeCodeCount }} active codes
-                {% endif %}
-            </span>
+            {% if activeCodeCount >= 1 %}
+                <span class="moj-badge moj-badge--green">
+                    {% if activeCodeCount == 1 %}
```
cooperaj: I *think* twig offers some sort of pluralisation feature. Maybe.

If you can't find it immediately just resolve this :)

service-front/app/src/Actor/templates/actor/partials/lpa-details.html.twig
```diff
@@ -16,13 +16,17 @@
...
+                        {{ activeCodeCount }} active code
+                    {% else %}
+                        {{ activeCodeCount }} active codes
+                    {% endif %}
+                </span>
+            {% else %}
+                <span class="moj-badge moj-badge--grey">{{ activeCodeCount }} active codes</span>
```
cooperaj: May as well just output '0 active codes' here.

service-front/app/src/Common/src/View/Twig/LpaExtension.php
```diff
@@ -155,7 +155,7 @@ public function daysRemaining(?string $expiryDate) : string
     public function hasCodeExpired(?string $expiryDate) : ?bool
     {
         if (!empty($expiryDate && $date = new DateTime($expiryDate))) {
-            return $date < (new DateTime('now'));
+            return $date <= (new DateTime('now'));
```
cooperaj: This may need the time setting to the end of the day as well, as you did in the viewercodeservice?

service-front/app/test/CommonTest/Service/Lpa/ViewerCodeServiceTest.php
```diff
@@ -101,9 +101,11 @@ public function returns_empty_array_if_no_share_codes_generated()
     public function gets_number_of_active_codes_for_lpa()
     {
         $lpaId = '98765432-01234-01234-01234-012345678902';
-        $now =  (new DateTime('now'))->format('Y-m-d');
-        $futureWeek = (new DateTime('+ 1 week'))->format('Y-m-d');
-        $pastWeek = (new DateTime('- 1 week'))->format('Y-m-d');
+
+        $endOfToday =  (new DateTime('now'))->setTime(23,59,59)->format('Y-m-d H:i:s');
```
cooperaj: Why are the formats being set on the datetime objects?

The comparisons below won't work on strings (well they will, but that can be chalked up to PHP things) and the formats will make a string. Would you not want the comparison to be on the DateTimes themselves?

service-front/app/src/Actor/templates/actor/check-access-codes.html.twig
```diff
@@ -0,0 +1,79 @@
...
+                    <h2 class="govuk-heading-m">Active codes</h2>
+
+                    <p class="govuk-body">Give an organisation their access code so they can view this LPA.<br>
+                        They should go to www.gov.uk/view-an-lpa to use the code.</p>
+
+                    {% set allActive = true %}
+
+                    {% for code in shareCodes %}
+
+                    {% if check_if_code_has_expired(code.Expires|date("Y/m/d", false)) == true %}
```
cooperaj: You shouldn't need to ever compare a function return to a boolean. This is because the function return is already a boolean and would make the conditional check work.

cooperaj: Also, not that this will matter so much, you'll want to always use a triple equals (===) in a conditional with PHP. PHP does some weird and wonderful things with automatic type casting if you only use two equals.

service-front/app/test/CommonTest/Service/Lpa/ViewerCodeServiceTest.php
GemTay: The expiry dates are stored as strings and the getNumberOfActiveCodes function takes the string and converts it into a DateTime object to do the comparison - without the formats the test fails

cooperaj: Oops. Serves me right for not reading properly :).

I'd double check the format then. We use (or should be) DATE_ATOM or 'c' (they're the same) for storage as it encodes timezones and such.

GemTay: Ahh okay, I've changed the formats to 'c' to neaten them up :)

cooperaj:
👍 LGTM

PR #155, opened 2019-11-06
Uml 312 unit tests

johnnolan:
Added 100% coverage to new PDF Service. Split app.js into 2 files to make it easier to test and maintain.

### DIFF
--- a/pdf/app/package-lock.json
+++ b/pdf/app/package-lock.json
@@ -1765,6 +1765,12 @@
         }
       }
     },
+    "cookiejar": {
+      "version": "2.1.2",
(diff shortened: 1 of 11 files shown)


mattmachell:
> I have a query about how this will scale though, as our current LPA online setup uses amazon SQS queuing and workers. How does this service handle load?

We had a discussion about this. Basically it's down to a difference in scale, in that this will be doing single simple pages so should be faster. There's a further ticket to look at load testing and container optimisation that John took away, to make sure we don't hit any bottlenecks with the way this uses chrome to render.

johnnolan:
Thanks @cooperaj I will create a ticket regarding `service-pdf` rename.

PR #158, opened 2019-11-06
UML-258 : Added the feedback url

SeemaMenon:


cooperaj:
This is now conflicting and will need some work before we can merge.

PR #169, opened 2019-11-15
Add in container diagrams C4puml.

mattmachell:


cooperaj:
Having just encountered this as a part of my integration testing the viewer and actor applications also access KMS in order to encrypt cookies.

mattmachell:
Oh, good catch! Will revise.

PR #172, opened 2019-11-21
Behat testing

cooperaj:
This implements a working behat setup for acceptance/integration testing.

It currently features a set of integration testing that tests service level components and an acceptance suite that accesses the application as a user would.


service-front/app/composer.json
```diff
@@ -56,12 +56,14 @@
         "zendframework/zend-stdlib": "^3.1"
     },
     "require-dev": {
+        "behat/behat": "^3.5",
+        "cooperaj/behat-psr-extension": "dev-master",
```
mattmachell: We tend to be a little wary of making our projects dependent on people's personal github projects. Probably worth bringing this into an MOJ repo.

service-front/app/behat.yml
```diff
@@ -0,0 +1,28 @@
...
+  gherkin:
+    filters:
+      tags: '~@wip'
+
+  extensions:
+    Acpr\Behat\Psr\ServiceContainer\Extension:
+      application: "%paths.base%/features/bootstrap/app.php"
+      container: "%paths.base%/features/bootstrap/container.php"
+    Behat\MinkExtension:
+      base_url: 'http://localhost'
```
andrewpearce-digital: Assuming the suite would be able to run against deployed infra, it would be good to have a way to override the base_url so that we can inject a deployed environment's url.

cooperaj: This configuration is for whitebox testing of the service and so won't be tested like that. A different config, running in a separate behat instance will be needed for end to end infra testing.

service-front/app/features/context/Integration/AccountContext.php
```diff
@@ -0,0 +1,208 @@
...
+        assertContains($expectedEmail, $requestBody);
+    }
+
+    /**
+     * @Then /^I receive unique instructions on how to reset my password$/
+     */
+    public function iReceiveUniqueInstructionsOnHowToResetMyPassword()
+    {
+        $expectedEmail = 'test@example.com';
+        $expectedUrl = 'http://localhost/forgot-password/1234567890';
```
andrewpearce-digital: if we use a variable for the base_url here we could run this test against otheer envs.

cooperaj: It's possible this might work in some cases but I don't think in this one would. 

This context file is for whitebox integration testing the service layers of service-front. Not requests ever leave the test environment. 

The acceptance feature file could possibly do it but is setup around mocking out outgoing requests and therefore doesn't actually hit infrastructure. It could be possible I guess to write code within the context to handle those two environments but the approach I was going to take was just to have a different context file for smoke-testing.

service-front/app/composer.json
cooperaj: We're now running off a fork of this in ministryofjustice using a repo overlay

PR #177, opened 2019-11-29
Uml 377 - Add the missing unit tests

SeemaMenon:


service-front/app/test/ActorTest/Handler/CheckAccessCodesHandlerTest.php
```diff
@@ -139,4 +142,154 @@ public function check_access_codes_page_is_displayed()
...
+                'actorToken' => self::LPA_ID,
+                'user' => self::IDENTITY_TOKEN,
+                'lpa' => $lpa,
+                'shareCodes' => $shareCodes
+            ])
+            ->willReturn('');
+
+        $response = $handler->handle($this->requestProphecy->reveal());
+
+        foreach ($shareCodes as $key => $code) {
```
cooperaj: I'm not sure these checks are necessary. The system under test (SUT) is the CheckAccessCodesHandler not the lpa objects - which you created as a part of the test setup.

service-front/app/test/ActorTest/Handler/CheckAccessCodesHandlerTest.php
```diff
@@ -139,4 +142,154 @@ public function check_access_codes_page_is_displayed()
...
```
cooperaj: Same as previous comment.

PR #184, opened 2019-12-06
Move healthcheck and factory to common folder

GemTay:


service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -40,18 +50,23 @@ public function __construct(string $version, ApiClient $api)
...
+            ],
+            "healthy" => $this->isHealthy()
         ]);
     }
 
     protected function isHealthy() : bool
     {
-        // TODO actual checks that verify service health
-        return true;
+        if ($this->checkDynamoEndpoint()['healthy'] && $this->checkApiEndpoint()['healthy'])
```
cooperaj: This can be simplified right down to

`return ($this->checkDynamoEndpoint()['healthy'] && $this->checkApiEndpoint()['healthy']);`

service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -61,15 +76,42 @@ protected function checkApiEndpoint() : array
...
+        $data['response_time'] = round(microtime(true) - $start, 3);
+
+        return $data;
+    }
+
+    protected function checkDynamoEndpoint() : array
+    {
+        $data = [];
+
+        $tableNames = ["ActorCodes", "ActorUsers", "UserLpaActorMap", "ViewerActivity", "ViewerCodes"];
```
cooperaj: Checking for so many tables makes the test more brittle (if we rename or something this test will break). It's probably better to just check for the 1 table - assuming this is just a connection check and not a "is my database as expected" check.

Better yet, just check that the returned TableNames contains at least 1 entry.

service-front/app/src/Common/src/Handler/HealthcheckHandler.php
```diff
@@ -43,24 +43,31 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
             "healthy" => $this->isHealthy(),
             "version" => $this->version,
             "dependencies" => [
-                "api" => $this->checkApiEndpoint()
+                "apiAndDynamo" => $this->checkApiEndpointAndDynamo()
```
cooperaj: From the point of view of the front end code all we're doing is calling the app-api. We don't really care that dynamo is involved so I don't think this rename is needed.

The data that embeds here, i.e. the api healthcheck response, will mention the dynamo stuff (which it does care about)

service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -61,15 +71,40 @@ protected function checkApiEndpoint() : array
...
+    protected function checkDynamoEndpoint() : array
+    {
+        $data = [];
+
+        $start = microtime(true);
+
+        try {
+            $dbTables = $this->actorCodes->get('XXXXXXXXXXXX');
+
+            if (is_null($dbTables["TableNames"])) {
```
cooperaj: Is this correct? An actor code query returns a "TableNames"?

service-api/app/test/AppTest/Handler/HealthcheckHandlerTest.php
```diff
@@ -4,88 +4,62 @@
...
+        $this->actorCodesProphecy = $this->prophesize(ActorCodesInterface::class);
+        $this->lpaInterface = $this->prophesize(LpasInterface::class);
     }
 
-    /**
-     * @dataProvider responseDataProvider
-     */
-    public function testReturnsExpectedJsonResponse(int $status, string $response)
+    /** @test */
+    public function testReturnsExpectedJsonResponse()
```
cooperaj: `/** @test */` and the "test" function name prefix don't both need to be specified. Pick one :)

PR #185, opened 2019-12-10
Add LPA decorator to the API service

cooperaj:
Filters empty actors from the response from the API Gateway before passing on the LPA to  the caller.

### DIFF
--- a/service-api/app/src/App/src/Service/Lpa/LpaService.php
+++ b/service-api/app/src/App/src/Service/Lpa/LpaService.php
@@ -33,17 +33,24 @@ class LpaService
      */
     private $userLpaActorMapRepository;
 
+    /**
+     * @var LpaFilterFactory
+     */
+    private $lpaFilterFactory;
+
     public function __construct(
         Repository\ViewerCodesInterface $viewerCodesRepository,
         Repository\ViewerCodeActivityInterface $viewerCodeActivityRepository,
         Repository\LpasInterface $lpaRepository,
-        Repository\UserLpaActorMapInterface $userLpaActorMapRepository
+        Repository\UserLpaActorMapInterface $userLpaActorMapRepository,
+        LpaFilterFactory $lpaFilterFactory
     )
     {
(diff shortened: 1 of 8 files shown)


service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -54,7 +61,7 @@ public function __construct(
      */
     public function getByUid(string $uid) : ?Repository\Response\LpaInterface
     {
-        return $this->lpaRepository->get($uid);
+        return ($this->lpaFilterFactory)($this->lpaRepository->get($uid));
```
acsauk-mojd: Nice - I don't tend to use `__invoke` very often but makes things succinct. I'm assuming the parentheses around the factory are required to make it invokable?

service-api/app/test/AppTest/Service/Lpa/LpaDataCleanseDecoratorTest.php
```diff
@@ -0,0 +1,48 @@
...
+                        'surname' => ''
+                    ]
+                ]
+            ]);
+
+        $decoratedLpaInterface = new LpaDataCleanseDecorator($lpaResponseProphecy->reveal());
+        $lpa = $decoratedLpaInterface->getData();
+
+        $this->assertCount(0, $lpa['attorneys']);
+    }
```
acsauk-mojd: Could potentially add a further two elements to the array or write a new test with attorneys that contain a first and/or last name to cover all paths.

service-api/app/src/App/src/Service/Lpa/LpaService.php
cooperaj: Correct

service-api/app/test/AppTest/Service/Lpa/LpaDataCleanseDecoratorTest.php
cooperaj: I was literally just thinking that. Will do.

PR #188, opened 2019-12-18
Password reset coverage

cooperaj:
Some small test tweaks to see how that affects the code coverage report coming out of coveralls


cooperaj:
This PR removes/invalidates the password reset handlers for coverage. This is because they're 100% (or should be) glue code that ties framework (which we don't want to test) to services (which we're already testing).

The types of test that should touch these files are behat driven and hit these files via the framework in the acceptance/ui tests.

PR #190, opened 2020-01-02
UML 253: removed validation on viewer access code and allows lower ca…

SeemaMenon:
…se or upper case formats now


service-front/app/src/Viewer/src/Form/ShareCode.php
```diff
@@ -37,14 +37,6 @@ public function getInputFilterSpecification() : array
                 'filters'  => [
                     ['name' => StringTrim::class],
                 ],
-                'validators' => [
```
cooperaj: This will need amending to cope with the combinations we're expecting rather than removing entirely.

features/view-an-lpa.feature
```diff
@@ -20,12 +20,12 @@ Feature: View a lasting power of attorney
     And error message "Value is required and can't be empty" is displayed next to the LPA access code input
     And error message "Value is required and can't be empty" is displayed next to the Donor surname input
 
-  Scenario: Show an error message when I search for an LPA with an LPA code that is the wrong format
-    Given I go to the enter code page on the viewer service
-    And the share code input is populated with "wrong-format" and "donor-surname"
-    When the share code form is submitted
-    Then error message "Enter an LPA share code in the correct format." is displayed in the error summary
-    And error message "Enter an LPA share code in the correct format." is displayed next to the LPA access code input
+#  Scenario: Show an error message when I search for an LPA with an LPA code that is the wrong format
```
cooperaj: If you don't want a scenario to run tag it with '@wip' rather than commenting it out.

SeemaMenon: Now that the validator is put back, this test will be needed. So not removing it.

features/view-an-lpa.feature
```diff
@@ -22,32 +22,32 @@ Feature: View a lasting power of attorney
 
   Scenario: Show an error message when I search for an LPA with an LPA code that is the wrong format
     Given I go to the enter code page on the viewer service
-    And the share code input is populated with "wrong-format" and "donor-surname"
+    And the share code input is populated with "2DBM&ADB2U2F6" and "Sanderson"
```
cooperaj: This is looking really nice but I recommend you take a look at the `Scenario Outline` test type at the bottom of the password-reset.feature file in service front. This is a better way to test a list of values for results. It might not completely apply in this case but it could tidy things up.

features/view-an-lpa.feature
```diff
@@ -22,32 +22,32 @@ Feature: View a lasting power of attorney
...
+    And the share code input is populated with "111111111111" and "test-surname"
     When the share code form is submitted
     Then the "We could not find an LPA matching those details" page is displayed
 
   Scenario: Show an error page when I search for an LPA with an LPA code that has expired
     Given I go to the enter code page on the viewer service
-    And the share code input is populated with "N4KB-EBEZ-MNJF" and "Sanderson"
+    And the share code input is populated with "n4kb-ebez-mnjf" and "Sanderson"
     When the share code form is submitted
     Then the "Expired code" page is displayed
```
cooperaj: Rather than describing a "page is displayed" a more generic way to say this would be "Then the user is told their code has expired." This will help for future refactors.

features/view-an-lpa.feature
```diff
@@ -22,32 +22,32 @@ Feature: View a lasting power of attorney
...
+    And the share code input is populated with "P9H8 A6ML D3AM" and "Sanderson"
     When the share code form is submitted
     Then the "Is this the LPA you want to view?" page is displayed
 
   Scenario: Show the wrong details help comment on confirmation code page when I click the link
     Given I go to the enter code page on the viewer service
-    And the share code input is populated with "P9H8-A6ML-D3AM" and "Sanderson"
+    And the share code input is populated with "p9h8 a6ml d3am" and "Sanderson"
     When the share code form is submitted
     Then the "Is this the LPA you want to view?" page is displayed
```
cooperaj: "Then the user is asked to confirm the LPA is as expected"

features/view-an-lpa.feature
```diff
@@ -22,32 +22,32 @@ Feature: View a lasting power of attorney
...
     Given the "If you need to access this LPA after" help section is not visible
```
cooperaj: This should not be a GIven (or maybe shouldn't be there at all?)

service-front/app/src/Viewer/src/Form/ShareCode.php
```diff
@@ -39,7 +39,7 @@ public function getInputFilterSpecification() : array
                 ],
                 'validators' => [
                     new Regex([
-                        'pattern' => '/[\w\d]{4,4}-[\w\d]{4,4}-[\w\d]{4,4}/',
+                        'pattern' => '/[\w\d\s\]{4,4}[\S\s?\-][\w\d\s]{4,4}[\S\s?\-][\w\d\s]{4,4}/',
```
cooperaj: This pattern won't work in all cases (and I think is slightly broken).

I've played around with it and come up with `[\w\d\s]{4,4}(\s|-)?[\w\d\s]{4,4}(\s|-)?[\w\d\s]{4,4}`. You can test it out here: https://regex101.com/r/dSzfs4/1

features/view-an-lpa.feature
SeemaMenon: I get what you say on this, but this being the existing Behave tests...I was applying the fix to get the tests working.New ticket to cover the Behat scenarios for the viewer side may be?

features/view-an-lpa.feature
SeemaMenon: Existing behave test again...

features/view-an-lpa.feature
SeemaMenon: This is not new again...existing behave.

service-front/app/src/Viewer/src/Form/ShareCode.php
SeemaMenon: amended ....thanks for this Adam.

cooperaj: Further to the work @GemTay has been doing on share code validation this https://regex101.com/r/dSzfs4/2/tests is a pretty nice solution

features/view-an-lpa.feature
SeemaMenon: This is an existing behave test. So shall consider the refactor when writing the Behat on viewer side for this if  that is OK.

features/view-an-lpa.feature
cooperaj: Ahh. Oops. misread

PR #194, opened 2020-01-06
changed regex for included dashes and spaces, added filtering dashes out

GemTay:


service-front/app/src/Actor/src/Form/LpaAdd.php
```diff
@@ -67,14 +67,14 @@ public function getInputFilterSpecification() : array
...
-                            'max'      => 12,
+                            'max'      => 14,
                             'message'  => 'Your passcode must be 12 characters long',
                         ],
                     ],
                     [
                         'name'    => Regex::class,
                         'options' => [
-                            'pattern' => '/[a-zA-Z0-9]{12}/',
+                            'pattern' => '/[[:alnum:]]{12}|[[:alnum:]]{4}-[[:alnum:]]{4}-[[:alnum:]]{4}|[[:alnum:]]{4}\s[[:alnum:]]{4}\s[[:alnum:]]{4}/',
```
GemTay: I used alnum as /w also accepts underscores. I had to explicitly describe the 3 cases (with dashes, & with & without spaces) in this way, as when writing the regex like /[[:alnum:]]{4,4}[\S\s?\-][[:alnum:]]{4,4}[\S\s?\-][[:alnum:]]{4,4}/, a combination of dashes/spaces could be entered and accepted eg 12HJ-98JH 43PL. This regex pattern ensures that if the user uses dashes for example, the dashes appear in both of the two places

service-front/app/src/Actor/src/Form/LpaAdd.php
```diff
@@ -97,10 +97,17 @@ public function getInputFilterSpecification() : array
                         'options' => [
                             'encoding' => 'UTF-8',
                             'min'      => 12,
-                            'max'      => 12,
+                            'max'      => 14,
                             'message'  => 'The reference number must be 12 numbers long',
```
GemTay: TODO - Now that spaces and dashes can be included for both OTP and ref number, the error messages could be misleading. Waiting for updated content from Lewis

service-front/app/src/Actor/src/Form/LpaAdd.php
cooperaj: Checkout the Regex at https://github.com/ministryofjustice/opg-use-an-lpa/blob/d48290fe5509e4a374583c141db135f33181970e/service-front/app/src/Viewer/src/Form/ShareCode.php#L42 for detail on how I solved this.

It doesn't enforce that the user uses spaces or dashes but not both - but I don't see that bing an issue we really need to care about.

cooperaj: https://regex101.com/r/dSzfs4/1 has a set of tests to test that particular regex and I think all bases are covered.

cooperaj: In fact https://regex101.com/r/dSzfs4/2/tests

After playing around (I love a bit of regex) this looks good and covers all tests cases I can figure.

GemTay: Ahh awesome! I couldn't figure out how to make sure the same thing (dash space or no space) was used in both places! Thanks Adam :)

GemTay: Just spotted that 1111 11111111 seems to match, going to have a play around to fix it

GemTay: This has fixed it: https://regex101.com/r/dSzfs4/3/tests I added the alternative case of a 'no space' to the capture group and removed the ? zero or more qualifiers from outside the groups

service-front/app/src/Actor/src/Form/LpaAdd.php
```diff
@@ -97,8 +97,15 @@ public function getInputFilterSpecification() : array
...
-                            'max'      => 12,
-                            'message'  => 'The reference number must be 12 numbers long',
+                            'max'      => 14,
+                            'message'  => 'The LPA reference number must be 12 numbers long',
+                        ],
+                    ],
+                    [
+                        'name'    => Regex::class,
+                        'options' => [
+                            'pattern' => '/\G(\d{4}(?\'dash\' |-|)\d{4}(\g{dash})\d{4})$/',
```
cooperaj: If I paste this into the tester it blows up and all the tests fail. 
I've never seen the \G(...) syntax. What does that do?

service-front/app/src/Actor/src/Handler/LpaAddHandler.php
```diff
@@ -71,6 +71,10 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
                 //  Attempt to retrieve an LPA using the form data
                 $postData = $form->getData();
 
+                //  Filter out dashes and whitespace
+                $postData = str_replace('-', '', $postData);
```
cooperaj: This is fine but I think we need to investigate whether the 'InputFilterSpecification' is capable of filtering the data as well as validating it?

service-front/app/src/Actor/src/Form/LpaAdd.php
```diff
@@ -67,15 +67,15 @@ public function getInputFilterSpecification() : array
...
+                            'max'      => 14,
                             'message'  => 'Your passcode must be 12 characters long',
                         ],
                     ],
                     [
                         'name'    => Regex::class,
                         'options' => [
-                            'pattern' => '/[a-zA-Z0-9]{12}/',
-                            'message' => 'Your passcode must only include letters and numbers',
+                            'pattern' => '/\G([[:alnum:]]{4}(?\'dash\' |-|)[[:alnum:]]{4}(\g{dash})[[:alnum:]]{4})$/',
```
cooperaj: See comment below

service-front/app/src/Actor/src/Form/LpaAdd.php
GemTay: Screenshot 2020-01-13 at 09 00 53
The /G represents the start of the match, similar to ^ and when I tested it without unit tests, and used the live matcher instead, it appeared to match if there were characters before/after, and this also happened when I'd implemented it in the service. The match didnt include the extra characters but I wanted to add /G at the start (or ^) and $ at the end to make sure that if any extra characters come before or after the code then there is no match

cooperaj: Ah ok. I'd recommend using the ^ symbol by preference as it's less work for the Regex engine to do. It's not super important in this case but it's also slightly easier to understand ;)

GemTay: Yeah sure, i'll make that change now :)

service-front/app/src/Actor/src/Form/LpaAdd.php
```diff
@@ -67,15 +67,15 @@ public function getInputFilterSpecification() : array
...
                             'message'  => 'Your passcode must be 12 characters long',
                         ],
                     ],
                     [
                         'name'    => Regex::class,
                         'options' => [
-                            'pattern' => '/[a-zA-Z0-9]{12}/',
-                            'message' => 'Your passcode must only include letters and numbers',
+                            'pattern' => '/^([[:alnum:]]{4}(?\'dash\' |-|)[[:alnum:]]{4}(\g{dash})[[:alnum:]]{4})$/',
+                            'message' => 'Your passcode must only include letters, numbers and dashes',
```
cooperaj: Spaces too?

PR #195, opened 2020-01-10
updated entities and sirius factory to include systemStatus, also add…

GemTay:
…ed inactive filtering in twig

### DIFF
--- a/service-front/app/src/Actor/templates/actor/view-lpa-summary.html.twig
+++ b/service-front/app/src/Actor/templates/actor/view-lpa-summary.html.twig
@@ -104,24 +104,24 @@
                 </dl>
 
                 {% for attorney in lpa.attorneys %}
-                    {% if attorney.systemStatus %}
-                        <h3 class="govuk-heading-m">{{ lpa.attorneys | length == 1 ? '1st' : loop.index | ordinal }} attorney</h3>
 
-                        <dl class="govuk-summary-list govuk-summary-list--no-border">
-                            <div class="govuk-summary-list__row">
-                                <dt class="govuk-summary-list__key">Name</dt>
-                                <dd class="govuk-summary-list__value">{{ actor_name(attorney) }}</dd>
-                            </div>
-                            <div class="govuk-summary-list__row">
-                                <dt class="govuk-summary-list__key">Date of birth</dt>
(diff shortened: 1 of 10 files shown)


service-front/app/src/Actor/templates/actor/view-lpa-summary.html.twig
```diff
@@ -104,23 +104,24 @@
...
-                        <div class="govuk-summary-list__row">
-                            <dt class="govuk-summary-list__key">Date of birth</dt>
-                            <dd class="govuk-summary-list__value">{{ lpa_date(attorney.dob) }}</dd>
-                        </div>
-                        <div class="govuk-summary-list__row">
-                            <dt class="govuk-summary-list__key">Address</dt>
-                            <dd class="govuk-summary-list__value">{{ actor_address(attorney) }}</dd>
-                        </div>
-                    </dl>
+                    {% if attorney.systemStatus %}
```
cooperaj: Filtering out bad actors is not really a frontend thing. There is a class in the api called `LpaDataCleanseDecorator` that you'll want to alter to do the filtering there.

GemTay: Oops okay i'll filter it out in there instead :)

cooperaj: You should just be able to duplicate the existing ghost attorneys method and alter the return logic of the filter function. Since the status field is a boolean you could even just straight up return it.

service-front/app/test/CommonTest/Service/Lpa/SiriusLpaFactoryTest.php
```diff
@@ -72,6 +72,7 @@ public function testCanCreateLpaFromSwaggerExample()
 
         $this->assertInstanceOf(CaseActor::class, $lpa->getAttorneys()[0]);
         $this->assertEquals(new DateTime('1980-10-10'), $lpa->getAttorneys()[0]->getDob()); // from full_example.json
+        $this->assertEquals(true, $lpa->getAttorneys()[0]->getSystemStatus());
```
cooperaj: These tests only really check for the most basic of information from the json file as there are so many properties.

This is fine but going forward it's probably not necessary to check everything.

PR #197, opened 2020-01-15
Uml 318 pdf download

cooperaj:


service-front/app/features/context/Integration/ViewerContext.php
```diff
@@ -0,0 +1,148 @@
...
+                        'country' => '',
+                        'type' => 'Primary',
+                        'addressLine1' => 'Test House',
+                        'addressLine2' => 'Test Road',
+                        'addressLine3' => ''
+                    ]
+                ]
+            ]
+        ];
+    }
```
acsauk-mojd: Like this approach - we're going through an overhaul of our behat stuff so may pinch this. We were considering actually making the API calls to create the data. The only downside I could see is with updates to the models this data is based on but for faster-running suites, its some overhead I could deal with!

cooperaj: This structure is *almost* replicated in the acceptance contexts so if it were to grow any bigger I'd probably move it out to a fixture  file and load it here instead.

service-front/app/features/context/UI/ViewerContext.php
```diff
@@ -0,0 +1,180 @@
...
+                        'country' => '',
+                        'type' => 'Primary',
+                        'addressLine1' => 'Test House',
+                        'addressLine2' => 'Test Road',
+                        'addressLine3' => ''
+                    ]
+                ]
+            ]
+        ];
+    }
```
acsauk-mojd: Some duplication here from the looks of it...does your setup allow you to add this to some base class to extend from so you only have to maintain it in one place?

service-front/app/features/context/UI/ViewerContext.php
```diff
@@ -0,0 +1,180 @@
...
+            ->respondWith(new Response(StatusCodeInterface::STATUS_OK, [], json_encode([
+                'lpa'     => $this->lpaData,
+                'expires' => (new \DateTime('+30 days'))->format('c')
+            ])))
+            ->inspectRequest(function (RequestInterface $request, array $options) {
+                $params = json_decode($request->getBody()->getContents(), true);
+
+                assertInternalType('array', $params);
+                assertEquals($params['name'], $this->lpaSurname);
+                assertEquals($params['code'], str_replace('-', '', $this->lpaShareCode));
```
acsauk-mojd: Could be extracted into an `assertLPAIsCorrect` function or something similar to ease maintenance as its being used three times in this context file.

service-front/app/test/CommonTest/Service/Pdf/PdfResponseTest.php
```diff
@@ -0,0 +1,22 @@
...
+namespace CommonTest\Service\Pdf;
+
+use Common\Service\Pdf\PdfResponse;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\StreamInterface;
+
+class PdfResponseTest extends TestCase
+{
+    /** @test */
+    public function it_is_a_custom_response_that_sets_pdf_headers() {
```
acsauk-mojd: Not a negative, but why snake case all of a sudden?

acsauk-mojd:
Nicely structured and self-contained additions in this PR - testing looks spot on to me. I'm not too close to the domain or problem so can't comment on all paths but seems comprehensive 👍

service-front/app/features/context/Integration/ViewerContext.php
```diff
@@ -0,0 +1,148 @@
...
+        $this->container = $container;
+
+        $this->apiFixtures = $this->container->get(MockHandler::class);
+    }
+
+    /**
+     * @Given /^I have been given access to an LPA via share code$/
+     */
+    public function iHaveBeenGivenAccessToAnLPAViaShareCode()
+    {
```
GemTay: I've had to get full LPA responses for the add an LPA feature tests. I went about this by reading in contents of the full_example.json file in the fixtures folder and assigned it to $this->lpa - i'm not sure if thats a better way to do this? Checkout my PR for reference :) https://github.com/ministryofjustice/opg-use-an-lpa/pull/200/files

GemTay:
Looks really good to me! :)

PR #199, opened 2020-01-15
Uml 416 life sustaining treatment bug

GemTay:


service-front/app/src/Common/src/Entity/Lpa.php
```diff
@@ -227,12 +227,12 @@ public function setLpaDonorSignatureDate(DateTime $lpaDonorSignatureDate) : void
         $this->lpaDonorSignatureDate = $lpaDonorSignatureDate;
     }
 
-    public function getLifeSustainingTreatment() : ?bool
+    public function getLifeSustainingTreatment() : ?string
```
GemTay: After help from Andrew who investigated and found the actual LST value's being returned are strings "Option A" or "Option B". This is the reason for this change and also covers the rare case that it could be null

service-front/app/src/Viewer/templates/viewer/view-lpa.html.twig
```diff
@@ -59,9 +59,9 @@
                                             This LPA can be used as soon as it's registered unless instructions say otherwise.
                                         {% endif %}
                                     {% else %}
-                                        {% if lpa.lifeSustainingTreatment %}
+                                        {% if lpa.lifeSustainingTreatment == 'Option A' %}
                                             The attorneys have the authority to make decisions about life-sustaining treatment
-                                        {% else %}
+                                        {% elseif lpa.lifeSustainingTreatment == 'Option B' %}
                                             The attorneys do NOT have the authority to make decisions about life-sustaining treatment
                                         {% endif %}
```
GemTay: We show the message 'To view this, ask to see the paper LPA' next to the 'Life sustaining treatment' title in the summary when the value is null, however we have decided to not show this message in the banners at the top of the page as it would not make sense

service-front/app/src/Actor/templates/actor/view-lpa-summary.html.twig
```diff
@@ -47,9 +47,9 @@
                             This LPA can be used as soon as it's registered unless instructions say otherwise.
                         {% endif %}
                     {% else %}
-                        {% if lpa.lifeSustainingTreatment %}
+                        {% if lpa.lifeSustainingTreatment in ['Option A', 'option A', 'option a', 'Option a', 'OPTION A'] %}
```
cooperaj: possible to do a twig filter to make it case insensitive? would simplify this a bit

service-front/app/src/Actor/templates/actor/view-lpa-summary.html.twig
```diff
@@ -177,11 +177,15 @@
...
                             <dt class="govuk-summary-list__key">Life sustaining <br>treatment</dt>
-                            {% if lpa.lifeSustainingTreatment %}
-                                <dd class="govuk-summary-list__value">The attorneys have the authority to make decisions about life-sustaining treatment</dd>
-                            {% else %}
-                                <dd class="govuk-summary-list__value">The attorneys do not have the authority to make decisions about life-sustaining treatment</dd>
-                            {% endif %}
+                            <dd class="govuk-summary-list__value">
+                                {% if lpa.lifeSustainingTreatment in ['Option A', 'option A', 'option a', 'Option a', 'OPTION A'] %}
+                                    The attorneys have the authority to make decisions about life-sustaining treatment
+                                {% elseif lpa.lifeSustainingTreatment in ['Option B', 'option B', 'option b', 'Option b', 'OPTION B'] %}
```
cooperaj: all of these infact

PR #200, opened 2020-01-20
Uml 329 cancel adding an lpa

GemTay:
Completed happy path feature tests of adding an LPA first. Will change the functionality of clicking cancel button on the add form once the feature tests of that scenario have been written


service-front/app/features/context/UI/AccountContext.php
```diff
@@ -312,4 +323,68 @@ public function iAmGivenInstructionOnHowToChangeDonorOrAttorneyDetails()
...
+    }
+
+
+    /**
+     * @Given /^The LPA is successfully added$/
+     */
+    public function theLPAIsSuccessfullyAdded()
+    {
+        $this->assertPageAddress('/lpa/dashboard');
+        //$this->assertPageContainsText('Babara Suzanne Gilson');
```
cooperaj: commented code

service-front/app/features/context/UI/AccountContext.php
```diff
@@ -12,6 +12,8 @@
 use GuzzleHttp\Psr7\Response;
 use Psr\Http\Message\RequestInterface;
 use function random_bytes;
+use Exception;
```
cooperaj: For tidyness please sort the imports so that 'functions' are at the bottom. Is guzzlehttp needed?

service-front/app/features/context/Integration/AccountContext.php
```diff
@@ -211,4 +239,68 @@ public function iAmToldThatMyPasswordIsInvalidBecauseItNeedsAtLeast($reason)
...
+
+        // API call for checking LPA
+        $this->apiFixtures->post('/v1/actor-codes/summary')
+            ->respondWith(new Response(StatusCodeInterface::STATUS_OK, [], json_encode([
+                'lpa' => $fullLPA
+            ])));
+
+        $lpa = $this->lpaService->getLpaByPasscode($this->userToken, $this->passcode, $this->referenceNo, $this->userDob);
+
+        if ($lpa->getDonor()->getFirstname() !== 'Ian') {
```
cooperaj: I know this achieves the same thing but it confused me as it's more of a negative test. Probably better to check that the $lpa object matches what you're expecting i.e. the fixture data.

service-front/app/features/add-an-lpa.feature
```diff
@@ -0,0 +1,76 @@
...
+
+  @integration @ui
+  Scenario: The user can add an LPA to their account
+    Given I am on the add an LPA page
+    When I request to add an LPA with valid details
+    Then The correct LPA is found and I can confirm to add it
+    And The LPA is successfully added
+
+  @integration @ui
+  Scenario: The user cannot add an LPA to their account
```
cooperaj: Would probably clarify the name of this but thats just a picky thing. Why can they not add it?

service-front/app/features/add-an-lpa.feature
```diff
@@ -0,0 +1,76 @@
...
+    Given I am on the add an LPA page
+    When I fill in the form and click the cancel button
+    Then I am taken back to the dashboard page
+    And The LPA has not been added
+
+  @ui
+  Scenario Outline: The user cannot add an LPA with an invalid passcode
+    Given I am on the add an LPA page
+    When I request to add an LPA with an invalid passcode format of "<passcode>"
+    Then I am told that my input is invalid and needs to be <reason>
```
cooperaj: Then I am told that my passcode is invalid because <reason>?

service-api/app/test/AppTest/Lpa/fixtures/example_lpa.json
```diff
@@ -0,0 +1,149 @@
+{
```
cooperaj: I think fixture files like this should live at /service-api/app/test/fixtures/ that way we can use them all over the place without weird include paths. What do you think?

I envisage in the future we'll consolidate our tests into something like:

/service/api/app/test/
  unit/
  features/
  fixtures/
  results/

service-front/app/features/context/UI/AccountContext.php
```diff
@@ -125,6 +174,21 @@ public function iEnterCorrectCredentials()
...
     }
 
+    /**
+     * @Given /^I am signed in$/
+     */
+    public function iAmSignedIn()
+    {
+        $this->ui->assertPageAddress('/lpa/dashboard');
+
+        $this->ui->assertPageContainsText('Add your first LPA');
```
cooperaj: I'd remove these two assertions as we may in the future want to use this check on any number of possible pages and being on the dashboard page with this text will not work in that case.

service-front/app/test/CommonTest/Service/Lpa/fixtures/full_example.json
```diff
@@ -1,6 +1,6 @@
 {
```
cooperaj: As with the service-api this file should probably live in app/test/fixtures/

PR #205, opened 2020-01-22
UML 380: Added feature test files for the actor account creation scenarios

SeemaMenon:


service-front/app/features/actor-account-creation.feature
```diff
@@ -0,0 +1,67 @@
...
+    Then I receive unique instructions on how to activate my account
+
+  @ui @integration
+  Scenario Outline: As a new user I want to be shown the mistakes I make while creating an account
+    Given I am not a user of the lpa application
+    And I want to create a new account
+    When I have not provided required information for account creation such as <email1> <email2> <password1> <password2> <terms>
+    Then I should be told my account could not be created due to <reasons>
+    Examples:
+      | email1 | email2 | password1 | password2 | terms | reasons                          |
```
cooperaj: These look great now. My only comment would be to make the conditions match more exactly. For instance for "Confirm you password" I would expect that all fields _except_ password2 be filled with correct data. That way we're making clear we're testing the exact thing we're after.

service-front/app/features/actor-account-creation.feature
```diff
@@ -0,0 +1,67 @@
...
+  @ui @integration
+  Scenario Outline: As a new user I want to be shown the mistakes I make while creating an account
+    Given I am not a user of the lpa application
+    And I want to create a new account
+    When Creating account I provide mismatching <password> <confirm_password>
+    Then I should be told my account could not be created due to <reasons>
+    Examples:
+      | password       | confirm_password  |reasons                     |
+      | password       | pass              | The passwords did not match|
+      | password       | pass              | Your password must include at least one capital letter (A-Z)|
```
cooperaj: Here for instance I'd set confirm_password to be "password". We'd still get the capital letter error but it's more explicit.

service-front/app/features/context/UI/AccountContext.php
```diff
@@ -312,4 +493,83 @@ public function iAmGivenInstructionOnHowToChangeDonorOrAttorneyDetails()
         $this->assertPageContainsText('Let us know if a donor\'s or attorney\'s details change');
         $this->assertPageContainsText('Find out more');
     }
+
+    /**
+     * @When /^I have not provided required information for account creation such as (.*)(.*)(.*)(.*)(.*)$/
```
cooperaj: It's possible to use named parameters for this. I'd look into that as it's a little more clear than (.*). I *think* its something like:

`@When I have not provided required information for account creation such as :email1 :email2 :password1 :password2 :terms`

Though double check that.

cooperaj: Notice its no longer a regular expression.

service-front/app/features/actor-account-creation.feature
SeemaMenon: Ok, done.

service-front/app/features/actor-account-creation.feature
SeemaMenon: Done

service-front/app/features/context/UI/AccountContext.php
SeemaMenon: Not able to get that working now. I doubt the named parameters working for table the way we thought above, may be I am wrong..but parking this for now to proceed with current ticket.

PR #209, opened 2020-01-24
Uml 476 redirect login when logged in

cooperaj:


cooperaj:
I'm going to get some unit test work done to bump up the coverage a bit now that I've removed the unit test of the login handler.

PR #223, opened 2020-02-14
UML-524 content changes for grammar and consistency

LewisCummings:
Minor content changes

## Purpose
Content changes to correct grammar and passive language


cooperaj:
and here the brittleness of tests looking for text fails us :)

Some work will need to happen on the feature tests we have to correct for the different text we're now returning.

PR #225, opened 2020-02-14
Uml 446 disable button after first click

GemTay:
## Purpose

Fixes UML-446

## Approach

Added JS function to add an click event handler to the page. If the click even occurs on a button, the button is disabled preventing the user from clicking again and the form is submitted 

## Checklist


service-front/web/src/javascript/disableButtonOnClick.js
```diff
@@ -0,0 +1,13 @@
+const disableButtonOnClick = () => {
+  var btn = document.getElementsByClassName('govuk-button');
```
cooperaj: This does not use the data attribute?

service-front/web/src/javascript/disableButtonOnClick.js
```diff
@@ -0,0 +1,13 @@
...
+  for (var i = 0; i < btn.length; i++) {
+
+      btn[i].addEventListener('click', function disableButton() {
+          this.disabled = true;
+          document.forms[this.form.name].submit()
```
mattmachell: You don't need the submit here as the submit event will fire anyway if you don't prevent the default events for the button.

service-front/web/src/index.js
```diff
@@ -3,7 +3,16 @@
 import './scss.js';
 import { initAll } from 'govuk-frontend';
 import jsEnabled from './javascript/jsEnabled';
+import disableButtonOnClick from './javascript/disableButtonOnClick'
 
 initAll();
 
 jsEnabled(document.body);
+disableButtonOnClick();
```
cooperaj: I would recommend doing the selector query in this file and passing the selection results to the disableButtonClick function. It keeps things in one place which is nice.

service-front/web/src/javascript/disableButtonOnClick.js
```diff
@@ -0,0 +1,13 @@
+const disableButtonOnClick = () => {
+  var btn = document.getElementsByClassName('govuk-button');
+  for (var i = 0; i < btn.length; i++) {
+
+      btn[i].addEventListener('click', function disableButton() {
+          this.disabled = true;
```
cooperaj: I would recommend putting a timeout on this so that after a set amount of time (30 seconds?) the button becomes reenabled - though we might want to discuss this as a pattern with others (@mattmachell)

service-front/web/src/javascript/disableButtonOnClick.js
mattmachell: Rather than adding the event to every button of a type, consider adding the event handler to the document and then filtering the events for those on a button. It's a generally better pattern called Event Delegation: https://davidwalsh.name/event-delegate

service-front/app/src/Actor/templates/actor/create-account.html.twig
```diff
@@ -56,7 +56,7 @@
 
                                 {{ govuk_form_element(form.get('terms'), { 'label': 'I agree to the <a href="#" class="govuk-link">terms of use</a>' }) }}
 
-                                <button type="submit" class="govuk-button">Create account</button>
+                                <button data-prevent-double-click="true" type="submit" class="govuk-button">Create account</button>
```
mattmachell: Data attribute pattern for events is a win!

service-front/web/src/javascript/disableButtonOnClick.js
cooperaj: I would tend to agree, but in this case we won't be adding or removing buttons on a page once rendered, which is the strength of delegating so I don't know what we gain by doing it?

mattmachell: Mainly it's also a good habit/pattern to get into, but also a memory/speed thing

service-front/web/src/javascript/disableButtonOnClick.js
mattmachell: It depends what the aim is. It's definitely useful to think about what the recovery pattern is if the button is clicked and then else something fails.

service-front/web/src/javascript/disableButtonOnClick.js
GemTay: @mattmachell Unfortunately I found that without this the form doesn't submit

service-front/app/src/Actor/templates/actor/create-account.html.twig
GemTay: @cooperaj Setting the data attribute here

service-front/web/src/javascript/disableButtonOnClick.js
cooperaj: I think you have to return a positive results from the function `return true` should do it.

service-front/web/src/javascript/disableButtonOnClick.js
GemTay: @mattmachell Made some changes to fit this pattern, let me know if you think its okay

service-front/app/src/Actor/templates/actor/create-account.html.twig
cooperaj: yeah, but you're not using it in the code. :)

You'll want to do a check for (basically) `if type is button and data-prevent attribute is true`

service-front/web/src/javascript/disableButtonOnClick.js
```diff
@@ -0,0 +1,17 @@
+const disableButtonOnClick = (form) => {
+
+  for (let i = 0; i < form.length; i++) {
+
+      form[i].addEventListener('click', function disableButton(e) {
+
+          if (e.target.nodeName == 'BUTTON') {
+              if (e.target.getAttribute('data-prevent-double-click')) {
```
cooperaj: I'd roll this into a single IF statement to clean it up a bit. 

`if (target == 'BUTTON' && attribute('data-prevent')) {`

service-front/web/src/javascript/disableButtonOnClick.js
```diff
@@ -0,0 +1,17 @@
...
+                  e.target.disabled = true;
+                  document.forms[form[i].name].submit();
```
cooperaj: You should just be able to `return true` from this function and it'll bubble the event as normal.

service-front/web/src/javascript/disableButtonOnClick.js
GemTay: @cooperaj Returning true doesn't submit the form either, as I think disabling the button prevents the default event

service-front/web/src/javascript/disableButtonOnClick.js
GemTay: That doesnt seem to work, I believe disabling the button prevents the default event from happening

GemTay:
Going to need @SeemaMenon 's cancel code ticket UML 267 to be merged first to add functionality to the cancel code button and test it works on check access page with multiple forms

PR #227, opened 2020-02-14
add SESSION_EXPIRES env var for view and use tasks

andrewpearce-digital:
## Purpose
Each front end wants to have independant session expiry values, and developers would like the ability to have shorter expiry times on dev environments to make it easier to test.

Fixes UML-536

## Approach

Use the account level variables to set a session expiry value, and pass that value into each front end's ECS containner definition.

This PR does not affect the current session expiry, onnly provides a new place to retrieve a value from.

## Checklist

### DIFF
--- a/terraform/environment/actor_ecs.tf
+++ b/terraform/environment/actor_ecs.tf
@@ -143,7 +143,7 @@ locals {
       "value": "${var.container_version}"
     }]
   }
-  
+
 EOF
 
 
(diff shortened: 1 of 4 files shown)


williamfalconeruk:
Looks ok to me

andrewpearce-digital:
got an update that the preference is to set a value in minutes. 
Production (and prod like) should have 30 minutes for view, and 20 minutes for use. I've set dev builds to be 10 and 5 respectively.

PR #228, opened 2020-02-17
Improve behat code coverage

cooperaj:
## Purpose
This installs a behat extension that allows us to pull code coverage metrics from behat runs.

## Approach
Install forked extension and configure. Also add some handy composer scripts to make it easy to run.

## TODO
Needs to be run in CI so changes need making to Circle config. Does, however generate really create reports locally.

## Checklist


cooperaj:
This is a dead end unfortunately so am abandoning for now

PR #231, opened 2020-02-18
add listener rule to redirect port 80 to port 443

andrewpearce-digital:
## Purpose
Users have reported that the service is not working when they don't use `https://` in the service URLs. The service is only listening for `https://` requests.

Fixes UML-531

## Approach

This PR adds a listener rule to each load balancer to redirect the user to `https://` if they omit it.

## Checklist


andrewpearce-digital:
load balancer rule to redirect requests to https in place
image

PR #235, opened 2020-02-19
UML-501 resend activation

williamfalconeruk:
## Purpose
To give a more meaningful flow when a user attempts to login and has not yet activated, so they are kept informed, and allowing them to resend activation code as needed.

Fixes UML-501

## Approach

This changes the flow from displaying an unusable error message on login to something more meaningful, redirecting to the account activation success page instead. 
There is currently a subtle title change - this will need a further review to see if there needs to be tweaks to the content. We have added a flag so we can switch based on the context.

## Learning
General learning around the PHP stack and its basic architecture, but nothing to share at this stage with the team.

## Checklist


service-front/app/src/Actor/src/Handler/CreateAccountSuccessHandler.php
```diff
@@ -71,11 +71,18 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
         $resend = (isset($params['resend']) && $params['resend'] === 'true');
+        $accountExists = (isset($params['accountExists']) && $params['accountExists'] === 'true');
 
         if (is_null($emailAddress)) {
             return $this->redirectToRoute('create-account');
         }
-
+        if ($accountExists === true)
+        {
+            return new HtmlResponse($this->renderer->render('actor::create-account-success', [
```
cooperaj: Would prefer to see this fall through to the final render call in this function. The only thing that is different is the passing of a bool and that can be worked with there

service-front/app/features/context/UI/AccountContext.php
```diff
@@ -136,13 +136,10 @@ public function iAmDirectedToMyPersonalDashboard()
      */
     public function iAmToldMyAccountHasNotBeenActivated()
     {
-        // TODO UML-501
-        //      When a user account has not been activated we will need a better user flow
-        //      around how this works. Probably taking them to a help page that details
-        //      what they need to do, including resending the email.
-        $this->ui->assertPageContainsText('Email and password combination not recognised');
+        $this->ui->assertPageAddress('/create-account-success');
```
cooperaj: Probably don't need to test the URL here. The test step is about the behaviour of being told something. That might not happen on a particular page - we may end up migrating to flash messages in the future etc.

williamfalconeruk: ok - noted. i'll change this now

service-front/app/src/Actor/src/Handler/CreateAccountSuccessHandler.php
williamfalconeruk: will refactor as requested.

service-front/app/src/Actor/src/Handler/CreateAccountSuccessHandler.php
```diff
@@ -102,6 +103,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
 
         return new HtmlResponse($this->renderer->render('actor::create-account-success', [
             'emailAddress' => $emailAddress,
+            'accountExists' => $accountExists === true ? 'true' : 'false'
```
cooperaj: Does accountExists need to be a string? Just pass it straight as a bool?

service-front/app/src/Actor/templates/actor/create-account-success.html.twig
```diff
@@ -1,6 +1,12 @@
 {% extends '@actor/layout.html.twig' %}
 
-{% block html_title %}Account created - {{ parent() }} {% endblock %}
+{% block html_title %}
+    {% if accountExists| default is same as('true') %}
```
cooperaj: With a bool this could be reduced to (iirc)

{% if accountExists %}

service-front/app/src/Actor/src/Handler/CreateAccountSuccessHandler.php
williamfalconeruk: so i tried this - for some reason it didn't work. let me have another go at it.

service-front/app/src/Actor/templates/actor/create-account-success.html.twig
williamfalconeruk: for the same reason above - but i'm willing to have another go at it

service-front/web/package-lock.json
```diff
@@ -4726,7 +4726,8 @@
             "ansi-regex": {
```
GemTay: How come the package-lock.json is included in the commit?

williamfalconeruk: it's a very good question - i'm not sure if we should put this into .gitignore this or not, it just regenerated when I spun up my local env. @cooperaj thoughts?

cooperaj: It should be in the project as we want to ensure we lock the dependencies for CI builds, but I do wonder why it’s a part of _this_ PR. It’s somehow become updated and you’ve included it I guess. 

The fix would be just to remove it from this PR.

williamfalconeruk: ok - so not needed - i'll remove now.

PR #239, opened 2020-02-20
Uml 560 adding same lpa twice

GemTay:
## Purpose
The is a missing bad path for attempting to adding the same LPA twice, where the user should be told that the LPA was not found if they try to add an LPA that has already been added

Fixes UML-560

## Approach

Added acceptance & integration tests on both front and API side to mock out the scenario and responses

## Checklist


service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -795,4 +796,87 @@ public function theFullLPAIsDisplayedWithTheCorrect($message)
...
+                'actor-code' => $this->oneTimeCode,
+                'uid'        => $this->lpaUid,
+                'dob'        => $this->userDob,
+            ],
+            [
+                'user-token' => $this->userAccountId
+            ]
+        );
+
+        $this->ui->assertSession()->statusCodeEquals(StatusCodeInterface::STATUS_NOT_FOUND);
```
cooperaj: The assertion is happening before the service is actually called?

service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -795,4 +796,87 @@ public function theFullLPAIsDisplayedWithTheCorrect($message)
...
+
+        $response = $this->getResponseAsJson();
```
cooperaj: This is likely the wrong response?

service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -795,4 +796,87 @@ public function theFullLPAIsDisplayedWithTheCorrect($message)
...
+    public function theIAmToldThatTheLPAWasNotFound()
+    {
+        // Not needed for this context
+    }
+
+    /**
+     * @Given /^The LPA should not be duplicated$/
+     */
+    public function theLPAShouldNotBeDuplicated()
+    {
```
cooperaj: Due to the way we test is this actually being used? Since this is just verifying that the fixture data we set for this page view doesn’t show duplicate results and we don’t set any in the fixture?

service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -740,6 +740,68 @@ public function theFullLPAIsDisplayedWithTheCorrect($message)
...
+     */
+    public function theIAmToldThatTheLPAWasNotFound()
+    {
+        // Not needed for this context
+    }
+
+    /**
+     * @Given /^The LPA should not be duplicated$/
+     */
+    public function theLPAShouldNotBeDuplicated()
```
cooperaj: This is similar to the above. We only return 1 lap in the fixture data?

service-api/app/features/context/Acceptance/AccountContext.php
GemTay: the service is called in the post just above on line 817

cooperaj: Ha yes, my bad. The comment made me dismiss it as a fixture. :)

service-api/app/features/context/Acceptance/AccountContext.php
GemTay: @cooperaj Yeah thats true, this is kinda just testing that the dashboard is showing one LPA when we expect it to - i'll remove the test

cooperaj: It it useful as a piece of language in the feature file so I’d keep that. Just put a “not used in this context” in the step file

PR #240, opened 2020-02-24
Uml 565 dashboard on login feature test

GemTay:
## Purpose
To add feature tests to cover the scenario of a returning user logging in, having logged in previously, and being taken to the dashboard page (rather than the add an LPA page)

Fixes UML-565

## Approach

Written scenario steps for the front ui and integration. For the step 'I have logged in previously', I set the LastLogin value to null to represent their first ever login (even though their previous login doesn't necessarily mean their first ever login) and had this changed to a date time in the step 'When I sign in' to show the difference in response from the first login and how that should affect the redirect

## Checklist

### DIFF
--- a/service-front/app/features/actor-login.feature
+++ b/service-front/app/features/actor-login.feature
@@ -39,9 +39,9 @@ Feature: A user of the system is able to login
     When I attempt to sign in again
     Then I am directed to my dashboard
 
-#  @ui @integration
-#  Scenario: A user is taken to the dashboard page when the login having logged in previously
-#    Given I am a user of the lpa application
-#    And I have logged in previously
-#    When I sign in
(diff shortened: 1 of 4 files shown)


service-front/app/features/actor-login.feature
```diff
@@ -6,6 +6,7 @@ Feature: A user of the system is able to login
 
   Background:
     Given I am a user of the lpa application
+    And I have been given access to use an LPA via credentials
```
williamfalconeruk: i'm just wondering - do we need the "via credentials" part?

GemTay: @williamfalconeruk Technically we write it to clarify the difference between using an LPA with paper document and using the 'Use an LPA' service with credentials

PR #252, opened 2020-03-04
UML-431 - Codecov added to pipeline

williamfalconeruk:
- All lcov uploads to same as php one
- Tweaks to use node for jest test instead of puppeteer
- Use node-browser docker image
- Added arch key for uniqueness in caching
- Additional steps for calling webhook on coveralls 
- Added docker images for php-cli-node for the upload
this is subject to the result of issue request:
https://github.com/lemurheavy/coveralls-public/issues/1390#issuecomment-594410482
- reverted to use codecov. 

## Purpose

Fixes UML-431

## Approach

- Changes to CircleCI process (running parallel with coveralls) 
 - **NOTE** both codecov and coveralls are running at this time.  
- Codecov's behaviour is set so that:
     -  it shows all of its data and edits the coverage comment.
     - It will always update that comment every time a change is made even if there is no change.

Next steps will be to: 
- switch off coveralls and and configure git to expect codecov to be completed.
- change the settings so it does 80% coverage as per coveralls.

## Learning

- Circle CI Build process
- Coveralls
- CodeCov.io

## Checklist


.circleci/config.yml
```diff
@@ -585,7 +585,7 @@ orbs:
 
       # Code Coverage
 
-      codecov_php-unit_upload:
+      codecov_upload:
```
cooperaj: Don't think codecov is even used at this point. This task definition just hasn't been taken out yet

cooperaj:
Wait, codecov is working?

🕺

williamfalconeruk:
> Wait, codecov is working?
> 
> 🕺

Yup! only. Thing is now it needs to add a PR comment when done. currently not possible as CodeCov is waiting for the build to complete, as it has an approval state. Once the build split is done in UML-540, this should theoretically work as i'd expect.

williamfalconeruk:
Closing for now - will reinstate at a later date.

williamfalconeruk:
Screenshot 2020-06-10 at 11 20 08 (2)
codecov coverage working now.

.circleci/config.yml
```diff
@@ -288,19 +310,21 @@ orbs:
         docker: [image: circleci/php:7-cli-browsers]
       node:
         docker: [image: circleci/node:14.0]
+      node-browsers:
+        docker: [image: circleci/node:14.0-browsers]
```
andrewpearce-digital, line 314: Do you think it's worth consolidating the node and node-browsers executors to all be node-browsers?

As it's a circleci image it'll be cached and quickly pulled anyway.

codecov.yml
```diff
@@ -0,0 +1,6 @@
+codecov:
+  bot: cooperaj
```
andrewpearce-digital, line 2: How come we're calling this cooperaj? We could/should use opg-use-an-lpa (admin user on github)

williamfalconeruk: lets try that out. i wasn't keen on using this anyway.

.circleci/config.yml
williamfalconeruk: I'll take a look...

codecov.yml
williamfalconeruk: done.

.circleci/config.yml
cooperaj: I the browsers one will start a number of background processes IIRC so if theres a cost/speed requirement it's probably better just to have different. Is it an issue cause the same could be said for the PHP side too

williamfalconeruk: Doesn't seem to be any issues with this as it stands. I've made the change, for now, I can always fix this back later if it does become a problem...

codecov.yml
williamfalconeruk: can't remember why I used adams one TBH...

PR #260, opened 2020-03-12
fix up python version to 3 and add apk upgrade

williamfalconeruk:
## Purpose

To fix an issue with Zap DockerFile that was causing issues locally when spinning up in docker-compose. This does not commission the docker file to run the zap scripts it has yet.

## Approach
- Upgrade PIP to PIP3 and add some prerequisites and no-cache to prevent warnings.
- add an Apk upgrade to prevent vuln creep.
- Made the docker file clearer in terms of steps taken by running each step separately.

## Learning

- general research around docker and python:
https://stackoverflow.com/questions/7446187/no-module-named-pkg-resources 
https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Upgrade_a_Running_System
 - also, what is zap? 
https://www.zaproxy.org/

## Checklist

### DIFF
--- a/zap/Dockerfile
+++ b/zap/Dockerfile
@@ -4,12 +4,10 @@ WORKDIR /zap
 
 USER root
 
-RUN apk upgrade
-RUN  apk add python3 py3-pip
-RUN rm -rf /var/cache/apk/*
-RUN pip3 install --no-cache-dir --upgrade pip
-RUN pip3 install --no-cache-dir --upgrade setuptools
-RUN pip3 install --no-cache-dir --upgrade zapcli
+RUN apk upgrade \
+ && apk add python3 py3-pip \
+ && rm -rf /var/cache/apk/* \
+ && pip3 install --no-cache-dir --upgrade pip zapcli
 
 COPY ./scripts/ /scripts
 
--- a/zap/Dockerfile
+++ b/zap/Dockerfile
@@ -4,8 +4,12 @@ WORKDIR /zap
 
 USER root
 
-RUN apk add python3 py-pip && rm -rf /var/cache/apk/* \
-    && pip install --upgrade zapcli
+RUN apk upgrade
+RUN  apk add python3 py3-pip
+RUN rm -rf /var/cache/apk/*
+RUN pip3 install --no-cache-dir --upgrade pip
+RUN pip3 install --no-cache-dir --upgrade setuptools
+RUN pip3 install --no-cache-dir --upgrade zapcli
 
 COPY ./scripts/ /scripts
 
--- a/zap/Dockerfile
+++ b/zap/Dockerfile
@@ -7,7 +7,7 @@ USER root
 RUN apk upgrade \
  && apk add python3 py3-pip \
  && rm -rf /var/cache/apk/* \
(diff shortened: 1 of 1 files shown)


zap/Dockerfile
```diff
@@ -4,8 +4,12 @@ WORKDIR /zap
 
 USER root
 
-RUN apk add python3 py-pip && rm -rf /var/cache/apk/* \
-    && pip install --upgrade zapcli
+RUN apk upgrade
```
cooperaj: Splitting up a run command makes it clearer but it actually does bad juju to the resultant images. Each command in a docker file makes a layer that is always a part of the image. If you run the upgrade and install and *then* run a cache clear up your image will contain the cache files twice and then be removed.

zap/Dockerfile
```diff
@@ -4,8 +4,12 @@ WORKDIR /zap
...
+RUN  apk add python3 py3-pip
+RUN rm -rf /var/cache/apk/*
+RUN pip3 install --no-cache-dir --upgrade pip
```
cooperaj: The install of pip and setup tools is not needed as they're bundled as a part of py3-pip system package

zap/Dockerfile
williamfalconeruk: ah yes, makes a bit of sense now. will revert to doing that

zap/Dockerfile
williamfalconeruk: ok - will remove. I apologize for not understanding this.

zap/Dockerfile
```diff
@@ -4,12 +4,10 @@ WORKDIR /zap
...
-RUN apk upgrade
-RUN  apk add python3 py3-pip
-RUN rm -rf /var/cache/apk/*
-RUN pip3 install --no-cache-dir --upgrade pip
-RUN pip3 install --no-cache-dir --upgrade setuptools
-RUN pip3 install --no-cache-dir --upgrade zapcli
+RUN apk upgrade \
+ && apk add python3 py3-pip \
+ && rm -rf /var/cache/apk/* \
+ && pip3 install --no-cache-dir --upgrade pip zapcli
```
williamfalconeruk: Note pip has a version 20 which is why it is here (the base one comes down as 18)

cooperaj: Yeah, I don't see that being an issue. The OS packages will contain security fixes and having the latest version of PIP isn't necessary unless the installation of zapcli requires it.

williamfalconeruk: @cooperaj - ready for another review - let me know if ok with this.

williamfalconeruk: I was trying to get rid of the warnings that came up. but happy either way :)

zap/Dockerfile
```diff
@@ -7,7 +7,7 @@ USER root
 RUN apk upgrade \
  && apk add python3 py3-pip \
  && rm -rf /var/cache/apk/* \
- && pip3 install --no-cache-dir --upgrade pip zapcli
+ && pip3 install --no-cache-dir --upgrade zapcli
```
williamfalconeruk: Should be good to go now.

PR #261, opened 2020-03-12
Monitoring Uptime

GemTay:
## Purpose
As a webops engineer/dev/PMI want to monitor the status of the use service in UALSo that I can be notified when there is an outage or service disruption in order to restore the service to it's normal operating state

Fixes UML-425

## Approach

* Create a cloudwatch alarm for each side of the service, that does a route53 dns health check for a HTTP response.
* We created new security groups for the health-checks and attached them to the relevant frontend load balancers

## Learning

Route53 - The HTTP checks we were doing have to receive a response within 2 seconds, which was not possible with our current health check endpoint - https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/dns-failover-determining-health-of-endpoints.html

Recommendation: We should split out the health checks for the service into lambda functions for both sides of the service. These lambda functions should be called periodically, and have their response reported to a cloud watch alarm

## Checklist


terraform/environment/.envrc
```diff
@@ -1,4 +1,4 @@
 export TF_VAR_default_role=operator
 export TF_VAR_management_role=operator
 export TF_CLI_ARGS_init="-backend-config=role_arn=arn:aws:iam::311462405659:role/operator -upgrade=true -reconfigure"
-# export TF_WORKSPACE=
+export TF_WORKSPACE=261-UML425Mon
```
williamfalconeruk: this will need to be emptied

terraform/environment/credentials.tf
```diff
@@ -26,6 +26,16 @@ provider "aws" {
   }
 }
 
+provider "aws" {
+  region = "us-east-1"
```
williamfalconeruk: just a question - is this the right region setting?
 (i thought we're operating from ireland...)

GemTay: We have to use us-east-1 as thats the only region where route53 healthchecks are done

PR #266, opened 2020-03-16
Fixed missing created by bug

GemTay:
## Purpose
Fix the bug causing the missing "Created By" value

Fixes UML-426

## Approach

I could not re-create the scenario despite following the steps required to do so. Instead, I was shown the 'Oops' 500 error message page when attempting to visit the check access codes page on the 2nd LPA i had added and created access codes for (The first LPA added always worked but this is expected due to the bug found). I found that the issue was in the API side 'LpasResourceCodesCollectionHandler' where we get all LPA's for the user, and assign it to $actorId. Next, inside a for loop, for each code (of the LPA selected to check access codes) we create an 'ActorId' key and assign it the value of the first ActorId in the list of all the users LPA's: **$actorId[0]['ActorId'];** This is the bug, as we were not assigning the correct ActorId for an given LPA's access codes. To fix this I changed the $actorId variable to get the LPA in question. Finally I changed the ActorId key to be assigned the actorId of the LPA found 

## Learning

## Checklist


service-api/app/src/App/src/Handler/LpasResourceCodesCollectionHandler.php
```diff
@@ -150,11 +150,11 @@ private function handleGet(ServerRequestInterface $request) : ResponseInterface
         if (!empty($viewerCodes)) {
             $viewerCodesAndStatuses = $this->viewerCodeActivityRepository->getStatusesForViewerCodes($viewerCodes);
 
-            $actorId = $this->userLpaActorMap->getUsersLpas($request->getAttribute('actor-id'));
+            $lpa = $this->userLpaActorMap->get($request->getAttribute('user-lpa-actor-token'));
```
cooperaj: is the intention of this handler not to return a collection of lpas. this will just return one?

GemTay: @cooperaj the purpose is to return a collection of codes for a single lpa - its called inside of a loop in the dashboard handler, where it loops through each LPA, gets that LPA's codes and adds the actor Id

cooperaj: ok. that makes sense. It's taken me a few minutes to figure this out so might be worth adding a few comments to the code.

Has this change in behaviour been tested properly in the api acceptance tests?

GemTay: all done :)

PR #272, opened 2020-03-18
changes to make the email field case insensitive and tests added

SeemaMenon:
## Purpose
Ticket is make email address case in sensitive  when creating an account, logging in and resetting password

Fixes UML-600

## Approach

Added filters and feature tests

## Learning

## Checklist

### DIFF
--- a/service-front/app/src/Actor/src/Form/PasswordResetRequest.php
+++ b/service-front/app/src/Actor/src/Form/PasswordResetRequest.php
@@ -11,6 +11,8 @@ use Laminas\Filter\StringToLower;
 use Laminas\InputFilter\InputFilterProviderInterface;
 use Laminas\Validator\Identical;
 use Laminas\Validator\NotEmpty;
+use Laminas\Filter\StringTrim;
+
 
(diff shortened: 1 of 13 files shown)


service-front/app/src/Actor/src/Form/PasswordResetRequest.php
```diff
@@ -71,7 +77,7 @@ public function getInputFilterSpecification(): array
                 'required' => true,
                 'filters'  => [
                     [
-                        'name' => StringToLower::class,
+                        'name' => StringTrim::class,
```
williamfalconeruk: does this filter need both to lower and trim as per the above email field?

SeemaMenon: Nope, this was deliberately removed. This is because the validator Identical for email_confirm field does a comparison of the raw unfiltered email field data against the email_confirmation field data . Hence converting the confirmation email to lowercase would make the check fail always if user has entered the email addresses in uppercase.

The ticket is support lower and upper case entries for the email address fields and allow spaces too.  we can speak if this is vague.

PR #273, opened 2020-03-19
Uml 296 viewer code brute force

cooperaj:
## Purpose
This PR implements brute force protection for the viewer code entry.

Fixes UML-296

## Approach
This code implements a moving window rate limiter on the failure of viewer code entries. It *only* tracks failures of viewer codes but when hit will deny all requests from the identified user until the window moves enough or the overall TTL value expires (in this PR configured as the same value)

The initial problem is to uniquely identify the user. The approach here is to has the value of the users IP address and a number of browser sent HTTP headers. This identity is passed into the limiters which are manually triggered by code (at this time). 

Once an identity is calculated it is stored in the user session. Subsequent accesses will have this value compared and if a difference is discovered this is logged with an error code for future AWS WAF integration work.

For this PR the CheckCodeHandler will save a limit access attempts when the code is rejected as not found. Upon hitting the limit (as a part of that call, or subsequent accesses) the identity will be rejected at the middleware level.

## Checklist


service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
```diff
@@ -0,0 +1,37 @@
...
+     * @param \Psr\Http\Message\ServerRequestInterface $request
+     * @return string
+     */
+    public function id(\Psr\Http\Message\ServerRequestInterface $request): string
+    {
+        $headersToHash = [
+            'accept',
+            'accept-encoding',
+            'accept-language',
+            'user-agent',
```
mattmachell, line 23: Might have to be a bit careful using user-agent here, as a common pattern is to randomise an element of it per request...

hawx: And there is the opposite problem because of [Intent to Deprecate and Freeze: The User-Agent string ](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/-2JIRNMWJ7s/yHe4tQNLCgAJ). Are we going to end up blocking the n+1th Chrome user if n successful requests have happened in the same time frame?

service-front/app/src/Common/src/Middleware/Security/RateLimitMiddleware.php
```diff
@@ -0,0 +1,51 @@
...
+    private $rateLimitServiceFactory;
+
+    public function __construct(RateLimitServiceFactory $rateLimitServiceFactory)
+    {
+        $this->rateLimitServiceFactory = $rateLimitServiceFactory;
+    }
+
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        if (null !== $identity = $request->getAttribute(UserIdentificationMiddleware::IDENTIFY_ATTRIBUTE)) {
```
hawx, line 37: cheeky, but confusing, to have logic and assignment in here?

service-front/app/src/Common/src/Service/Log/ErrorCodes.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    /**
+     * A share code has been attempted and not found
+     */
+    public const SHARE_CODE_NOT_FOUND = 'SHRC_N_F';
+
+    /**
+     * An incoming identity hash has unexpectedly changed
+     */
+    public const IDENTITY_HASH_CHANGE = 'IDHSH_C';
```
hawx, line 25: why not just the actual name (like `'IDENTITY_HASH_CHANGE'`) if these are going to get logged out?

service-front/app/src/Common/src/Service/Security/RateLimit/KeyedRateLimitService.php
```diff
@@ -0,0 +1,73 @@
...
+     * @param string $identity A unique identity to track
+     * @param string $key An optional piece of information to key this limit for
+     * @return bool Is the identity limited for the given key
+     */
+    function isLimited(string $identity, string $key = ''): bool
+    {
+        $recordKey = $this->getRecordKey($identity, $key);
+
+        /** @var array $accessRecords */
+        if (null === $accessRecords = $this->cacheService->getItem($recordKey)) {
```
hawx, line 28: so many `=`s

service-front/app/src/Common/src/Service/Security/RateLimitService.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Common\Service\Security;
+
+use Laminas\Cache\Storage\StorageInterface;
+use Psr\Log\LoggerInterface;
+
+abstract class RateLimitService implements RateLimiterInterface
```
hawx, line 10: Is an abstract class necessary since there is only KeyedRateLimitService?

service-front/app/src/Common/src/Service/Security/RateLimitServiceFactory.php
```diff
@@ -0,0 +1,86 @@
...
+
+        if (!isset($config['ratelimits'][$limitName]['type'])) {
+            throw new RuntimeException('Missing rate limit type configuration for ' . $limitName);
+        }
+
+        $config['ratelimits'][$limitName]['storage']['options']['namespace'] = $limitName;
+        $cacheAdaptor = StorageFactory::factory($config['ratelimits'][$limitName]['storage']);
+
+        $type = $config['ratelimits'][$limitName]['type'];
+        switch ($type) {
```
hawx, line 51: same question as for the abstract class I guess, a `switch` with only one `case` seems kind of lonely

service-front/app/src/Common/src/Service/Security/RateLimitServiceFactory.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public function all(): array
+    {
+        $config = $this->container->get('config');
+
+        if (!isset($config['ratelimits'])) {
+            throw new RuntimeException('Missing rate limits configuration');
+        }
+
+        $limits = [];
```
hawx, line 79: `array_map` with `array_keys`?

service-front/app/config/pipeline.php
```diff
@@ -46,6 +46,8 @@
     $app->pipe(\Mezzio\Session\SessionMiddleware::class);
     $app->pipe(\Mezzio\Csrf\CsrfMiddleware::class);
     $app->pipe(\Common\Middleware\Logging\RequestTracingMiddleware::class);
+    $app->pipe(\Common\Middleware\Security\UserIdentificationMiddleware::class);
+    $app->pipe(\Common\Middleware\Security\RateLimitMiddleware::class);
```
hawx, line 50: You're always going to need both of these middleware right? So would it make more sense to combine them to one?

service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
hawx: The ticket is kind of vague with 'malicious actor'. Who are we trying to protect against?

cooperaj: The hash is a combination of all those values so if one of those values happens to be the same in all versions of chrome it only reduces the uniqueness of the generated token.

That said if an attacker is randomising it then we have a bit more of an issue. 

The way we generate the id is completely compartmentalised here so whatever we come up with will be easy to slot in.

I was thinking of making it a two part hash. Having just the IP address hashed and concatenated into the combination of hashes. Then a limiter class could be built that watches for combinations of the first and second parts. It would have to be generous to protect users behind proxy servers or NATs but would provide protection against an attacker doing this kind of header randomisation.

service-front/app/src/Common/src/Middleware/Security/RateLimitMiddleware.php
cooperaj: Maybe. It's a supported language construct that I'm not really against, especially in this case where the property is *only* accessed in the statement (which I guess is the only time you can really do it)

service-front/app/src/Common/src/Service/Log/ErrorCodes.php
cooperaj: No real reason. Just figured it'd make the logs easier to parse, but given we log in json now i'll change these to read properly

service-front/app/src/Common/src/Service/Security/RateLimit/KeyedRateLimitService.php
cooperaj: always triple ='s.

anything else is incorrect

service-front/app/src/Common/src/Service/Security/RateLimitService.php
cooperaj: The plan is to add more dependant on limiter requirements.

service-front/app/src/Common/src/Service/Security/RateLimitServiceFactory.php
cooperaj: further rate limter work will likely involve the creation of new types. this was just forward planning

service-front/app/src/Common/src/Service/Security/RateLimitServiceFactory.php
cooperaj: noted. will change.

service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
mattmachell: @hawx they aren't the only browser planning on that either. Although it would be n+1 chrome user from the x-forwarded-for IP combined with the other elements too.

service-front/app/config/pipeline.php
cooperaj: This violates the single responsibility principle and goes against our choice of framework.

Expressive/Mezzio and other PSR15 frameworks are built around the idea of composable request pipelines using small single action components. We run a number of middlewares right now that we could conceivably combine into a mega one but that would be a nightmare of entangled logic/nested if's etc.

hawx: True, you wouldn't want to do that. But (just for additional 🤔, not saying do what I suggest because of...) you could also consider the act of "user identification" an implementation detail of "rate limiting". Then you could argue it should be hidden as nothing else should come to rely on this context/domain specific definition of a User.

service-front/app/src/Common/src/Service/Security/RateLimitService.php
hawx: Having maintained codebases that have had plans to expand things, but never happened, I can only say [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and walk away

service-api/app/src/App/src/ConfigProvider.php
```diff
@@ -7,7 +7,7 @@
 use Aws;
 use Http;
 use Psr;
-use Zend;
+use Laminas;
```
williamfalconeruk, line 10: hmm - I wonder if I missed these in the upgrade? sure I searched for this....

service-front/app/composer.json
```diff
@@ -92,16 +95,26 @@
...
-        "unit-test": "docker-compose -f ../../docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose-integration.yml run actor-app /app/vendor/bin/phpunit --colors=always",
-        "acc-test": "docker-compose -f ../../docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose-integration.yml run actor-app php -dapc.enable_cli=1 /app/vendor/bin/behat -p acceptance",
-        "int-test": "docker-compose -f ../../docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose-integration.yml run actor-app /app/vendor/bin/behat -p integration"
+        "dc": [
+            "Composer\\Config::disableProcessTimeout",
+            "docker-compose -f ../../docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose.yml -f ../../../opg-sirius-api-gateway/docker-compose-integration.yml"
+        ],
+        "dc-run": "@dc run actor-app",
+        "dc-php": "@dc-run php",
```
williamfalconeruk, line 111: That's a nice feature of composer :)

service-front/app/src/Common/src/Service/Security/RateLimitServiceFactory.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        $config = $this->container->get('config');
+
+        if (!isset($config['ratelimits'])) {
+            throw new RuntimeException('Missing rate limits configuration');
+        }
+
+        return array_map(function (string $limitName) {
+            return $this->factory($limitName);
+        }, array_keys($config['ratelimits']));
```
hawx, line 81: I was hoping this could neaten to `array_map($this->factory, array_keys($config['ratelimits']))`, but I guess I am still thinking in JS everywhere.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -1,6 +1,11 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0">
     <option name="myName" value="Project Default" />
+    <inspection_tool class="PhpCSValidationInspection" enabled="true" level="WEAK WARNING" enabled_by_default="true">
```
hawx, line 4: :eyes: Is this the formatting stuff you mentioned on the other PR?

terraform/environment/viewer_ecs.tf
```diff
@@ -52,6 +52,25 @@ resource "aws_security_group_rule" "viewer_ecs_service_egress" {
   security_group_id = aws_security_group.viewer_ecs_service.id
 }
 
+// Brute force Elasticache
+resource "aws_security_group_rule" "viewer_ecs_service_elasticache_egress" {
+  type              = "egress"
+  from_port         = 0
```
andrewpearce-digital, line 58: This sg rule scope is covered by the anything out rule above it.

andrewpearce-digital: I think we should remove

service-api/app/src/App/src/ConfigProvider.php
cooperaj: Yeah, no idea whats up with it. The code obvs works without the change but PHPStorm also tells me it's used in the file. So... 🤷

service-front/app/composer.json
cooperaj: Yeah. Was fed up with all the docker-composer lines

.idea/inspectionProfiles/Project_Default.xml
cooperaj: Yeah. Though it's not automatic it at least prompts you in the file that your formatting isn't quite right.

terraform/environment/viewer_ecs.tf
cooperaj: ok. Didn't realise :)

service-front/app/src/Common/src/Service/Security/RateLimitServiceFactory.php
cooperaj: To be honest, much as I like the functional approach it is technically slower (micro optimisation ahoy) and is a little harder to read for new eyes.

service-front/docker/app/Dockerfile
```diff
@@ -2,14 +2,16 @@ FROM composer AS composer
 
 COPY service-front/app /app
 
-RUN composer install --prefer-dist --no-suggest --no-interaction --no-scripts --optimize-autoloader
+RUN composer install --prefer-dist --no-suggest --no-interaction --no-scripts --optimize-autoloader --ignore-platform-reqs
```
williamfalconeruk, line 5: QQ out of interest why was this needed? as learning all things docker :)

service-api/app/src/App/src/ConfigProvider.php
williamfalconeruk: seen this a few times recently, might need to go through again... 😕

service-front/docker/app/Dockerfile
cooperaj: I've added two system dependencies to the composer.json. This nails down in composer that we need these extensions to run the app. But this part of the installation process simply does the install and doesn't run the code so it doesn't *need* them. This flag just turns off that check.

It does mean that if anybody is attempt to run code using a local install of PHP rather than the docker-compose one they'll need to install the extensions (or use the same flag)

service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
mattmachell: > It would have to be generous to protect users behind proxy servers or NATs but would provide protection against an attacker doing this kind of header randomisation.

Yeah, it's a balancing act given our target audience are banks sitting behind proxies. As long as we plan to make the markers we track and the limiter configurable and modifiable later, I think this still gives us more than we had (which is what the ticket is for). 

There are infra level tools like Guard Duty we can also apply as well, but it's always good to have it in app.

Tangent, do you have a minimum time from initial form render to post set up on the form handlers? Can't recall if we did tickets for that.

.idea/.gitignore
```diff
@@ -1,2 +1,3 @@
 # Default ignored files
-/workspace.xml
\ No newline at end of file
+/workspace.xml
+/shelf/
```
andrewpearce-digital, line 3: end of line character needed here

service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
cooperaj: No. No minimum time is setup, I think the moving window approach covers that assuming accurate identification.
It's certainly something that we can implement, though I don't know about tickets.

PR #278, opened 2020-03-24
pass security group names in config file

andrewpearce-digital:
## Purpose
The python script for managing CI ingress is failing to remove security group rules.

Fixes UML-638

## Approach

Review and refine script

## Learning

## Checklist


andrewpearce-digital:
Adding and removing ingress working again
image

cooperaj:
I'm not 100% this has fixed the whole issue. 

We were seeing large numbers of sg rules *not* being deleted beacuse the behave tests were failing and the removal step wasn't being run. Won't this still be the case with the new code?

andrewpearce-digital:
> I'm not 100% this has fixed the whole issue.
> 
> We were seeing large numbers of sg rules _not_ being deleted beacuse the behave tests were failing and the removal step wasn't being run. Won't this still be the case with the new code?

The job to remove ingress is set to always run even if the previous step fails.
```          - run:
              name: Remove CircleCI ingress to environment
              command: |
                export TF_WORKSPACE=$(bash ~/project/pipeline_scripts/set_environment_variables/set_workspace.sh <<parameters.workspace>>) >> $BASH_ENV
                echo $TF_WORKSPACE
                python pipeline_scripts/ci_ingress.py /tmp/cluster_config.json
              when: always
```

The reason the script wasn't removing ci ingress was it was assuming that the first set of rules would contain the ingress to be removed. It used to but an update to the AWS API changed that a bit.

The update to the script iterates over all sets of rules to find and remove ingress, so we shouldn't get this issue return.

The script output shows that it's removing ingress rules to show it's working. Before this change, there was no such output when there should have been.

andrewpearce-digital:
An additional behaviour of this script is that it was supposed to find and remove any existing ingress rules when adding new ones as a bit of a catchall, so the next run of this script on preproduction and production to add ingress will remove all the previous ingress rules since I last manually removed them.

andrewpearce-digital:
hmm, that said, if the rule happens to be in the port 80 set of permissions there will be an error.

I'm going to make another update to the script to handle that

andrewpearce-digital:
image

PR #282, opened 2020-03-27
Log reasons validation of code details failed

hawx:
## Purpose
Log failures when validating a code to add to an LPA so problems can be debugged/monitored/etc.

Fixes [UML-603](https://opgtransform.atlassian.net/browse/UML-603)

## Approach

It adds logging. Maybe it is too defensive?

## Learning

## Checklist

### DIFF
--- a/service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
+++ b/service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
@@ -6,6 +6,7 @@ namespace App\Service\ActorCodes;
 
 use App\DataAccess\Repository;
 use App\Service\Lpa\LpaService;
+use Psr\Log\LoggerInterface;
 use Ramsey\Uuid\Uuid;
 
 class ActorCodeService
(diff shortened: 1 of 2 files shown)


hawx:
Does that exception handler do what we want? If _any_ problem occurs then delete the map we just created seems weird.

hawx:
Also I hate that UUID loop, the whole point of UUIDs is that in practice you won't see a collision.

service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
```diff
@@ -50,48 +59,66 @@ public function __construct(
      */
     public function validateDetails(string $code, string $uid, string $dob): ?array
     {
-        //-----------------------
-        // Lookup the LPA and the Actor's LPA ID.
-
         $details = $this->actorCodesRepository->get($code);
 
         if (is_null($details)) {
+            $this->logger->info('Validating code could not find details for code {code}', ['code' => $code]);
```
mattmachell: I'd expect these to be logged at **error** rather than **info** level, given that they occur when an error occurs that results in no data. Unless there's a pattern the team have agreed on not to do this, of course.

hawx: Hmm, yeah interesting. I've changed it so that some of them log at error level because we would expect the data in those cases to be there. But the others are comparing data from a user so could just be mistypings, which I don't think we should class as "errors". 

If anyone else on the team has opinions on this it would be cool to hear them.

williamfalconeruk: seems like a reasonable assumption. Let me review those changes to confirm that looks ok

williamfalconeruk:
> Does that exception handler do what we want? If _any_ problem occurs then delete the map we just created seems weird.

I think it's because we don't know what those exceptions might be, hence the catch-all.
This is a cleanup thing so we do not end up with orphaned records, not ideal but at least keeps the data consistent.  If we can capture what the exceptions are, via logging, we could change this to be more specific, and handle them more gracefully. Maybe a tech debt item for later?

PR #285, opened 2020-03-30
Added security headers to nginx.conf

Raffers:
## Purpose
Adds headers to all succesful HTTP requests, for security hardening

Fixes UML-343

## Approach

## Learning

## Checklist


service-front/docker/web/etc/confd/templates/nginx.conf
```diff
@@ -22,6 +22,13 @@ server {
     gzip_vary on;
     gzip_types text/plain text/css application/json application/javascript text/xml application/xml text/javascript image/svg+xml;
 
+    # add security headers
+    #
+    add_header X-XSS-Protection: 1; mode=block;
```
hawx: I think the problem is that these shouldn't have `:` in them, the story wasn't exactly copy paste. So it would be like `add_header X-XSS-Protection "1; mode=block";`, etc.

hawx: Ha, if you look at the image on the story and not the text it shows the correct incantations

williamfalconeruk: good spot @hawx!

PR #288, opened 2020-04-01
Ensure actor codes are used as uppercase

hawx:
## Purpose
We are not treating codes as case-insensitive when looking them up to add LPAs.

(Partially) fixes [UML-621](https://opgtransform.atlassian.net/secure/RapidBoard.jspa?rapidView=149&projectKey=LPA&modal=detail&selectedIssue=UML-621)

## Approach

It changes the data access layer to uppercase codes before calling DynamoDB.

## Learning

## Checklist

### DIFF
--- a/service-front/app/features/actor-add-an-lpa.feature
+++ b/service-front/app/features/actor-add-an-lpa.feature
@@ -10,11 +10,17 @@ Feature: Add an LPA
     And I have been given access to use an LPA via credentials
 
   @integration @ui
-  Scenario: The user can add an LPA to their account
+  Scenario Outline: The user can add an LPA to their account
     Given I am on the add an LPA page
-    When I request to add an LPA with valid details
+    When I request to add an LPA with valid details using <passcode>
     Then The correct LPA is found and I can confirm to add it
     And The LPA is successfully added
+
+    Examples:
+      | passcode       |
+      | xyuphwqrechv   |
+      | XYUPHWQRECHV   |
+      | XYUP-hwqr-EcHv |
 
   @integration @ui
   Scenario: The user cannot add an LPA to their account as it does not exist
--- a/service-front/app/features/context/Integration/AccountContext.php
+++ b/service-front/app/features/context/Integration/AccountContext.php
@@ -378,9 +378,9 @@ class AccountContext extends BaseIntegrationContext
     }
 
     /**
-     * @When /^I request to add an LPA with valid details$/
+     * @When /^I request to add an LPA with valid details using (.*)$/
      */
-    public function iRequestToAddAnLPAWithValidDetails()
+    public function iRequestToAddAnLPAWithValidDetailsUsing(string $code)
     {
         // API call for checking LPA
         $this->apiFixtures->post('/v1/actor-codes/summary')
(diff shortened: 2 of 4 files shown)


service-api/app/test/AppTest/DataAccess/DynamoDb/ActorCodesTest.php
```diff
@@ -26,13 +26,14 @@ protected function setUp()
     /** @test */
     public function can_lookup_a_code()
     {
-        $testCode = 'test-code';
+        $testCode = 'test-code-1234';
```
williamfalconeruk, line 29: can't see this lowercase test code being used in the test. either it should be used as an input or be removed.

service-api/app/test/AppTest/DataAccess/DynamoDb/ActorCodesTest.php
```diff
@@ -111,9 +113,10 @@ public function cannot_lookup_a_missing_code()
     /** @test */
     public function can_make_a_code_as_used()
     {
-        $testCode = 'test-code';
+        $testCode = 'test-code-1234';
```
williamfalconeruk, line 116: can't see this lowercase test code being used in the test. either it should be used as an input or be removed

service-api/app/test/AppTest/DataAccess/DynamoDb/ActorCodesTest.php
hawx: https://github.com/ministryofjustice/opg-use-an-lpa/pull/288/files/1f0f30c90133e49f7923b4deb30f3f1d6792ceff#diff-c58ea8d36cdae2f908439dbf17bdf533R71 GitHub has helpfully hid the bit where it is used

service-api/app/src/App/src/DataAccess/DynamoDb/ActorCodes.php
```diff
@@ -41,7 +41,7 @@ public function get(string $code): ?array
             'TableName' => $this->actorLpaCodesTable,
             'Key' => [
                 'ActorCode' => [
-                    'S' => $code,
+                    'S' => strtoupper($code),
```
GemTay, line 44: filtering should be handled much earlier on, within the form - add a stringToUpper filter to the passcode field 👍

hawx: Yes! I had forgotten about the forms, much easier.

service-front/app/features/actor-add-an-lpa.feature
```diff
@@ -16,6 +16,13 @@ Feature: Add an LPA
     Then The correct LPA is found and I can confirm to add it
     And The LPA is successfully added
 
+  @ui
+  Scenario: The user can add an LPA to their account
+    Given I am on the add an LPA page
+    When I request to add an LPA with valid details using a lowercase code
```
GemTay, line 22: This is perfectly fine, but I've just though... you could create a scenario outline for the "The user can add an LPA to their account" scenario, and different test cases since the user can enter the code in all different forms such as uppercase, lowercase, mixed upper & lower, with or without spaces, and with or without dashes

hawx: I could, I have no strong feelings either way. Tie break @williamfalconeruk?

GemTay: It would reduce adding more code, and checking we accept all allowed formats, i think it'd be useful

williamfalconeruk: actually agree with @GemTay  on this - lets cover those scenarios

PR #291, opened 2020-04-03
Uml 640 fix account password change

Raffers:
## Purpose
Password change email still being sent when password was failed to change, along with providing no error that user password could not be changed

Fixes UML-640

## Approach

Mistake in service-front class `userService` not passing Exception along when failing to validate user password

## Learning

## Checklist


service-front/app/src/Actor/src/Form/PasswordChange.php
```diff
@@ -112,6 +109,9 @@ public function getInputFilterSpecification()
                             ],
                         ],
                     ],
+                    [
+                        'name' => PasswordValidator::class,
```
hawx, line 113: We may want a story to change these validation messages as we get told about "your password" which is a bit ambiguous, especially if you get the too short message at the same time

image

PR #294, opened 2020-04-06
Uml 400 deleting account

GemTay:
## Purpose
Added account deletion feature, with logging and full integration testing

Fixes UML-400

## Approach

Used the user service to get the user's account ID, which is passed in the API call to delete the account by ID from the ActorUsers table. No other data is deleted.
I have added logging various places where I see necessary, and covered all scenario's in integration tests

## Checklist


service-api/app/src/App/src/Handler/ConfirmCurrentPassword.php
```diff
@@ -1,10 +0,0 @@
-<?php
-
-
-namespace App\Handler;
-
-
-class ConfirmCurrentPassword
-{
-
-}
```
GemTay, line 10: I found and deleted a blank class that had somehow got through code review...

service-front/app/config/routes.php
```diff
@@ -108,13 +118,6 @@
...
         Actor\Handler\CancelCodeHandler::class
     ],'lpa.cancel-code');
-    $app->get('/lpa/change-details', [
-        Mezzio\Authentication\AuthenticationMiddleware::class,
-        Actor\Handler\ChangeDetailsHandler::class
-    ], 'lpa.change-details');
-    $app->get('/lpa/terms-of-use', [
-        Actor\Handler\ActorTermsOfUseHandler::class
-    ], 'lpa.terms-of-use');
 };
```
GemTay, line 121: I organised/moved some of the routes into groups they were more suited to ^

service-api/app/src/App/src/Handler/ConfirmCurrentPassword.php
Raffers: Yup, that's my fault. Waiting for my current PR to be merged to also fix this.

service-front/app/src/Actor/templates/actor/confirm-delete-account.html.twig
```diff
@@ -0,0 +1,37 @@
...
+
+                    <p class="govuk-body">OPG can update the paper lasting power of attorney (LPA) and the online LPA summary so they show your new name.</p>
+
+                    <a href="{{ path('your-details')}}" role="button" draggable="false" class="govuk-button govuk-button govuk-button--secondary govuk-!-margin-right-1">
+                        No, return to my details
+                    </a>
+
+                    <a href="{{ path('delete-account') }}" role="button" draggable="false" class="govuk-button">
+                        Yes, continue deleting my account
+                    </a>
```
hawx, line 30: I read the story as saying the content should be

```
Are you sure you want to delete your account?

If you delete your Use a lasting power of attorney account:

   - any LPAs you added will be removed

   - you‘ll not be able to see how those LPAs are being used

If you want to use the LPA online after you've deleted your account, you’ll need to contact us.

You can still use paper versions of these LPAs if:

   - the LPAs are still valid

   - you have not been removed as an attorney

No, return to my details Yes, delete my account
```

am I looking in the wrong place?

service-front/app/src/Actor/src/Handler/DeleteAccountHandler.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $user = $this->getUser($request);
+
+        $this->userService->deleteAccount($user->getIdentity());
+
+        $session = $this->getSession($request, 'session');
+        $session->unset(UserInterface::class);
+        $session->regenerate();
+
+        $this->getLogger()->info(
```
hawx, line 72: Do we need to log successes?

service-api/app/src/App/src/DataAccess/DynamoDb/ActorUsers.php
```diff
@@ -258,4 +259,26 @@ public function recordPasswordResetRequest(string $email, string $resetToken, in
 
         return $this->getData($user);
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function delete(string $accountId) :bool
```
hawx, line 266: Do we care about cleaning up more than just the user? e.g. should codes that the user has generated also be removed

service-front/app/config/routes.php
```diff
@@ -64,16 +64,26 @@
     $app->route('/forgot-password', Actor\Handler\PasswordResetRequestPageHandler::class, ['GET', 'POST'], 'password-reset');
     $app->route('/forgot-password/{token}', Actor\Handler\PasswordResetPageHandler::class, ['GET', 'POST'], 'password-reset-token');
 
+    // User deletion
```
cooperaj, line 67: These don't appear to have authentication middleware attached?

service-api/app/src/App/src/Service/User/UserService.php
```diff
@@ -261,4 +259,32 @@ public function completeChangePassword(string $userId, string $password, string
...
+
+        if (is_null($user)) {
+            $this->logger->notice(
+                'Account not found for user Id {Id}',
+                ['Id' => $accountId]
+            );
+            throw new NotFoundException('User not found for account with Id ' . $accountId);
+        }
+
+        if (!$this->usersRepository->delete($accountId)) {
```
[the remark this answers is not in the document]
cooperaj: The logging for what the error was should then happen in the repository? The repository exists to abstract DB operations out of the service so I think doing this would muddle concerns

service-front/app/src/Actor/src/Handler/DeleteAccountHandler.php
cooperaj: We should be logging pretty much everything for audit purposes, though I'd question why this log entry is in the handler and not just in the deleteAccount method.

service-front/app/src/Actor/templates/actor/confirm-delete-account.html.twig
GemTay: My bad, i read the content off the prototype - fixed it now!

service-api/app/src/App/src/DataAccess/DynamoDb/ActorUsers.php
GemTay: Its just the user account data, any codes can still remain 👍

service-front/app/src/Common/src/Service/Log/EmailHash.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Common\Service\Log;
+
+class EmailHash
```
hawx, line 7: I think either this should just expose a hash function instead of a class; or it should become an `Email` class which is used everywhere an email is passed around, and then you'd need to add a method to expose the raw email too.

Or even more ideally, and with no idea how any of the frameworks or PHP work, you would create an `Email` class to wrap the email string, then teach the logging library to always hash objects of that type 🤷‍♂ I doubt that is possible though

hawx: or can we teach the logging library that params with the key 'email' are to be hashed 🤔 that would be great

mattmachell: Making it an email class and then making its __toString method do the hashing would save some of the boilerplate. Would mean you had to be explicit if you wanted to getAddress() from it as well.

service-front/app/src/Common/src/Service/Log/EmailHash.php
```diff
@@ -0,0 +1,22 @@
...
+    private $email;
+
+    public function __construct($email)
+    {
+        $this->email = $email;
+    }
+
+    public function hash($email)
+    {
+        $this->email = hash('sha256', $email);
```
mattmachell, line 18: Always worth adding a salt value to avoid any issues with rainbow table attacks.

service-front/app/src/Common/src/Service/Log/EmailHash.php
mattmachell: As a tangent, it's worth being familiar with the various magic methods PHP has for classes, as they provide lots of useful tools for this kind of situation. https://www.php.net/manual/en/language.oop5.magic.php

cooperaj: Overriding the toString will do what we want here.

service-front/app/src/Common/src/Service/Log/EmailHash.php
cooperaj: Just be sure to include that hash as a part of the output. Unlike LPA where we have a single non-rotatable hash in config :/

service-front/app/src/Common/src/Service/User/UserService.php
```diff
@@ -264,4 +265,31 @@ public function changePassword(string $id, string $password, string $newPassword
             );
         }
     }
+
+    public function deleteAccount(string $accountId, string $userEmail) : void
```
cooperaj, line 269: I'm unsure about this function needing an email for the express purpose of logging the delete request.

If the delete functionality needs that then this should first do a request to verify the accountId exists and then delete it.

service-api/app/src/App/src/Service/User/UserService.php
```diff
@@ -261,4 +259,22 @@ public function completeChangePassword(string $userId, string $password, string
 
         $this->usersRepository->resetPassword($userId, $newPassword);
     }
+
+    /**
+     * @param string $accountId
+     */
+    public function deleteUserAccount(string $accountId) : void
+    {
+        $user = $this->usersRepository->get($accountId);
```
cooperaj, line 268: Better yet, return this user when you do the delete operation. That way you know the details of the account you just deleted.

PR #295, opened 2020-04-08
UML-634 Inactive attorney journey

hawx:
## Purpose
Prevents inactive attorneys from viewing/accessing an LPA they have previously added to their account.

Fixes [UML-634](https://opgtransform.atlassian.net/secure/RapidBoard.jspa?rapidView=149&projectKey=LPA&modal=detail&selectedIssue=UML-634)

## Approach

It checks `systemStatus` on the actor returned by the API. On the dashboard view this shows as options being greyed out, I've also made it so that directly navigating to the pages which allow you to create/check/view stuff will 404.

## Learning

## Checklist


service-front/app/features/context/UI/AccountContext.php
```diff
@@ -2205,4 +2252,217 @@ public function iAmTakenToTheChangeDetailsPage()
...
+
+        $this->ui->assertPageAddress('/lpa/dashboard');
+
+        try {
+            $this->ui->clickLink('Give an organisation access');
+        } catch (ElementNotFoundException $e) {
+            return;
+        }
+
+        throw new AssertionFailedError('Expected not to click link: Give an organisation access');
```
hawx, line 2334: Is there a neater way of doing this assertion?

service-front/app/src/Actor/templates/actor/partials/lpa-details.html.twig
```diff
@@ -6,17 +6,29 @@
 
     <div class="govuk-summary-list__row govuk-!-margin-bottom-0 govuk-!-margin-top-4">
         <dt class="govuk-summary-list__key govuk-!-margin-bottom-0 govuk-!-width-one-third">
-            <span class="govuk-heading-m govuk-!-margin-bottom-0">{{ lpa.caseSubtype == 'pfa' ? 'Property and finance' : 'Health and welfare' }}</span>
+            {% if actorActive %}
+                <span class="govuk-heading-m govuk-!-margin-bottom-0">{{ lpa.caseSubtype == 'pfa' ? 'Property and finance' : 'Health and welfare' }}</span>
+            {% else %}
+                <span class="govuk-heading-m govuk-!-margin-bottom-0" style="color:#626a6e;">{{ lpa.caseSubtype == 'pfa' ? 'Property and finance' : 'Health and welfare' }}</span>
```
hawx, line 12: I've hard-coded the style here because I have no idea what the correct class(es?) would be, anyone know?

johnnolan: You can find an index of colours and other component classes here mate https://design-system.service.gov.uk/styles/colour/

johnnolan: It may also be worth bumping the logic for the caseSubType to the PHP if you can. Keep it all in one place rather than duplicating strings, will be easier when it comes to refactoring in the future. Maybe a new property like caseSubTypeDescription?

service-front/app/features/context/UI/AccountContext.php
cooperaj: Can you not assert that the link doesn't exist? rather than attempting to click i?

service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
```diff
@@ -64,6 +65,9 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         $identity = (!is_null($user)) ? $user->getIdentity() : null;
 
         $lpa = $this->lpaService->getLpaById($identity, $actorLpaToken);
+        if (is_null($lpa)) {
```
cooperaj, line 68: Does the underlying service not bubble up a 404 api exception already?

service-front/app/features/context/UI/AccountContext.php
```diff
@@ -91,6 +93,51 @@ public function iHaveBeenGivenAccessToUseAnLPAViaCredentials()
         ];
     }
 
+    /**
+     * @Given /^I am inactive against an LPA on my account$/
+     */
+    public function iAmInactiveAgainstAnLpaOnMyAccount()
```
cooperaj, line 99: How is this different to `iHaveBeenGivenAccessToUseAnLPAViaCredentials`?

It's ok to call another step from a step definition. Then alter just the specific status value

service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
cooperaj, line 68: The underlying api service will throw a 404 api exception in this case?

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -85,6 +86,10 @@ public function getLpaById(string $userToken, string $actorLpaToken): ?Lpa
 
         $lpaData = $this->apiClient->httpGet('/v1/lpas/' . $actorLpaToken);
 
+        if (!$lpaData['actor']['details']['systemStatus']) {
```
cooperaj, line 89: Ah. This is why the 404 exception stuff. 

I'd be hesitant to have this logic here. The method getLpaById should just be getting an LPA and not really caring that it is or isn't active for the user.

If this is something we decide we want to do then it should live in the API side of this call.

service-front/app/test/ActorTest/Handler/CreateViewerCodeHandlerTest.php
```diff
@@ -79,6 +80,8 @@ public function setUp()
         $this->requestProphecy = $this->prophesize(ServerRequestInterface::class);
         $this->sessionProphecy = $this->prophesize(SessionInterface::class);
         $this->lpaServiceProphecy = $this->prophesize(LpaService::class);
+        $this->lpaServiceProphecy->getLpaById(Argument::any(), Argument::any())
```
cooperaj, line 83: If you could nail down the requirements using `Argument::type('string')` I think that might be more explicit

service-front/app/test/ActorTest/Handler/LpaDashboardHandlerTest.php
```diff
@@ -99,10 +100,16 @@ public function dashboard_is_displayed_with_lpas_added()
             $this->viewerCodeServiceProphecy->reveal()
         );
 
+        $actor = new CaseActor();
```
cooperaj, line 103: The handler unit tests are deprecated across the project. Where work is done that touches upon them it is better to verify that the feature tests are more appropriate and then remove the test file completely.

Remember to mark the handler class as `@codCoverageIgnore`

service-front/app/features/context/UI/AccountContext.php
hawx: Do you know what the specific call for that would be? From looking I can't really see any good candidates, unless `assertElementNotOnPage` works differently to what I expect.

service-front/app/src/Common/src/Service/Lpa/LpaService.php
hawx: Sure, my thinking was that all the call sites only deal with "active" LPAs, but if that is the case maybe the function name would be misleading. I'll just push up the behaviour to the handlers.

service-front/app/features/context/UI/AccountContext.php
williamfalconeruk: @hawx @cooperaj - did we get a conclusion on this? I'm wondering if there is also a neater way to do this... also could possibly use `assertPageContainsText` - though not an exact match to what we need?  just a thought...

service-front/app/features/context/UI/AccountContext.php
```diff
@@ -2205,4 +2215,217 @@ public function iAmTakenToTheChangeDetailsPage()
...
+
+        $this->ui->assertPageAddress('/lpa/dashboard');
+
+        try {
+            $this->ui->clickLink('Check access codes');
+        } catch (ElementNotFoundException $e) {
+            return;
+        }
+
+        throw new AssertionFailedError('Expected not to click link: Check access codes');
```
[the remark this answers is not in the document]
hawx: is there any point spending more time thinking about this when what I have works?

service-front/app/src/Common/src/Exception/NotFoundException.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    const NOT_FOUND_CODE = 404;
+
+    /**
+     * InvalidRequestException constructor.
+     * @param string $message
+     * @param int $code
+     * @param \Exception|null $previous
+     */
+    public function __construct(string $message, int $code = self::NOT_FOUND_CODE, \Exception $previous = null)
```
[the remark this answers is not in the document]
hawx: This does seem like something that should be provided. I just copy-pasted how we'd done it for 400s.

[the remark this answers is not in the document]
hawx: I guess if we ended up needing more HTTP codes we could just define a `HttpException` that takes it as an arg.

service-front/app/features/context/UI/AccountContext.php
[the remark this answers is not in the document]
williamfalconeruk: if @cooperaj hasn't got anything on this we can close this off?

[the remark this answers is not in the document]
cooperaj: I think that we should probably find a more intuitive way to do this but since you've had a hunt we can live with it for now.

service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
```diff
@@ -63,7 +64,12 @@ public function handle(ServerRequestInterface $request): ResponseInterface
         $user = $this->getUser($request);
         $identity = (!is_null($user)) ? $user->getIdentity() : null;
 
-        $lpa = $this->lpaService->getLpaById($identity, $actorLpaToken);
+        $response = $this->lpaService->getLpaById($identity, $actorLpaToken);
```
cooperaj, line 67: The API should be throwing the 404 exception if we've decided that an inactive user on an lpa does not see it. 
For the most part we want all business type decisions there. If this were the case then we wouldn't need the NotFoundException either.

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -76,10 +77,10 @@ public function getLpas(string $userToken): ?ArrayObject
     /**
      * @param string $userToken
      * @param string $actorLpaToken
-     * @return Lpa|null
+     * @return array|null
      * @throws Exception
      */
-    public function getLpaById(string $userToken, string $actorLpaToken): ?Lpa
+    public function getLpaById(string $userToken, string $actorLpaToken): array
```
cooperaj, line 83: Would prefer not to 'downgrade' the return type by making it a mixed array. Moving this to the API would clear up quite a lot of the changes that have been made to move this here.

service-front/app/test/ActorTest/Handler/CreateViewerCodeHandlerTest.php
```diff
@@ -79,6 +80,8 @@ public function setUp()
         $this->requestProphecy = $this->prophesize(ServerRequestInterface::class);
         $this->sessionProphecy = $this->prophesize(SessionInterface::class);
         $this->lpaServiceProphecy = $this->prophesize(LpaService::class);
+        $this->lpaServiceProphecy->getLpaById(Argument::type('string'), Argument::type('string'))
+            ->willReturn(['lpa' => new Lpa(), 'actorActive' => true]);
```
cooperaj, line 84: This assertion is needed for all tests?

service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
hawx: Fine by me, it will clean up the changes I had to make for surfacing the data

service-front/app/features/context/UI/AccountContext.php
mattmachell: A typical way to do this is just to assert the link is not present on the page. And also to assert the page linked to cannot be reached or returns a not authorised message for the unauthorised user.

http://mink.behat.org/en/latest/guides/traversing-pages.html#selectors findAll returns an array of all occurrences on a page, which you can assert is zero. Use a css selector link a[href="/mylink-url"] rather than the text content of the link.

service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -95,20 +95,20 @@ public function getByUserLpaActorToken(string $token, string $userId): ?array
...
             return null;
         }
 
-        //---
-
         $lpaData = $lpa->getData();
         $actor = $this->lookupActorInLpa($lpaData, $map['ActorId']);
         unset($lpaData['original_attorneys']);
 
+        if (is_null($actor)) {
```
cooperaj, line 108: A line comment here to explain what just happened would be great :)

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -76,7 +76,7 @@ public function getLpas(string $userToken): ?ArrayObject
     /**
      * @param string $userToken
      * @param string $actorLpaToken
-     * @return Lpa|null
+     * @return array|null
```
cooperaj, line 79: This is now incorrect

service-api/app/src/App/src/Service/Lpa/LpaService.php
hawx: > The best comment is a good name for a method or class.

I renamed `lookupActorInLpa` to `lookupActiveActorInLpa` which makes it clearer, as it doesn't make sense to return a Lpa to an actor who is inactive on it 😄

service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -239,7 +239,7 @@ public function getByViewerCode(string $viewerCode, string $donorSurname, bool $
      * @param int $actorId
      * @return array|null
      */
-    public function lookupActorInLpa(array $lpa, int $actorId): ?array
+    public function lookupActiveActorInLpa(array $lpa, int $actorId): ?array
```
cooperaj, line 242: For some reason, I'm guessing it snuck through a prior request, there are a number of const's declared below this function. Consts are always declared at the top of the file as a part of our adopted style guides. Also, please mark the attorneyStatus function as private as there will never be a reason to inherit from this class.

service-front/app/src/Actor/src/Handler/CreateViewerCodeHandler.php
```diff
@@ -113,8 +113,6 @@ public function handle(ServerRequestInterface $request): ResponseInterface
 
     private function fetchLpa(string $userToken, string $actorLpaToken): ?Lpa
     {
-        $lpa = $this->lpaService->getLpaById($userToken, $actorLpaToken);
-
-        return $lpa;
+        return $this->lpaService->getLpaById($userToken, $actorLpaToken);
```
cooperaj, line 116: This function has now been simplified to the point where it is not required. Best to refactor it away.

cooperaj: (6 months ago it had actual logic in it)

PR #299, opened 2020-04-15
Update readme with details about `--no-cache` option in docker-compose

williamfalconeruk:
## Purpose
To update the readme of UaL so that developers are aware of the use of the `--no-cache` option when using `docker-compose` commands on this repo, to keep docker images up to date, mitigating any vulnerabilities found in containers. 

Fixes UML-539

## Approach
Documentation only change

## Learning
N/A
## Checklist

* [X] I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant


hawx:
I can approve this because it does what it should, but have we ever thought about using Wikis instead for this kind of thing? It always seems a waste to go through a build for a setup doc change, separating it out would get round this.

williamfalconeruk:
> I can approve this because it does what it should, but have we ever thought about using Wikis instead for this kind of thing? It always seems a waste to go through a build for a setup doc change, separating it out would get round this.

To be fair I've seen both ways used in industry. 

This format is used as this is an open-source project. A lot of OS repo's out there use this sort of format for their set up and is a quick way to get started for anyone pulling the repo, and they may not have access to Confluence for example.

One advantage here is that the readme is source controlled in GitHub so you can track the changes over time. I know you can do this with wiki tools potentially too.

Another thing we could do in the future is potentially modify the build pipeline so as to not run it fully when a readme is edited, but not sure of how we could do that. Open to suggestions (tech debt perhaps?)

README.md
```diff
@@ -16,22 +16,35 @@ All commands assume a working directory of `opg-use-my-lpa`.
...
+
+You can also use the `--no-cache` option on other docker-compose commands,
+to clear out previously cached images.  
+
+***Note:*** this can take several minutes to run.  
+
 To bring up the local environment
 
-```bash
+```shell script
```
hawx, line 41: I think `shell` and `bash` are just aliases. 

shell:
```shell
if [ -z $X ]; then; echo 'hi'; fi
```

bash:
```bash
if [ -z $X ]; then; echo 'hi'; fi
```

andrewpearce-digital: I think you need to get rid of `script` here. The syntax highlighting is done by ```shell. I couldn't find reference to ```shell script.

williamfalconeruk:
> I think script needs to be removed from ```shell script

PHP storm suggested this - will change now.

williamfalconeruk:
> > I think script needs to be removed from ```shell script
> 
> PHP storm suggested this - will change now.

it highlights in PHPStorm differently - this is not a major issue.

andrewpearce-digital:
> > I think script needs to be removed from ```shell script
> 
> PHP storm suggested this - will change now.

That's interesting. I've been searching for similar usage, and couldn't find it. All examples only said shell or bash or a similar alias.

PR #300, opened 2020-04-15
Add elasticache terraform configuration

cooperaj:
## Purpose
Introduce elasticache to the project to handle the brute force protection store

## Approach

New terraform configuration to setup the cache. Permissions onto the cache will be dealt with by the main PR.

## Checklist

### DIFF
--- a/terraform/account/elasticache.tf
+++ b/terraform/account/elasticache.tf
@@ -1,17 +1,10 @@
 resource "aws_elasticache_subnet_group" "brute_force_cache_subnet" {
   name       = "brute-force-cache-subnet"
-  subnet_ids = [aws_subnet.private[0].id]
-
-  tags = merge(
-    local.default_tags,
-    {
-      "Name" = "private"
-    },
-  )
+  subnet_ids = aws_subnet.private[*].id
 }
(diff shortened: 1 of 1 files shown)


terraform/account/elasticache.tf
```diff
@@ -0,0 +1,24 @@
+resource "aws_elasticache_subnet_group" "brute_force_cache_subnet" {
```
andrewpearce-digital, line 1: I think this could be called "private". It can be used on any number of elasticache clusters

```suggestion
resource "aws_elasticache_subnet_group" "private_subnets" {
```

terraform/account/elasticache.tf
```diff
@@ -0,0 +1,24 @@
...
+  name       = "brute-force-cache-subnet"
+  subnet_ids = [aws_subnet.private.id]
```
andrewpearce-digital, line 3: aws_subnet.private is actually 3 resources created with a count (see network.tf).

This means you can access each by index `aws_subnet.private[0].id`. I think you can also pull them all into your list with `aws_subnet.private[*].id`

PR #301, opened 2020-04-15
Acceptance and integration tests for login feature

GemTay:
## Purpose
Completes api login feature tests for all scenarios

Fixes UML-616

## Approach

Tests happy path, incorrect email/password, and inactive account scenarios

## Checklist


service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -125,6 +146,104 @@ public function iAmCurrentlySignedIn()
...
+        ]));
+
+        $us = $this->container->get(UserService::class);
+
+        try {
+            $us->authenticate($this->userAccountEmail, '1nc0rr3ctPa33w0rd');
+        } catch (ForbiddenException $fe){
+            assertEquals('Authentication failed for email ' . $this->userAccountEmail, $fe->getMessage());
+            assertEquals(403, $fe->getCode());
+        }
```
cooperaj, line 184: After here you'll need to add a call to fail() (something like that anyway) to ensure that the Forbidden exception is the only code path.

service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -125,6 +146,104 @@ public function iAmCurrentlySignedIn()
...
+        $this->awsFixtures->append(new Result([]));
+
+        $us = $this->container->get(UserService::class);
+
+        try {
+            $us->authenticate('incorrect@email.com', $this->userAccountPassword);
+        } catch (NotFoundException $ex) {
+            assertEquals('User not found for email', $ex->getMessage());
+            assertEquals(404, $ex->getCode());
+        }
```
cooperaj, line 210: same here

service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -125,6 +146,104 @@ public function iAmCurrentlySignedIn()
...
+        ]));
+
+        $us = $this->container->get(UserService::class);
+
+        try {
+            $us->authenticate($this->userAccountEmail, $this->userAccountPassword);
+        } catch (UnauthorizedException $ex) {
+            assertEquals('Authentication attempted against inactive account with Id ' . $this->userAccountId, $ex->getMessage());
+            assertEquals(401, $ex->getCode());
+        }
```
cooperaj, line 246: and here

service-api/app/features/context/Integration/AccountContext.php
williamfalconeruk: ah of course now I get it! sorry @cooperaj for missing that...

service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -125,6 +146,107 @@ public function iAmCurrentlySignedIn()
...
+        $this->awsFixtures->append(new Result([]));
+
+        $us = $this->container->get(UserService::class);
+
+        try {
+            $us->authenticate('incorrect@email.com', $this->userAccountPassword);
+        } catch (NotFoundException $ex) {
+            assertEquals('User not found for email', $ex->getMessage());
+            assertEquals(404, $ex->getCode());
+            exit();
```
cooperaj, line 211: No, not quite we don't want to exit the process. We need to tell behat that this test has failed if the exception *isn't* thrown. 

The two assertions you have will only trigger if the call throws the exception. What happens when it doesn't?

There is a behat call that allows you to fail the test (I believe, or it may just be throw a TestFailedException or sumsuch). This will need to be done as the last part of this test. Instead of this exit call you could then 'return'

cooperaj: All in all, testing exceptions in behat requires a little fiddling as it's not as clear as how PHPUnit works in this.

service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -245,8 +250,10 @@ public function iAmToldMyAccountHasNotBeenActivated()
         } catch (UnauthorizedException $ex) {
             assertEquals('Authentication attempted against inactive account with Id ' . $this->userAccountId, $ex->getMessage());
             assertEquals(401, $ex->getCode());
-            exit();
+            return;
         }
+
+        throw new ExpectationFailedException('Expected unauthorized exception was not thrown');
```
williamfalconeruk, line 256: ah yes - this is a similar pattern to what I'm used to seeing in tests when asserting correct exceptions.

PR #304, opened 2020-04-16
Handle dashboard status check when actor is the donor

hawx:
## Purpose

Fix donors who added LPAs being able to see the dashboard. 

Fixes UML-679

## Approach

The response for a donor doesn't seem to include `systemStatus`, which was naively being checked to see if the actor was active. I've changed it to check whether the actor is a donor first.

## Learning

## Checklist

### DIFF
--- a/test-data/api-gateway.json
+++ b/test-data/api-gateway.json
@@ -27,8 +27,7 @@
           "addressLine3": ""
         }
       ],
-      "companyName": null,
-      "systemStatus": true
+      "companyName": null
     },
     "applicationType": "Classic",
     "caseSubtype": "pfa",
@@ -152,8 +151,7 @@
           "addressLine3": ""
         }
       ],
-      "companyName": null,
-      "systemStatus": true
+      "companyName": null
     },
     "applicationType": "Classic",
     "caseSubtype": "pfa",
@@ -252,8 +250,7 @@
           "addressLine3": ""
         }
       ],
-      "companyName": null,
-      "systemStatus": true
+      "companyName": null
     },
     "applicationType": "Classic",
     "caseSubtype": "pfa",
@@ -401,8 +398,7 @@
           "addressLine3": ""
         }
       ],
-      "companyName": null,
-      "systemStatus": true
+      "companyName": null
     },
     "applicationType": "Classic",
     "caseSubtype": "pfa",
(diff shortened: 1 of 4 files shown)


test-data/api-gateway.json
```diff
@@ -27,8 +27,7 @@
           "addressLine3": ""
         }
       ],
-      "companyName": null,
-      "systemStatus": true
```
cooperaj, line 31: This data is what we get back from the sirius api. Have they changed that?

hawx: This is what I got when I got some earlier today

hawx: in my head it makes sense: the donor is always active

cooperaj: Guess they changed that then.

hawx: Yeah, I've had a look through the Sirius code and I think the swagger is wrong. It says that both [`donor`](https://github.com/ministryofjustice/opg-sirius/blob/e99cd97265a65609de53663547e6eeda59c3cb5b/back-end/docs/api/swagger/api.public.v1.yaml#L254) and [`attorneys`](https://github.com/ministryofjustice/opg-sirius/blob/e99cd97265a65609de53663547e6eeda59c3cb5b/back-end/docs/api/swagger/api.public.v1.yaml#L258) are `CaseActor`s, which should include [`systemStatus`](https://github.com/ministryofjustice/opg-sirius/blob/e99cd97265a65609de53663547e6eeda59c3cb5b/back-end/docs/api/swagger/api.public.v1.yaml#L325).

But the code eventually returns an `Lpa` where:
- the `donor` is of type [`Donor`](https://github.com/ministryofjustice/opg-sirius/blob/e99cd97265a65609de53663547e6eeda59c3cb5b/back-end/module/Opg/src/Opg/Core/Model/Entity/CaseActor/Donor.php), with no `systemStatus`
- but `attorneys` is a list of `Attorney`, extending [`AttorneyAbstract`](https://github.com/ministryofjustice/opg-sirius/blob/e99cd97265a65609de53663547e6eeda59c3cb5b/back-end/module/Opg/src/Opg/Core/Model/Entity/CaseActor/AttorneyAbstract.php#L33) which has `systemStatus`

PR #306, opened 2020-04-20
edited python script to seed test user into database

GemTay:
## Purpose
Seeds a test user account into the database to save dev's having to create a new account each time they bring up the containers 

## Checklist

### DIFF
--- a/service-api/seeding/dynamodb.py
+++ b/service-api/seeding/dynamodb.py
@@ -1,6 +1,7 @@
 import os
 import boto3
 import json
+import passlib
 from passlib.hash import sha256_crypt
 import datetime
 
--- a/service-api/seeding/dynamodb.py
+++ b/service-api/seeding/dynamodb.py
@@ -1,6 +1,8 @@
 import os
 import boto3
 import json
+from passlib.hash import sha256_crypt
+import datetime
 
 if 'AWS_ENDPOINT_DYNAMODB' in os.environ:
     # For local development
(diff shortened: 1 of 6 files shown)


service-api/seeding/dynamodb.py
```diff
@@ -113,3 +116,25 @@
...
     #print(json.dumps(response['Item'], indent=4, separators=(',', ': ')))
+
+# test user details
+
+actorUsersTable = dynamodb.Table(os.environ['DYNAMODB_TABLE_ACTOR_USERS'])
+
+actorUsers = [
+    {
+        'Id': 'bf9e7e77-f283-49c6-a79c-65d5d309ef77',
+        'Email': 'test.user@digital.justice.gov.uk',
```
andrewpearce-digital, line 127: I'd recommend using an email address we can get to if we ever want to be able to unlock an account on preprod etc.

I've been using something similar to opg-use-an-lpa+test-user@digital.justice.gov.uk. That's an email address we can get to if we need.

service-api/seeding/dynamodb.py
```diff
@@ -113,3 +116,25 @@
         Key={'ActorCode': i['ActorCode']}
     )
     #print(json.dumps(response['Item'], indent=4, separators=(',', ': ')))
```
andrewpearce-digital, line 118: should this still be commented out?

service-api/seeding/dynamodb.py
```diff
@@ -1,6 +1,9 @@
 import os
 import boto3
 import json
+import passlib
```
andrewpearce-digital, line 4: I think this line is redundant now that you're importing sha256 directly, and not calling passlib itself anywhere

service-api/seeding/dynamodb.py
GemTay: When uncommented it produces this error: 
```
TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Decimal is not JSON serializable
```
Just looking into it, I'm guessing it's the format of the expiry date thats causing the error... unsure just yet though

service-api/seeding/dynamodb.py
cooperaj: I'll just quietly remove my review... _hides_

GemTay:
Just tested login in the UAT env - it works! Horray!

PR #311, opened 2020-04-21
split circle ci workflow

GemTay:
## Purpose
Splits the workflow into pr_build and path_to_live with their relevant jobs

Fixes UML-576

## Approach

Duplicated existing path to live workflow renamed 'pr_build', and removed jobs that were not relevant to the pr build (master only filter). Repeated the opposite for the path to live workflow, removing irrelevant jobs containing ignore master filter

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,202 +19,197 @@ workflows:
           requires: [redeploy_demo]
 
   pr_build:
-    jobs:
-      # Common tasks
-      - use-my-lpa/lint_terraform:
-          name: lint_terraform
-          filters: { branches: { ignore: [master] } }
+    filters: { branches: { ignore: [master] } }
+      jobs:
+        # Common tasks
+        - use-my-lpa/lint_terraform:
+            name: lint_terraform
 
-      - use-my-lpa/node_test_web:
-          name: test_web
-      - use-my-lpa/node_build_web:
-          name: build_web
-          requires: [test_web]
-      - use-my-lpa/node_test_service_pdf:
-          name: test_pdf
+        - use-my-lpa/node_test_web:
+            name: test_web
+        - use-my-lpa/node_build_web:
+            name: build_web
+            requires: [test_web]
+        - use-my-lpa/node_test_service_pdf:
+            name: test_pdf
 
-      - use-my-lpa/docker_build_front_app:
-          name: app_front_build
-          requires: [build_web]
-      - use-my-lpa/docker_build_front_web:
-          name: web_front_build
-          requires: [build_web]
+        - use-my-lpa/docker_build_front_app:
+            name: app_front_build
+            requires: [build_web]
 
-      - use-my-lpa/docker_build_api_app:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -18,171 +18,224 @@ workflows:
...
+            name: dev_seed_database
+            filters: { branches: { ignore: [master] } }
+            requires: [dev_apply_environment_terraform]
+
+        - use-my-lpa/run_behave_suite:
+            name: dev_run_behave_tests
+            filters: { branches: { ignore: [master] } }
+            requires: [dev_seed_database]
+
+        - slack_notify_domain:
```
williamfalconeruk, line 110: just a thought - could dev environment destroy be split out and spawned as a separate job also? this could solve my overall / codecov issue where it doesn't think the job is finished.

.circleci/config.yml
```diff
@@ -18,171 +18,224 @@ workflows:
...
-          name: coveralls_upload
-          requires:
-            [app_front_build, app_api_build, test_web, test_pdf]
+
+  pr_build:
+      jobs:
+        # Common tasks
+        - use-my-lpa/lint_terraform:
+            name: lint_terraform
+            filters: { branches: { ignore: [master] } }
```
GemTay, line 27: is there a way to apply a filter to all jobs in a workflow without having to add it to each task individually?

.circleci/config.yml
williamfalconeruk: open to ideas though...

cooperaj: Seems like it'd fix that yes

andrewpearce-digital: I've got separate work that will take environment cleanup out of the pipeline totally

.circleci/config.yml
andrewpearce-digital: alas, no. The structure of the yaml requires it to be defined at each job no a workflow
https://circleci.com/docs/2.0/configuration-reference/#workflows

.circleci/config.yml
```diff
@@ -18,171 +18,224 @@ workflows:
...
-      - use-my-lpa/coveralls_upload:
-          name: coveralls_upload
-          requires:
-            [app_front_build, app_api_build, test_web, test_pdf]
+
+  pr_build:
+      jobs:
+        # Common tasks
+        - use-my-lpa/lint_terraform:
+            name: lint_terraform
```
andrewpearce-digital, line 26: I think the names of the common jobs need to to be made unique across the pr_build and path_to_live workflows

So `lint_terraform` here would be `lint_terraform_pr_build`

any names you update will need to be updated also in any `requires:` block

andrewpearce-digital: The reason I'm suggesting this is that if requires sees two jobs with the same name it can cause conflict, so this is intended to help us now and future people modifying the pipeline

andrewpearce-digital:
Looking good
image
image

PR #314, opened 2020-04-22
UML 647 - Adding logging to PDF download

Raffers:
## Purpose
I want to log all attempts to generate/download the LPA summary pdf

So that I can see how frequently users (viewers) are failing and succeeding to download the pdf 

Fixes UML-647

## Approach

Added logging class to DownloadLpaHandler & PdfService, logged response code were possible.
Added logging during successful/failed generation & when PDF has been presented for download.

## Learning

## Checklist


service-front/app/src/Common/src/Service/Pdf/PdfService.php
```diff
@@ -104,13 +112,25 @@ private function requestPdfFromService(string $htmlToRender): StreamInterface
...
+                    ]);
+
                     return $response->getBody();
                 default:
                     throw ApiException::create(null, $response);
             }
         } catch (ClientExceptionInterface $ex) {
             $response = ($ex instanceof HttpException) ? $ex->getResponse() : null;
 
+            $this->logger->debug(
```
mattmachell, line 127: Error is probably more suitable, as you may not have debug turned on in Prod and you still want to know if the service isn't generating PDFs.

service-front/app/src/Viewer/src/Handler/DownloadLpaHandler.php
```diff
@@ -36,13 +42,16 @@ class DownloadLpaHandler implements RequestHandlerInterface
...
      * ViewLpaHandler constructor.
      *
-     * @param LpaService $lpaService
-     * @param PdfService $pdfService
+     * @param LoggerInterface   $logger
+     * @param LpaService        $lpaService
+     * @param PdfService        $pdfService
      */
-    public function __construct(
+    public function __construct(,
```
williamfalconeruk, line 49: extra comma here, just needs removing.

service-front/app/src/Common/src/Service/Pdf/PdfService.php
williamfalconeruk: just wondering if the exception is captured further up the stack? if not then yep agree with this :+1

service-front/app/src/Common/src/Service/Pdf/PdfServiceFactory.php
```diff
@@ -31,6 +32,7 @@ public function __invoke(ContainerInterface $container, RequestedEntry $entityCl
         }
 
         return new PdfService(
+            $container->get(LoggerInterface::class),
```
cooperaj, line 35: Just for consistencies sake (and this is minor) typically the Logger is passed in last to a given service class - at least this is the pattern I've been following so far.

service-front/app/src/Viewer/src/Handler/DownloadLpaHandler.php
```diff
@@ -36,13 +42,16 @@ class DownloadLpaHandler implements RequestHandlerInterface
...
+        LoggerInterface $logger,
```
cooperaj, line 50: There's actually a trait and interface class setup for logging within handlers, should it be needed.

service-front/app/src/Common/src/Service/Pdf/PdfServiceFactory.php
williamfalconeruk: I think it's causing unit tests to fail also - so probably needs changing too...

service-front/app/src/Viewer/src/Handler/DownloadLpaHandler.php
cooperaj: Oh, you'll still need to do some work to get logging to happen. I recommend taking a look at the logout handler of the actor side for an example.

Raffers: After looking at it a bit more, I ended up removing this part of the logging. I don't think it really adds anything that isn't logged during PDF generation, unless we want to know that the session timed out before generating the PDF?

Raffers: Having just thought, we would want to log this. Due to clicking download giving no sort of redirection. I'll adjust this now.

service-front/app/src/Common/src/Service/Pdf/PdfService.php
Raffers: Changed this to error & the exception isn't passed further up.

service-front/app/src/Common/src/Service/Pdf/PdfService.php
```diff
@@ -104,13 +112,28 @@ private function requestPdfFromService(string $htmlToRender): StreamInterface
...
 
+            $this->logger->error(
+                'Failed to generate PDF from service {code} {response}',
+                [
+                    'code'      => $ex->getCode(),
+                    'response'  => $response
+                ]
+            );
+
             throw ApiException::create('Error whilst making http POST request', $response, $ex);
```
mattmachell, line 137: Might be worth being more specific in the error message here, i.e. to make it clear which service a post failed to.

PR #330, opened 2020-05-01
Changes to add only registered lpa's by actor

SeemaMenon:
## Purpose
https://opgtransform.atlassian.net/browse/UML-658

Fixes UML-658

## Approach

Going forwards only a registered LPA can be added by actor. Introduced a check in the handler to check the status prior to confirming addition of an LPA. Thereby a viewer code can only be generated for registered LPA's being added and henceforth viewers would only see Registered applications as Valid.

## Learning

## Checklist

* [X] I have added relevant logging with appropriate levels to my code


service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
```diff
@@ -127,24 +127,36 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
-                        'lpa'      => $lpa,
-                        'user'     => $user,
+                        'form' => $form,
+                        'lpa' => $lpa,
+                        'user' => $user,
                         'userRole' => $userRole,
                     ]));
+                } else {
+                    $this->getLogger()->debug(
+                        'Account with Id {id} has found an LPA with Id {uId} and LPA status is {status} and hence cannot be added',
```
GemTay, line 149: This log message could do with being shortened/simplified a little

SeemaMenon: Done  :)

service-front/app/src/Viewer/src/Handler/CheckCodeHandler.php
```diff
@@ -74,17 +76,18 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                 if ($lpa instanceof ArrayObject) {
                     // Then we found a LPA for the given code
                     $expires = new DateTime($lpa->expires);
-
                     if (isset($lpa->cancelled)) {
                         return new HtmlResponse($this->renderer->render('viewer::check-code-cancelled'));
-                    } else {
+                    } elseif (strtolower(($lpa->lpa)->getStatus()) == 'registered') {
```
GemTay, line 81: I'm unsure if this is the best way/place to do this. Its a perfectly valid solution ofcourse, however I'm wondering whether its the "best" solution. I feel this should be filtered further up the chain in the API layer, similar to how we filter out ghost/inactive attorneys. I'd like to hear what others opinions are on this for "best practice"

SeemaMenon: Also to mention, this scenario mostly will never happen on viewer and is added to further ensure non registered LPA's are not shows to viewers.

As an actor user,  I would no more be able to add an lpa [and generate code for viewers] if it's not registered.

service-front/app/features/actor-add-an-lpa.feature
```diff
@@ -9,6 +9,17 @@ Feature: Add an LPA
     And I am currently signed in
     And I have been given access to use an LPA via credentials
 
+  @ui
+  Scenario Outline: The user cannot add an LPA to their account when status is pending
```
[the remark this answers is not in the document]
SeemaMenon: changed to standard

service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
```diff
@@ -127,24 +127,35 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                     ]
                 );
 
-                if (!is_null($lpa)) {
+                if (!is_null($lpa) && (strtolower($lpa->getStatus()) == 'registered')) {
```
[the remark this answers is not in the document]
SeemaMenon: Corrected

service-front/app/src/Viewer/src/Handler/CheckCodeHandler.php
```diff
@@ -70,14 +72,14 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
-
                 if ($lpa instanceof ArrayObject) {
                     // Then we found a LPA for the given code
                     $expires = new DateTime($lpa->expires);
 
                     if (isset($lpa->cancelled)) {
                         return new HtmlResponse($this->renderer->render('viewer::check-code-cancelled'));
-                    } else {
+                    }
+                    if (strtolower(($lpa->lpa)->getStatus()) == 'registered') {
```
GemTay, line 82: Another place to change to "===" following Adam's comment

service-front/app/src/Viewer/src/Handler/CheckCodeHandler.php
SeemaMenon: done

service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -209,6 +210,11 @@ public function getByViewerCode(string $viewerCode, string $donorSurname, bool $
             throw new GoneException('Share code expired');
         }
 
+        if (isset($viewerCodeData['Cancelled'])) {
+            $this->logger->info('The code {code} entered by user is cancelled.', ['code' => $viewerCode]);
+            throw new BadRequestException('Access code cancelled');
```
cooperaj, line 215: We tend to use BadRequest for situations where the caller has made an error (missing params etc) I think The Gone exception is probably the only one we can use but we'll need to have a think how we separate that out on the frontend to provide the different responses.

service-front/app/features/view-lpa-using-sharecode.feature
```diff
@@ -17,7 +17,7 @@ Feature: View an LPA via sharecode
     Given I have been given access to a cancelled LPA via share code
     And I access the viewer service
     And I give a valid LPA share code
-    When I confirm the LPA is correct
+    When I confirm the LPA is correct but cancelled
     Then I can see the full details of a cancelled LPA
```
cooperaj, line 21: Wait. So even if it is cancelled we still allow them to view it?

service-front/app/src/Viewer/src/Handler/CheckCodeHandler.php
```diff
@@ -91,6 +89,9 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                 if ($apiEx->getCode() == StatusCodeInterface::STATUS_GONE) {
                     return new HtmlResponse($this->renderer->render('viewer::check-code-expired'));
                 }
+                if ($apiEx->getCode() == StatusCodeInterface::STATUS_BAD_REQUEST) {
```
cooperaj, line 92: Yeah, this is nicer. Though I don't know if this is the correct status code we should be using.

cooperaj:
> I think it's looking pretty nice now. We need to have a thing about how we doing the exceptions. The only real valid code for the cancelled is still 410 (Gone) so I think we need to look adding additional data to the response body that includes why it has been marked 410. 

Maybe write that up as a new ticket and we can let this through as debt for now.

service-front/app/features/view-lpa-using-sharecode.feature
SeemaMenon: We don't show the LPA details..but we just show the status has been cancelled message. The step wording is inappropriate. Shall correct as part of this now.

service-api/app/src/App/src/Service/Lpa/LpaService.php
SeemaMenon: Changed to use Gone exception and now depending on the exception message to  return the right http response. This will still need refactoring. new ticket created

service-front/app/src/Viewer/src/Handler/CheckCodeHandler.php
SeemaMenon: changed to use Gone

SeemaMenon:
> > I think it's looking pretty nice now. We need to have a thing about how we doing the exceptions. The only real valid code for the cancelled is still 410 (Gone) so I think we need to look adding additional data to the response body that includes why it has been marked 410.
> 
> Maybe write that up as a new ticket and we can let this through as debt for now.

SSM: New ticket created - https://opgtransform.atlassian.net/browse/UML-794

PR #332, opened 2020-05-04
UML-726 Behat smoke test

cooperaj:
## Purpose
This PR changes our smoke test suite from behave to behat. It utilises CircleCI containers with browsers embedded to run behat tests under chrome. 

## Approach

It's possible to run the tests locally by attaching the docker-compose.testing.yml file to your list of compose files and running 'smoke-test'

## Learning

Had to write a behat extension to get chrome and xvfb running as a part of the suite.

## Checklist


docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "2.1"
+version: "2.4"
```
williamfalconeruk, line 1: we've had some issues with this and the gateway before on up having a version mismatch. I guess this is ok in terms of versions, as it's a minor version?

cooperaj: no. quite right. 2.4 is needed to get phpstorms running of the tests working. i've made local changes to my api-gateway to have this work.

BUT

I think i've now worked around the issue so this isn't needed.

Still, probably should upgrade the version at some point right.

tests/smoke/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+```shell script
+# ensure your use-an-lpa environment is running already
+# use PHPStorm or a method devised in the top level README
+
+# run tests when wanted
+$ composer behat
+
+# create feature step definitions 
+$ composer behat -- --snippets-for
```
williamfalconeruk, line 12: so this creates the step definitions (i guess as empty functions) - that's good to know 👍

cooperaj: it spits them out on the console at the end of the run so you can copy and paste into a context file

You can of course always use PHPstorm to do it.

cooperaj:
Moving to a different PR

PR #333, opened 2020-05-04
UML-706 & UML-704 Created exception tests for ActorCodeConfirm & ActorCodeSummary

Raffers:
## Purpose
Creates exception path behat tests for both ActorCodeConfirm & ActorCodeSummary

Fixes UML-706/UML-704

## Approach

## Learning

## Checklist


service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -1912,4 +1912,237 @@ public function iAmLoggedOutOfTheServiceAndTakenToTheIndexPage()
...
+        // ActorCodes::get
+        $this->awsFixtures->append(new Result([]));
+
+        $this->apiFixtures->get('/v1/use-an-lpa/lpas/' . $this->lpaUid)
+            ->respondWith(
+                new Response(
+                    StatusCodeInterface::STATUS_BAD_REQUEST
+                )
+            );
+
```
GemTay, line 1929: Since the actor code attribute is missing from the data, the ActorCodes::get and the call to the API gateway above in this step wont be executed

service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -1912,4 +1912,237 @@ public function iAmLoggedOutOfTheServiceAndTakenToTheIndexPage()
...
+        $this->apiPost('/v1/actor-codes/summary', [
+            'actor-code' => null,
```
GemTay, line 1931: I dont think it's possible for the data passed into the api call to ever be null, as it would have been caught in the form validation - perhaps mock the response from $data = $request->getParsedBody(); in the ActorCodeSummaryHandler to have a null value

service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -1912,4 +1912,237 @@ public function iAmLoggedOutOfTheServiceAndTakenToTheIndexPage()
...
+        // ActorCodes::get
+        $this->awsFixtures->append(new Result([]));
+
+        $this->apiFixtures->get('/v1/use-an-lpa/lpas/' . $this->lpaUid)
+            ->respondWith(
+                new Response(
+                    StatusCodeInterface::STATUS_BAD_REQUEST
+                )
+            );
+
```
GemTay, line 1953: Same  two comments I made on the previous step to be applied this this step and the step below

service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -1912,4 +1912,237 @@ public function iAmLoggedOutOfTheServiceAndTakenToTheIndexPage()
...
+        ], [
+            'user-token' => $this->userLpaActorToken
+        ]);
+    }
+
+    /**
+     * @Given /^A malformed confirm request is sent which is missing date of birth$/
+     */
+    public function aMalformedConfirmRequestIsSentWhichIsMissingDateOfBirth()
+    {
```
GemTay, line 1991: Same issue repeated on these steps, these api and database calls aren't executed

service-api/app/features/context/Acceptance/AccountContext.php
Raffers: It wouldn't be, but these tickets UML-706 & UML-704 are both for exception path testing on the mentioned handlers. Also due to `isset` being used instead of `empty` if you pass an empty string it won't throw an exception.

service-api/app/features/context/Acceptance/AccountContext.php
```diff
@@ -1913,4 +1913,207 @@ public function iAmLoggedOutOfTheServiceAndTakenToTheIndexPage()
...
+        $this->awsFixtures->append(new Result([]));
+
+        $this->apiPost('/v1/actor-codes/confirm', [
+            'actor-code' => $this->oneTimeCode,
+            'uid'        => $this->lpaUid,
+            'dob'        => null
+        ], [
+            'user-token' => $this->userLpaActorToken
+        ]);
+
```
GemTay, line 2001: This post request should be the only code in this step, same for the other steps titled 'aMalformedConfirmRequestIsSentWhichIsMissing...'. No database/api calls are made as the bad request exception is thrown almost immediately inside the ActorCodeConfirmHandler

Raffers: Ah, I thought all the previous steps were required for the confirmation call. I'll adjust this now.

service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -1769,4 +1769,225 @@ public function iAmLoggedOutOfTheServiceAndTakenToTheIndexPage()
...
         // Not needed in this context
     }
+
+    /**
+     * @When /^I request to add an LPA with a missing actor code$/
+     */
+    public function iRequestToAddAnLPAWithAMissingActorCode()
+    {
+        // ActorCodes::get
+        $this->awsFixtures->append(new Result([]));
```
GemTay, line 1779: You can remove the get request too for all these steps

service-api/app/features/actor-add-an-lpa.feature
```diff
@@ -35,3 +35,45 @@ Feature: Add an LPA
...
+    When I request to add an LPA with a missing date of birth
+    Then The LPA is not found and I am told it was a bad request
+    And I request to go back and try again
+
+  @integration @acceptance
+  Scenario: The user cannot add an LPA to their account due to missing date of birth in confirmation
+    Given I am on the add an LPA page
+    When I request to add an LPA with valid details
+    Then The correct LPA is found and I can confirm to add it
+    And A malformed confirm request is sent which is missing date of birth
```
GemTay, line 65: I think these scenario outlines for the confirm request could be re written a little better, perhaps something like:
Given the LPA I want to add has been found
When I confirm to add it
Then A malformed confirm request is sent which is missing a date of birth 
And The LPA is not added and I am told it was a bad request

service-api/app/features/context/Integration/AccountContext.php
```diff
@@ -1769,4 +1769,225 @@ public function iAmLoggedOutOfTheServiceAndTakenToTheIndexPage()
...
+            ]
+        ]));
+
+        // ActorCodes::flagCodeAsUsed
+        $this->awsFixtures->append(new Result([]));
+
+        $actorCodeService = $this->container->get(ActorCodeService::class);
+
+        try {
+            $response = $actorCodeService->confirmDetails($this->passcode, $this->lpaUid, '', (string) $this->actorLpaId);
```
GemTay, line 1876: you dont need the try catch block here, as when the actorCodeService function returns null, it will not be returning to the ActorCodeConfirmHandler where an exception would be thrown

PR #341, opened 2020-05-07
Code generator instructions

andrewpearce-digital:
## Purpose
Document the full process for generating codes

## Approach

Update the README file for the script.
Update the script to prefix the environment name when requesting log streams
Provide a script to create an encrypted disk image to securely share generated codes

## Learning

There is a limit on the size of input you can pass into a bash script
https://medium.com/@chaitanya/4096-limit-c6f3ca88bcbf

## Checklist


scripts/generate_actor_codes/.envrc
```diff
@@ -1,6 +1,6 @@
 # development
-export AWS_ACCOUNT_ID=367815980639
+# export AWS_ACCOUNT_ID=367815980639
 # preproduction
 # export AWS_ACCOUNT_ID=888228022356
 # production
-# export AWS_ACCOUNT_ID=690083044361
+export AWS_ACCOUNT_ID=690083044361
```
williamfalconeruk, line 6: production is set as default - not sure we want this...

scripts/generate_actor_codes/make_encrypted_image.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+function make_encrypted_image() {
+
+  hdiutil create -size 2048k -fs HFS+ -encryption AES-256 -volname $FILENAME /tmp/$FILENAME/
```
cooperaj, line 5: Is this a 2MB disk image for a 4KB file?

andrewpearce-digital: If you set too small a size it fails to create the image
https://stackoverflow.com/questions/34119341/hdiutil-create-failed-error-5342

I decided to make a fixed size that wouldn't ever be a problem

andrewpearce-digital: When you set the file size to 512k you get this
image
stepping through the functions, you get the error above

andrewpearce-digital: reduced to 1MB

scripts/generate_actor_codes/.envrc
andrewpearce-digital: I've removed the need for this now.

Manually changing a file was inefficient.

Now the account ID will be set based on the environment we're creating codes for.

PR #350, opened 2020-05-13
UML-751 As a user I want my code to be accepted if I entered the C- or V- 

Raffers:
## Purpose
As a user (actor)
I want my valid actor code to be accepted even if I have included the 'C-' at the start
So that I can add my LPA to my account

As a user (viewer)
I want my valid viewer code to be accepted even if I have included the ‘V-’ at the start
So that I can view the LPA summary

Fixes UML-751

## Approach

Adjusted RegEx & applied additional filtering to remove C- or V-

## Learning

## Checklist


service-front/app/features/actor-add-an-lpa.feature
```diff
@@ -44,10 +45,9 @@ Feature: Add an LPA
 
     Examples:
       | passcode | reason |
-      | T3ST PA22C0D3 | Your activation key must only include letters, numbers and dashes |
-      | T3ST PA22-C0D3 | Your activation key must only include letters, numbers and dashes |
+      | T3ST$PA22-C0D3 | Your activation key must only include letters, numbers and dashes |
       | T3STP*22C0!? | Your activation key must only include letters, numbers and dashes |
-      | T3ST - PA22 - C0D3 | Your activation key must be 12 numbers and letters long |
+      | C - T3S* - PA22 - C0D3 | Your activation key must only include letters, numbers and dashes |
```
williamfalconeruk, line 50: Are both these invalid? i think we should test for both scenarios here (the old and the new)

service-front/app/features/view-lpa-using-sharecode.feature
```diff
@@ -45,7 +45,7 @@ Feature: View an LPA via sharecode
 
     Examples:
       | accessCode | reason |
-      | T3ST ACE2-C0D3 | LPA access codes are 13 numbers and letters long and start with a V |
+      | V-T3ST_ACE2-C0D3 | LPA access codes are 13 numbers and letters long and start with a V |
```
williamfalconeruk, line 48: I think we should test for both scenarios here old and new.

service-front/app/features/view-lpa-using-sharecode.feature
```diff
@@ -88,8 +88,8 @@ Feature: View an LPA via sharecode
     When I give an invalid <sharecode> and <surname>
     Then I am told that the share code is invalid because <reason>
     Examples:
-      |surname  |sharecode       | reason |
-      | Billson |1110-1111-0111  | We could not find an LPA matching those details |
+      |surname  | sharecode       | reason |
+      | Billson | V-1110-1111-0111  | We could not find an LPA matching those details |
```
williamfalconeruk, line 92: I think the old scenario value is also valid - please check if this is the case. (I think the V- is optional?)

service-front/app/features/actor-add-an-lpa.feature
Raffers: I changed this scenario as the error would no longer trigger with that error message. As my regEx changes now correctly account for spaces & dashes. This was already apart of the regEx but wasn't working correctly. This is the same for the other scenario change.

service-front/app/features/view-lpa-using-sharecode.feature
Raffers: The V- is now optional, it wasn't before. It's stripped out before it is checked.

service-front/app/features/view-lpa-using-sharecode.feature
```diff
@@ -88,8 +89,10 @@ Feature: View an LPA via sharecode
...
+      | Billson | V - 1110 - 1111 - 0111  | We could not find an LPA matching those details |
+      | Billson | 1110-1111-0111  | We could not find an LPA matching those details |
```
SeemaMenon, line 96: Is this 'V - 1110 - 1111 - 0111' not a valid entry? Thought the AC was to allow  'V-' in case users copy/oasted from the letter

Raffers: It is a valid entry, but it isn't a valid LPA. I added the V here just to make sure it was covered in the test.

PR #357, opened 2020-05-15
UML-805 Make code generation a single interactive script

williamfalconeruk:
## Purpose

- Using a single script, automatically:
    - Extract and format JSON data from the actor code generation task.
    - Generate an encrypted disk image file to send to the business.  

- Add suggestions from @cooperaj around the `hdiutil` and incorporate into the disk image script.
- Basic sanitizing of non-numeric characters on input (e.g. dashes, spaces)
- Add option to import from CSV,  with sanitizing (this will be preferred option)
- Legacy inline mode still supported.
- Reformat of input parameters for new options.

- Again to make our dev lives easier 😄  

associated with UML-805

## Approach
- Change options on Shell script to have the option to run files
- Enhance main shell script further to extract JSON out of running AWS task using awk 
- Add disk creation to the main script.
- Update readme to reflect the changes.

## Learning
amongst other things...
- Shell variable substitution
- Using `getopts` on shell scripts: https://stackoverflow.com/questions/11279423/bash-getopts-with-multiple-and-mandatory-options
- Shell conditional expressions: https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html
- Shell parameters and while loops: https://unix.stackexchange.com/questions/31414/how-can-i-pass-a-command-line-argument-into-a-shell-script
- Basic Awk: https://unix.stackexchange.com/questions/399560/using-awk-to-select-rows-with-specific-value-in-specific-column
- `tr -d` (remove spaces): https://www.geeksforgeeks.org/tr-command-in-unix-linux-with-examples/
- `python -u` (unbuffered python output): https://stackoverflow.com/questions/881696/unbuffered-stdout-in-python-as-in-python-u-from-within-the-program
- `tee` (pipe to stdout and file at the same time): http://man7.org/linux/man-pages/man1/tee.1.html
- `set -euxo pipefail` (thanks @cooperaj) - https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
- removing illegal chars using `export LC_ALL=C`: https://unix.stackexchange.com/questions/141420/tr-complains-of-illegal-byte-sequence

## Checklist

* [X] I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant


scripts/generate_actor_codes/make_encrypted_image.sh
```diff
@@ -2,12 +2,12 @@
 
 function make_encrypted_image() {
 
-  hdiutil create -size 1024k -fs HFS+ -encryption AES-256 -volname $FILENAME /tmp/$FILENAME/
+  hdiutil create -srcfolder /tmp/$FILENAME/ -fs HFS+ -encryption AES-256 -volname $FILENAME  /tmp/$FILENAME/
   hdiutil mount /tmp/$FILENAME.dmg
```
cooperaj, line 6: When you make an image from a folder you don't need to do anything beyond that. The contents of the folder will now be in the image. 

The rest of these lines assumed an empty image that you needed to mount and copy things into.

williamfalconeruk: I guessed these were there, so we could verify that the image worked - I can remove those lines though...

williamfalconeruk: sorted 😁

scripts/generate_actor_codes/README.md
```diff
@@ -23,42 +24,50 @@ or
...
-- All actor codes should be split with a space into 3 groups of 4. e.g `ABC1 D2E3 FG4H`
+note `${FILENAME}` is activation_codes_(current date).
 
-Then create an encrypted disk image using the script below. The script will prompt for a password to create and to open the disk image.
+Then create an encrypted disk image using the script below.
+The script will prompt for a password to create and to open the disk image.
 
 ``` shell
-./make_encrypted_image.sh $FILENAME
+FILENAME=activation_codes_$(date +%Y%m%d)
```
williamfalconeruk, line 68: editing this now.

scripts/generate_actor_codes/generate_actor_codes.sh
```diff
@@ -0,0 +1,57 @@
...
+
+        echo "/tmp/${FILENAME}/${FILENAME}.txt generated."
+        echo "Contents for checking:"
+        cat /tmp/${FILENAME}/${FILENAME}.txt
+        echo "removing intermediate file..."
+        rm /tmp/${FILENAME}/${FILENAME}.log
+
+        echo "creating encrypted disk image...."
+
+        ./make_encrypted_image.sh ${FILENAME}
```
andrewpearce-digital, line 53: Can the contents of this script be run here instead?

williamfalconeruk: i was going to - but thought i'd try not to bring it in for now as I wanted to isolate that for testing. Also, that script being separate may come in handy for other things too.

scripts/generate_actor_codes/generate_actor_codes.sh
```diff
@@ -0,0 +1,84 @@
...
+    echo "removing intermediate file..."
+    rm /tmp/${FILENAME}/${FILENAME}.log
+}
+
+function make_encrypted_image() {
+    hdiutil create -srcfolder /tmp/${FILENAME}/ -fs HFS+ -encryption AES-256 -volname ${FILENAME}  ~/Documents/${FILENAME}.dmg
+    rm -r /tmp/${FILENAME}
+}
+
+set -euo pipefail
```
cooperaj, line 54: this should live at the top of the file. I think.

williamfalconeruk: yep tried that, was for some reason interfering with some of the things I'm doing - there's another tranche of changes to follow after actually encountering a couple of data quality issues and business based scenarios yesterday, so I'll give it another go.

williamfalconeruk: yep, it's ok locally now, since I've made some changes to the command option... that change to follow.

scripts/generate_actor_codes/generate_actor_codes.sh
williamfalconeruk: Done.

PR #358, opened 2020-05-18
Uml 782 copy access code

nickjones-moj:
## Purpose
Adding in copy access code functionality to allow users to copy code to clipboard.

Fixes UML-782

## Approach

## Checklist


service-front/web/src/scss/_lpa-copyclipboard.scss
```diff
@@ -0,0 +1,76 @@
...
+	box-shadow: none
+}
+
+.app-copy-button:active {
+	margin-top: 2px;
+	-webkit-box-shadow: none;
+	box-shadow: none
+}
+
+.offscreen {
```
johnnolan, line 68: This can be within the app-copy-button class, else if its to reuse then maybe in a separate helpers file?

service-front/web/src/scss/_lpa-copyclipboard.scss
```diff
@@ -0,0 +1,76 @@
...
+}
+
+@media print {
+	.app-copy-button {
+		font-size: 14pt;
+		line-height: 1.2
+	}
+}
+
+.app-copy-button:focus:not(:hover) {
```
cooperaj, line 46: The nice thing about scss and sass is it lets you do this:

```
.app-copy-button {
	font-family: "GDS Transport", Arial, sans-serif;
	-webkit-font-smoothing: antialiased;
       ...

        &:focus,
        &:active {
                ...
        }
}
```

cooperaj: which imo makes it all much more readable

service-front/web/src/javascript/copyAccessCode.js
```diff
@@ -0,0 +1,14 @@
+const copyAccessCode = (element) => {
```
cooperaj, line 1: Just so I'm aware, what browsers does this target?

service-front/web/src/scss/_lpa-copyclipboard.scss
nickjones-moj: The offscreen class is for a hidden text field that the code is copied from, so not directly related to the button.

service-front/web/src/scss/_lpa-copyclipboard.scss
nickjones-moj: Good point! I've now replaced the css with the code used in the gov design system scss for the button.

service-front/web/src/scss/_lpa-copyclipboard.scss
williamfalconeruk: would a visibility: hidden be better here? absolute positioning might mess up the screen...

nickjones-moj: The javascript doesn't work with a hidden field unfortunately. Absolute positioning seems to ok. Another approach could be to make the field visible and heavily style it to not look like a field - so the field would be the large code in the green box?

williamfalconeruk: Ah cool, all good - just found this SO post which mentions the approach:
 https://stackoverflow.com/questions/31593297/using-execcommand-javascript-to-copy-hidden-text-to-clipboard 
The comments mention IE compatibility, so might also want to work that in too, unless we're not supporting it of course...

nickjones-moj: This is where the code was originally from: https://www.w3schools.com/howto/howto_js_copy_clipboard.asp it's not supported in ie8 and below. According to this https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices we support ie11 so should all be good

service-front/web/src/javascript/copyAccessCode.js
nickjones-moj: The Javascript supports all browsers apart from IE8 and below. I believe we need to only support IE11 and above though.

PR #359, opened 2020-05-19
Uml 727 & 728 - Stats page

Raffers:
## Purpose
Creates a base stats page for view & use lpa. Based on the design by NJ

Fixes UML-727 & UML-728

## Approach

## Learning

## Checklist


service-front/app/config/routes.php
```diff
@@ -127,6 +128,11 @@
         Actor\Handler\LpaRemovedHandler::class
     ], 'lpa.removed');
 
+    // Admin pages
```
cooperaj, line 131: Is the intention that this will be hidden behind an admin area? I though it was going to be public. 

That being the case this would probably be better at (or near) the top of the actor routes, near healthcheck etc.

service-front/app/features/actor-add-an-lpa.feature
```diff
@@ -44,10 +45,9 @@ Feature: Add an LPA
 
     Examples:
       | passcode | reason |
-      | T3ST PA22C0D3 | Your activation key must only include letters, numbers and dashes |
-      | T3ST PA22-C0D3 | Your activation key must only include letters, numbers and dashes |
+      | T3ST$PA22-C0D3 | Your activation key must only include letters, numbers and dashes |
       | T3STP*22C0!? | Your activation key must only include letters, numbers and dashes |
-      | T3ST - PA22 - C0D3 | Your activation key must be 12 numbers and letters long |
+      | C - T3S* - PA22 - C0D3 | Your activation key must only include letters, numbers and dashes |
```
cooperaj, line 50: Rather than changing this test case it would be better to add a new one. We know this old case works and it should continue to work after the changes you've made. Additional cases need making for newly introduced features (in most cases)

service-front/app/features/view-lpa-using-sharecode.feature
```diff
@@ -45,9 +45,10 @@ Feature: View an LPA via sharecode
 
     Examples:
       | accessCode | reason |
-      | T3ST ACE2-C0D3 | LPA access codes are 13 numbers and letters long and start with a V |
+      | V-T3ST_ACE2-C0D3 | LPA access codes are 13 numbers and letters long and start with a V |
```
cooperaj, line 48: Same here

service-front/app/src/Actor/src/Handler/LpaAddHandler.php
```diff
@@ -71,6 +71,9 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                 //  Attempt to retrieve an LPA using the form data
                 $postData = $form->getData();
 
+                // Remove C from start of the code if present
+                $postData['passcode'] = preg_replace('/^[cC]?/', '', $postData['passcode']);
```
cooperaj, line 75: Regular expressions can be marked as case-insensitive. I believe by adding an 'i' after the termination character (/ in this case). Something like `/^c?/i` would do it.

service-front/app/src/Actor/src/Form/LpaAdd.php
```diff
@@ -69,14 +69,14 @@ public function getInputFilterSpecification(): array
...
-                            'max'      => 14,
+                            'max'      => 31,
                             'message'  => 'Your activation key must be 12 numbers and letters long',
                         ],
                     ],
                     [
                         'name'    => Regex::class,
                         'options' => [
-                            'pattern' => '/^([[:alnum:]]{4}(?\'dash\' |-|)[[:alnum:]]{4}(\g{dash})[[:alnum:]]{4})$/',
+                            'pattern' => "/(^[cC]?|^)(?'dash' |-|){0,6}[[:alnum:]]{4}(\g'dash'){0,6}[[:alnum:]]{4}(\g'dash'){0,6}[[:alnum:]]{4}$/",
```
cooperaj, line 79: **/(^[cC]?|^)(?'dash' |-|){0,6}[[:alnum:]]{4}(\g'dash'){0,6}[[:alnum:]]{4}(\g'dash'){0,6}[[:alnum:]]{4}$/**

`(?'dash' |-|)` this does not allow underscores which I have seen in a few of the examples?
`{0,6}` why do we allow up to 6 spaces or hyphens? (`c     -1234     -3353     -4322` is a valid code)

cooperaj: Note. 3 spaces are truncated from each of those separators by githubs markdown.

service-front/app/test/ViewerTest/Handler/EnterCodeHandlerTest.php
```diff
@@ -20,7 +20,7 @@
 class EnterCodeHandlerTest extends TestCase
```
cooperaj, line 20: Where we are forced to touch handler unit tests the general approach should be to verify that all functionality is covered by feature/behat testing and then to delete the test file completely. 

In the class that would be tested by the now deleted file you need to mark it as `@codecoverageignore`. This is because unit testing framework/glue code is overly complex and given good testing of the business code also unnecessary.

service-front/app/src/Actor/src/Form/LpaAdd.php
Raffers: The underscores, were only used in tests that were failing. The additional count for spaces & dashes was at the request of Katie for a bigger margin of error on input.

service-front/app/test/ViewerTest/Handler/EnterCodeHandlerTest.php
Raffers: Made the changes & removed the test

service-front/app/features/actor-add-an-lpa.feature
Raffers: With this being an extension of a feature I wouldn't say a new test is required, just the adjusting of existing tests to make sure the scenarios are covered.

service-front/app/src/Actor/src/Handler/LpaAddHandler.php
Raffers: changes made.

PR #375, opened 2020-06-01
Created sorting algorithm

GemTay:
## Purpose
Sort the lpa's by donors surname for the dashboard page so that it is organised and easier to search

Fixes UML-675

## Approach

I run a for on the list of lpa's and extract the donors surname to a new array. This array of surnames is then sorted alphabetically. Finally, I run a for loop on each of the surnames in the now ordered surname array. Within this for loop, we loop through each lpa in the list until we find an lpa that matches the current surname from the surname array. When found, we add that lpa to a new array, remove it from the lpa array and continue to the next ordered surname to find its match. 

## Checklist


service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -256,6 +256,30 @@ public function confirmLpaAddition(string $userToken, string $passcode, string $
...
+        $names = [];
+        foreach ($lpas as $lpaKey => $lpaData) {
+            $surname = $lpaData['lpa']->getDonor()->getSurname();
+            array_push($names, $surname);
+        }
+
+        sort($names);
+        $orderedLpas = [];
+
+        foreach ($names as $key => $name) {
```
williamfalconeruk, line 270: i wonder if there is an out of the box  PHP sorting algorithm?

williamfalconeruk: https://www.php.net/manual/en/function.asort.php
https://www.php.net/manual/en/function.ksort.php - might be of use - not sure....

williamfalconeruk: or even... https://www.php.net/manual/en/array.sorting.php

GemTay: @williamfalconeruk From what I've read, the php sort functions that are out there work for simple structured key => value arrays, however theres no simple sort function for sorting by a specific key value nested within an array. At least that's what I think...

mattmachell: Worth looking at the usort/uksort functions for sorting arrays with a custom sort function. https://www.php.net/manual/en/function.usort.php

cooperaj: Or, alternatively use a functional approach and looking into `array_walk` and its sibling functions.

I agree that this function should be refactored.

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -256,28 +256,23 @@ public function confirmLpaAddition(string $userToken, string $passcode, string $
...
-        foreach ($names as $key => $name) {
-            foreach ($lpas as $lpaKey => $lpaData) {
-                $surname = $lpaData['lpa']->getDonor()->getSurname();
-                if ($name === $surname) {
-                    $orderedLpas[$lpaKey] = $lpaData;
-                    unset($lpas[$lpaKey]);
-                }
-            }
-        }
+        uasort($lpas, function ($a, $b) {
```
williamfalconeruk, line 269: Much cleaner :)

PR #379, opened 2020-06-04
UML-829

andrewpearce-digital:
## Purpose

Create and Seed data into the lpa-codes API service.

Fixes UML-829

## Approach

Create a script to produce data for lpa-codes from data in our legacy code generator service.
Create a script to push generated data into the lpa-codes DynamoDB table.
Update the api-seeding container to run `put_lpa-codes` script when running `docker-compose`.

## Learning

It was necessary to enable `-sharedDb` mode on the lpa-codes service DynamoDB-local container. Without this, each connected client could not see the lpa-codes-local table.

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.UsageNotes.html

## Checklist


andrewpearce-digital:
docker-compose successfully seeds lpa-codes
image

andrewpearce-digital:
Imported cases are successfully returned with calls to the api
image
image

scripts/remove_default_sg_rules/remove_default_sg_rules.py
```diff
@@ -67,7 +67,7 @@ def get_ingress_permissions(self):
                         'Values': ['default']
                     },
                 ],
-                MaxResults=1
+                MaxResults=5
```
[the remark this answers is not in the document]
andrewpearce-digital: the sg rules script returned an error stating the max returned results is a range from 5 to inf. It must have been an edit I had made when cleaning things up before merging that script in a previous PR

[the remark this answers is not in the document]
andrewpearce-digital: I suppose I could remove MaxResults alltogether but I wanted to be cautious in the event of a bad match

service-api/seeding/seeding_lpa_codes.json
```diff
@@ -0,0 +1,57 @@
+[
```
cooperaj, line 1: Is this our new development seeded data? 

What about the seeding that happens in the current seeding script?

andrewpearce-digital: The current seeding puts things into our dynamodb tables and still puts codes into the old codegen table.

I've added an extra command to the seeding container so it runs our seeding then runs lpa-codes seeding using this json data.

When we no longer need to put actor codes in our old table, we can remove that section of the code from dynamodb.py.

cooperaj: 👍

PR #391, opened 2020-06-10
Contract testing on the pipeline

andrewpearce-digital:
## Purpose
Run contract tests in the pipeline

Fixes UML-822

## Approach

Configure pipeline environment variables
Add a step in the docker build and test job to run contract testing.

## Learning

Also attempting this fix for behat smoke test failures
https://stackoverflow.com/questions/54748481/how-to-fix-behat-dmore-chromedriver-streamreadexception-in-behat-script

## Checklist


.circleci/config.yml
```diff
@@ -583,6 +583,18 @@ orbs:
                 docker exec inttests /app/vendor/bin/behat -p integration -f progress -o std -f junit -o /app/build/reports/int
                 docker exec inttests /usr/local/bin/php -dapc.enable_cli=1 \
                   /app/vendor/bin/behat -p acceptance -f progress -o std -f junit -o /app/build/reports/acc
+          - run:
+              name: Run pact tests
+              command: |
+                ~/project/scripts/pipeline/set_environment_variables/set_pact.sh >> $BASH_ENV
+                docker-compose -f ~/project/docker-compose.testing.yml up -d lpa-codes-pact-mock
+                sleep 3
```
williamfalconeruk, line 591: are we waiting for something to settle here or is there another reason for the `sleep 3`?

scripts/pipeline/set_environment_variables/set_pact.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [ $(aws --version 2>&1 | grep -c "aws-cli") -ne 0 ]
+then
+  export SECRET_STRING=$(aws sts assume-role \
+  --role-arn "arn:aws:iam::${PACT_BROKER_ACCOUNT}:role/get-pact-secret-development" \
+  --role-session-name AWSCLI-Session | \
+  jq -r '.Credentials.SessionToken + " " + .Credentials.SecretAccessKey + " " + .Credentials.AccessKeyId' 2>/dev/null)
+
+  #local export so they only exist in this stage
+  export AWS_ACCESS_KEY_ID=$(echo "${SECRET_STRING}" | awk -F' ' '{print $3}' 2>/dev/null)
```
williamfalconeruk, line 17: just a query, whats the reason for pumping stderr to /dev/null?

williamfalconeruk: just read this - https://stackoverflow.com/questions/40714202/what-is-the-meaning-of-2-in-2-dev-null# think explains it. i guess we are not bothered with the errors coming out.

andrewpearce-digital: This is taken from the other consumer products. I didn't see a reason not to suppress the error messages. I might rewrite in the future to make it fit more stylistically with the rest of our bash, but for now we know it gets the secrets and sets some values. More will come when we actually implement the tests.

.circleci/config.yml
williamfalconeruk: might be for Pact to settle

PR #395, opened 2020-06-11
set to 80% with 1% leniency above that level

williamfalconeruk:
## Purpose
set exact target with initial 5% drop leniency. we can adjust this later once more confident.

Relates to UML-431

## Approach

adjust codecov.yml

## Learning

## Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,3 +4,8 @@ codecov:
     layout: "reach, diff, flags, files"
     behavior: default
     require_changes: false  # if true: only post the comment if coverage changes
+  status:
+    project:
+      default:
+        target: 80%    # the required coverage value
+        threshold: 5%  # the leniency in hitting the target setting to 5% now.
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,5 +7,5 @@ codecov:
   status:
     project:
       default:
-        target: 80%    # the required coverage value
-        threshold: 5%  # the leniency in hitting the target setting to 5% now.
+        target: 80%    # our target PR coverage to be successful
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -4,3 +4,8 @@ codecov:
     layout: "reach, diff, flags, files"
     behavior: default
     require_changes: false  # if true: only post the comment if coverage changes
```
cooperaj, line 6: This is probably clutter we can turn off right?

codecov.yml
```diff
@@ -4,3 +4,8 @@ codecov:
...
+  status:
+    project:
+      default:
+        target: 80%    # the required coverage value
```
cooperaj, line 10: Is there no way just to specify 80 as the minimum? This would suggest a single PR could go in that would drop our coverage to 75% (which would be a maaaaaaasive PR, but still)

williamfalconeruk: upon further reading, I think the hard limit is 80%, then massive drop of 5 percent is when we are above that limit. could warrant a tweak to say 1%. may ask you to re-review @cooperaj and @andrewpearce-digital  in a minute :)

codecov.yml
williamfalconeruk: yup - though the way the PR reporting works is that it edits the existing comment, so the clutter is hidden in the edits.

andrewpearce-digital: useful to keep it on for a little bit while we see how this works in practice, otherwise we'll have to force a coverage change to see what's been implemented.

PR #396, opened 2020-06-15
Fix docker typos

andrewpearce-digital:
## Purpose
LPA Codes seeding cannot run because of a duplication of the port in the waitforit command.

## Approach

Remove extra port in command

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -183,7 +183,7 @@ services:
 
       # Local only
       AWS_ACCESS_KEY_ID: "-"
-      AWS_ENDPOINT_DYNAMODB: http://localstack:4569
+      AWS_ENDPOINT_DYNAMODB: http://localstack:14569
       AWS_SECRET_ACCESS_KEY: "-"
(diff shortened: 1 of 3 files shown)


cooperaj:
The port number is also fixed in the codes integration branch. It doesn't stop the seeding as far as I'm aware as the failure of the waitforit command doesn't exit the script.

docker-compose.yml
```diff
@@ -183,8 +183,9 @@ services:
 
       # Local only
       AWS_ACCESS_KEY_ID: "-"
-      AWS_ENDPOINT_DYNAMODB: http://localstack:4569
+      AWS_ENDPOINT_DYNAMODB: http://localstack:14569
       AWS_SECRET_ACCESS_KEY: "-"
+      LPA_CODES_STATIC_AUTH_TOKEN: asdf1234567890
```
cooperaj, line 188: This is only needed in my branch (or Gemma's forthcoming one)

cooperaj: It's not a requirement for bringing up the env.

PR #400, opened 2020-06-16
Uml 820 healthcheck lpa codes

GemTay:
## Purpose
Checks that the service is able to talk to the lpa codes api

Fixes UML-820

## Approach

Changed db dependancy check to use the actor users table, and added another dependency to the healthcheck to ensure that the service can talk to the lpa codes api healthcheck with an OK response

## Checklist


GemTay:
image

service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -50,18 +75,18 @@ public function __construct(
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
         return new JsonResponse([
-            "version" => $this->version,
-            "dependencies" => [
-                "api" => $this->checkApiEndpoint(),
-                "dynamo" => $this->checkDynamoEndpoint()
-            ],
-            "healthy" => $this->isHealthy()
+            "api" => $this->checkApiEndpoint(),
```
cooperaj, line 78: how come the rejig of the data structure? We've lost version and healthy has been renamed? I would have prefered to see the lpa codes api line getting slotted in the dependencies array ie. a single line change

GemTay: @cooperaj It seemed a little cluttered and repetitive before, with a second "dependencies" => inside "api" => inside "dependencies" => so I though I could remove some of that. I removed the version from the api side as we have it in the front (more repetition), and the api and front version are always going to be the same arent they?

cooperaj: > It seemed a little cluttered and repetitive before,

Yes, I guess so. But it did allow use to at-a-glance track down what part of the system had failed. Realistically this structure was intended to allow AWS to do automated things based on this information and machines don't care so much about repetition. Could be worth getting Andrews ideas on this.

> and the api and front version are always going to be the same arent they?

For the most part yes but it's not guaranteed. If we have to rollback a release of the API or viewer or actor side, because fixing forward will take too long, we will have different versions.

GemTay: Added the api version back in and renamed "all_dependencies_healthy"

GemTay:
image

PR #409, opened 2020-06-24
UML- 878 - Account Creation Metric Filter

andrewpearce-digital:
## Purpose
Count account creation events

Fixes UML-878

## Approach

Reorganise log groups so that every environment has it's own
Update application logging so that event_codes are produced on certain events
Create a metric filter resource that increments when matching event_codes appear in logs

## Learning

https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html
https://www.terraform.io/docs/providers/aws/r/cloudwatch_log_metric_filter.html

Metric filters increment a count when a matching pattern is found in the logs. I cannot count events that happened before the metric filter was put in place.
Metric filters are applied to log groups. As we consolidate application logs into an account level log group, metric filters are being applied to logs for all dev environments. We can change this pattern and create log groups at the environment level.

## Checklist


terraform/environment/cloudwatch_metrics.tf
```diff
@@ -0,0 +1,12 @@
+resource "aws_cloudwatch_log_metric_filter" "account_creation" {
+  name           = "account_creation_messages"
+  pattern        = "{ $.message = \"Account with Id*created using email*\" }"
```
cooperaj, line 3: Would it be useful to have a particular code associated with this? i.e. ACCOUNT_CREATED

cooperaj: I think we already use code for certain errors. we can move to event if that makes more sense.

cooperaj: Mind if I work on this PR to add those?

andrewpearce-digital:
image

cooperaj:
https://stackoverflow.com/questions/41051769/how-can-i-access-aws-cloudwatch-custom-metrics-via-php-sdk

terraform/environment/cloudwatch_metrics.tf
andrewpearce-digital: Didn't see this message due to needing a refresh, You're welcome to pull and have a go.

terraform/environment/terraform.tfvars.json
```diff
@@ -17,7 +17,8 @@
       "cookie_expires_use" : 1440,
       "logging_level": 100,
       "pagerduty_service_name": "Use a Lasting Power of Attorney Non-Production",
-      "use_legacy_codes_service": false
+      "use_legacy_codes_service": true,
```
williamfalconeruk, line 20: was this meant to be turned back to legacy mode in dev?

andrewpearce-digital:
```suggestion
      "use_legacy_codes_service": false,
```

PR #416, opened 2020-07-02
Uml 882 viewer session expired page

GemTay:
## Purpose
Takes the viewer to a new session expired page after 30 minutes of inactivity

Fixes UML-882

## Approach

Created a new session expiry handler in the viewer side, with the same path as the actor side (/session-expired). Therefore the uri is created for the users environment (viewer/actor), passing the request onto the related session expiry handler and displaying the page

## Checklist


service-front/app/src/Actor/templates/actor/actor-session-expired.html.twig
```diff
@@ -13,7 +13,7 @@
 
                     <p class="govuk-body">For your security, we've signed you out of this service because you did not use it for 20 minutes.</p>
 
-                    <a href="{{ path('login') }}" draggable="false" class="govuk-button">Log back in</a>
+                    <a href="{{ path('login') }}" draggable="false" class="govuk-button">Sign in</a>
```
GemTay, line 16: Changed this button text after from Katie's request - rather than create a whole new PR w/ ticket

PR #418, opened 2020-07-03
Uml-824 - Pact test implementation 

Raffers:
## Purpose
This ticket is to use Pact to test our contracts with the integration of the new Code generation service.

As we will be consumers of the Code service, we can then work with the integration team to get any changes done as needed. 

Fixes UML-824

## Approach

Creates a containers to run the Pact broker & pushes all tests through here. Before uploading the Pact file

## Learning

## Checklist


.idea/runConfigurations/API_Pact.xml
```diff
@@ -0,0 +1,11 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="API Pact" type="PhpBehatConfigurationType" singleton="false">
+    <BehatRunner configuration_file="$PROJECT_DIR$/service-api/app/behat.yml" scenario="" scope="Yml" options="--profile=pact --tags=pact&amp;&amp;pact" use_alternative_configuration_file="true" />
```
Raffers, line 3: Just noticed the double pact, will adjust shortly

.idea/php.xml
```diff
@@ -5,6 +5,41 @@
       <PhpCSConfiguration beautifier_path="$PROJECT_DIR$/service-front/app/vendor/squizlabs/php_codesniffer/bin/phpcbf" standards="MySource;PEAR;PSR1;PSR12;PSR2;Squiz;Zend" tool_path="$PROJECT_DIR$/service-front/app/vendor/squizlabs/php_codesniffer/bin/phpcs" />
     </phpcs_settings>
   </component>
+  <component name="PhpIncludePathManager">
```
cooperaj, line 8: Would remove this component block of changes as it's not relevant I don't think and only applies to the new container.

service-api/app/composer.json
```diff
@@ -29,6 +29,10 @@
         {
             "type": "vcs",
             "url": "git@github.com:ministryofjustice/behat-psr-extension.git"
+        },
+        {
+            "type": "vcs",
+            "url": "git@github.com:Raffers/pact-behat-extension.git"
```
cooperaj, line 35: Do we need this fork now?

service-api/docker/app/Dockerfile
```diff
@@ -18,7 +18,4 @@ COPY --from=composer /app/vendor /app/vendor
 COPY service-api/docker/app/app-php.ini /usr/local/etc/php/conf.d/
 COPY service-api/docker/app/fpm-pool.ini /usr/local/etc/php-fpm.d/zz-logging.conf
 
-WORKDIR /app
-
-CMD ([[ -z "${ENABLE_XDEBUG}" ]] || docker-php-ext-enable xdebug) \
-    && php-fpm
+WORKDIR /app
```
cooperaj, line 21: Is this and the entrypoint necessary since we don't run in container any more?

.idea/php.xml
cooperaj: It's odd that we have to micromanage these files like this. I dislike the way PHPStorm does this.

service-api/app/composer.json
Raffers: My change hasn't been pushed upstream, so the fork will be needed.

service-api/docker/app/Dockerfile
Raffers: Not sure what you mean? We're still running in api-app with the interactions done via lpa-pact-most

.idea/php.xml
williamfalconeruk: its a real pain.

service-api/app/composer.json
williamfalconeruk: wonder if we need to push this into a forked MoJ repo? just a thought...

service-api/docker/app/Dockerfile
cooperaj: This alteration (and the entrypoint file) was made when we would be running the pact mock in the api-app alpine container. Now we're running it as the original separate container, IIRC and based on the docker-compose.testing.yml

Raffers: Ah, yer, good point. Sorry still in a Monday daze.

docker-compose.testing.yml
```diff
@@ -36,3 +36,9 @@ services:
         - use_a_lasting_power_of_attorney
         - --provider
         - lpa-codes
+
+  api-app:
+    container_name: api-app
```
cooperaj, line 41: Should only need to specify things you want to change IIRC

.circleci/config.yml
```diff
@@ -587,11 +587,20 @@ orbs:
               name: Run pact tests
               command: |
                 ~/project/scripts/pipeline/set_environment_variables/set_pact.sh >> $BASH_ENV
-                docker-compose -f ~/project/docker-compose.testing.yml up -d lpa-codes-pact-mock
+
+                echo 'export DOCKER_REMOTE_PACT_IP="$(docker run -d --name lpa-codes-pact-mock pactfoundation/pact-cli:latest \
```
cooperaj, line 591: I would suggest looking at running the container first and then using docker inspect to pull the container IP on the default network (which is where it'll be assigned). It'll probably look something like 172.18.0.4. Unless otherwise specified all containers are spun up in that network. You should be able to run the docker inspect commands output though something like grep and awk to extract the value.

.circleci/config.yml
```diff
@@ -587,11 +587,20 @@ orbs:
...
+                                                -p "80" --pact-dir /tmp/pacts \
+                                                --consumer use_a_lasting_power_of_attorney --provider lpa-codes \
+                                                wget -qO - https://checkip.amazonaws.com)"' >> $BASH_ENV
+
+                docker run -d --name pacttests --add-host lpa-codes-pact-mock:$DOCKER_REMOTE_PACT_IP api-app:latest
```
cooperaj, line 596: Bang on.

.circleci/config.yml
```diff
@@ -587,11 +587,20 @@ orbs:
...
+
                 sleep 3
                 # run some tests
-                # docker-compose -f ~/project/docker-compose.testing.yml exec lpa-codes-pact-mock cat /tmp/pacts/use_a_lasting_power_of_attorney-lpa-codes.json > pact.json
-                docker-compose -f ~/project/docker-compose.testing.yml stop lpa-codes-pact-mock
+                docker exec pacttests /app/vendor/bin/behat -p pact -f progress -o std -f junit -o /app/build/reports/int
+                docker exec pacttests cat /tmp/pacts/use_a_lasting_power_of_attorney-lpa-codes.json > pact.json
+
+                docker stop lpa-codes-pact-mock
```
cooperaj, line 603: I don't think we manually stop any containers (though I could be wrong) as they all get killed off when we're done anyway.

.circleci/config.yml
Raffers: I hadn't seen this comment but lastest commit this is what I've done. Latest error though is it's timing out.

.circleci/config.yml
Raffers: Combining like that just gave an output of the full container ID

service-api/app/composer.json
cooperaj: Thats the plan I believe

.circleci/config.yml
cooperaj: Oh no, I meant that this line looks how I envisaged it.

.idea/runConfigurations/PactMock.xml
```diff
@@ -0,0 +1,15 @@
+<component name="ProjectRunConfigurationManager">
```
cooperaj, line 1: Is this configuration needed now that the mock is started as a dependency?

Raffers: Nope, just removed it in my latest commit.

.idea/php.xml
cooperaj: Still need to remove this

docker-compose.yml
```diff
@@ -187,6 +187,7 @@ services:
       AWS_ENDPOINT_DYNAMODB: http://localstack:14569
       AWS_SECRET_ACCESS_KEY: "-"
       LPA_CODES_STATIC_AUTH_TOKEN: asdf1234567890
+      INSTALL_TEST_DEPENDENCIES: "true"
```
cooperaj, line 190: Won't need this anymore as we've removed the install on start entrypoint stuff

service-api/app/composer.json
cooperaj: Need to make a ticket to get this brought in as an MoJ repo

docker-compose.yml
```diff
@@ -187,7 +187,8 @@ services:
       AWS_ENDPOINT_DYNAMODB: http://localstack:14569
       AWS_SECRET_ACCESS_KEY: "-"
       LPA_CODES_STATIC_AUTH_TOKEN: asdf1234567890
-      INSTALL_TEST_DEPENDENCIES: "true"
+      INSTALL_TEST_DEPENDENCIES: "false"
```
cooperaj, line 190: I dont think we need it at all (sorry). It was introduced for this PR and is not used at all now

PR #419, opened 2020-07-03
Uml 881 expire dialog

johnnolan:
## Purpose
Creates a dialog box in JS ready for the implementation of the session timeout. 

Fixes UML-881

## Approach

Adds in a timeout and a dialog for when when a session is about to expire.

## Learning

Cannot test key pressses in JSDom which means our testing is a little short on the JS. One to investigate in the future or cover via Behat in the future.

## Checklist

### DIFF
--- a/service-front/web/src/javascript/sessionDialog.js
+++ b/service-front/web/src/javascript/sessionDialog.js
@@ -3,6 +3,7 @@ import countdownTimer from "./countdownTimer";
 export default class SessionDialog {
     constructor(element, countdownMinutes)
     {
+        this.showDialogMinutesBeforeLogout = 5;
         this.element = element;
         this.countdownMinutes = countdownMinutes; // Temporary until final solution in place
         this.dialogOverlay = document.getElementById("dialog-overlay");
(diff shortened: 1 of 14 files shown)


service-front/web/src/javascript/sessionDialog.js
```diff
@@ -0,0 +1,92 @@
...
+    {
+        this.element = element;
+        this.countdownMinutes = countdownMinutes; // Temporary until final solution in place
+        this.dialogOverlay = document.getElementById("dialog-overlay");
+        this.dialogFocus = document.querySelector(".dialog-focus");
+        this._setupEventHandlers();
+        this._trapFocus();
+        this.timer = new countdownTimer(this.element.querySelector('#time'), this.countdownMinutes);
+        this.timer.on('tick', (event) => {
+            if (event === 5) {
```
williamfalconeruk, line 14: could we make it clearer what the `event`value of 5 is in a `const` or something similar?

johnnolan: Good call, have updated this 👍

service-front/web/src/javascript/sessionDialog.js
```diff
@@ -0,0 +1,94 @@
+import countdownTimer from "./countdownTimer";
+
+export default class SessionDialog {
+    constructor(element, countdownMinutes)
+    {
+        this.showDialogMinutesBeforeLogout = 5;
```
williamfalconeruk, line 6: I guess this value will need to be configurable when we plumb this dialog in?

johnnolan: Yeah that sounds like a good idea, also the countdown time as well would be nice to configure this way!

PR #421, opened 2020-07-03
Uml 804 enable encrypted brute force cache Part 1 - create replication group

williamfalconeruk:
## Purpose
Creates the account-wide replication group for elasticache on redis, which will replace the current cluster.

Fixes UML-804 

## Approach
**1 of 3 PR's**, each will need to be merged in turn. covered by subtask UML-919

- create the new replication group.
- enable encryption at rest and encryption in transit

Further PR'S will:
- associate the cache to app using a primary endpoint
- remove the old cluster

## Learning
- Terraform for Elasticache and redis: https://www.terraform.io/docs/providers/aws/r/elasticache_replication_group.html
Note: this is more to prove the concept at this stage from basic terraform, will investigate the module approach further
- background: https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Replication.Redis.Groups.html
- investigating this approach: https://github.com/tmknom/terraform-aws-elasticache-redis

## Checklist

* [x] I have added tests to prove my work (i have run terraform plan)


terraform/account/elasticache.tf
```diff
@@ -22,3 +23,23 @@ resource "aws_elasticache_cluster" "brute_force_cache" {
 
   tags = local.default_tags
 }
+
+
+resource "aws_elasticache_replication_group" "brute_force_cache_rg" {
```
TomTucka, line 28: I'd drop the `_rg` off the end, I don't think you need to be that explicit in the name as you'll always call it with something like this `aws_elasticache_replication_group.brute_force_cache.arn`

williamfalconeruk: yep - it's only the fact that the name clashes with the existing brute force cache, i can change that now

williamfalconeruk: Oh, i'll try again...seems it don't like the duplicate name...
image

PR #422, opened 2020-07-06
Display correct service name on cookie page

GemTay:
## Purpose
Displays the correct service title on the cookie preferences page

Fixes UML-912

## Approach

Renamed the cookie handler routes for actor and viewer sides so that they differ. The cookie handler fetches the route name and passes it to the twig page where an if statement decides which title to display

## Checklist


service-front/app/config/routes.php
```diff
@@ -46,7 +46,7 @@
     $app->get('/stats', Viewer\Handler\StatsPageHandler::class, 'viewer-stats');
     $app->get('/session-expired', Viewer\Handler\ViewerSessionExpiredHandler::class, 'session-expired');
 
-    $app->route('/cookies', Common\Handler\CookiesPageHandler::class, ['GET', 'POST'], 'cookies');
+    $app->route('/cookies', Common\Handler\CookiesPageHandler::class, ['GET', 'POST'], 'viewer-cookies');
```
johnnolan, line 49: In cookieConsent.js there is a piece of code that checks for the /cookies url, line 44.

You will need to update that for maybe a endsWith -cookies or similar so that it still works

johnnolan: On second look, maybe not, but if the visual url has changed then yeah it will need it... need to get up to speed on my php :)

PR #424, opened 2020-07-08
Link cookies page to terms of use and privacy policy

GemTay:
## Purpose
The ticket's original purpose was to add content to the cookie preferences pages, however this content already exists. Instead this ticket now add's links to the terms of use and privacy policy pages to take the user to the cookie preferences pages

UPDATE - this ticket now also adds a back button to the cookies page with the functionality to take the user back to the page they were on before they visited the cookies page

Fixes UML-896

## Checklist

### DIFF
--- a/service-front/app/src/Common/src/Handler/CookiesPageHandler.php
+++ b/service-front/app/src/Common/src/Handler/CookiesPageHandler.php
@@ -46,8 +46,7 @@ class CookiesPageHandler extends AbstractHandler implements UserAware, CsrfGuard
         TemplateRendererInterface $renderer,
         UrlHelper $urlHelper,
         UrlValidityCheckService $urlValidityCheckService
-    )
-    {
+    ) {
         parent::__construct($renderer, $urlHelper);
         $this->urlValidityCheckService = $urlValidityCheckService;
     }
@@ -88,7 +87,9 @@ class CookiesPageHandler extends AbstractHandler implements UserAware, CsrfGuard
         $cookies = $request->getCookieParams();
         $form->setData($request->getParsedBody());
 
-        $response = new RedirectResponse($form->get('referer')->getValue());
+        $response = new RedirectResponse(
+            $this->urlValidityCheckService->setValidReferer($form->get('referer')->getValue())
+        );
 
         if (array_key_exists(self::COOKIE_POLICY_NAME, $cookies)) {
             try {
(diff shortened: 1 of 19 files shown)


williamfalconeruk:
maybe add some behat tests to check that the links are all working

cooperaj:
You've done the test work now but it's worth clarifying that we shouldn't be testing content really. Also the step names are too specific in that we should be describing business actions not instructions. It's not "click the cookies link" but more like "I can navigate to cookie information"

GemTay:
I've refactored the referer validation steps and moved it into the get method (so that it is validated and ready to use on the back button). I added new function inside the `urlValidityCheckService` to return an appropriate url referer depending on the result of both functions `isValid` and `checkRefererRouteValid`. Since all this sanitisation is now done in the get method, no checks are done again in the post method. 

I'd like to get others opinions on whether the referer needs to be checked again in the post method? Since it's stored as a hidden value in the form, could it be manipulated by the user in anyway?

cooperaj:
> Since it's stored as a hidden value in the form, could it be manipulated by the user in anyway?

Never trust user input. A clever person would be able to craft a post submission that could conceivably give a back button that goes anywhere they want.

service-front/app/src/Common/src/Handler/CookiesPageHandler.php
```diff
@@ -96,12 +88,7 @@ public function handlePost(ServerRequestInterface $request): ResponseInterface
         $cookies = $request->getCookieParams();
         $form->setData($request->getParsedBody());
 
-        $isValidRefererRoute = $this->urlValidityCheckService->checkRefererRouteValid($refererRoute = $form->get('referer')->getValue());
-
-        // After setting cookies settings user is taken where they were previously after validating the referer route
-        if ($isValidRefererRoute) {
-            $response = new RedirectResponse($isValidRefererRoute ? $refererRoute : $this->urlHelper->generate('home'));
-        }
+        $response = new RedirectResponse($form->get('referer')->getValue());
```
cooperaj, line 91: Yup, you've removed all the safety here. Don't trust user input.

GemTay: Pushing the referer through the validation service now

service-front/app/test/CommonTest/Service/Url/UrlValidityCheckServiceTest.php
```diff
@@ -93,12 +102,59 @@ public function it_checks_if_referer_route_exists()
...
         );
 
         $valid = $service->checkRefererRouteValid($refererUrl);
 
         $this->assertIsBool($valid);
         $this->assertTrue($valid);
     }
+
+    /** @test */
+    public function it_returns_a_valid_referer()
```
cooperaj, line 116: Should probably also test for a non-valid referrer

PR #431, opened 2020-07-10
Uml 931 perf analytics

johnnolan:
## Purpose
Adds performance and error capturing and send to Google Analytics

Fixes UML-931

## Approach

Uses the OGP performance library [https://www.npmjs.com/package/@ministryofjustice/opg-performance-analytics](https://www.npmjs.com/package/@ministryofjustice/opg-performance-analytics) to collect data and send to the existing GA integration.

## Checklist

### DIFF
--- a/service-front/web/src/javascript/googleAnalytics.js
+++ b/service-front/web/src/javascript/googleAnalytics.js
@@ -1,3 +1,7 @@
+import {
+    PerformanceAnalytics,
+    ErrorAnalytics,
+} from "@ministryofjustice/opg-performance-analytics";
 
 export default class GoogleAnalytics {
     constructor(analyticsId) {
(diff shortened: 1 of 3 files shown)


service-front/web/src/javascript/googleAnalytics.js
```diff
@@ -1,11 +1,17 @@
...
-    constructor(analyticsId) {
+    constructor(analyticsId)
+    {
         this.analyticsId = analyticsId;
         this._setUpOnLoad();
     }
 
-    _setUpOnLoad() {
+    _setUpOnLoad()
+    {
```
williamfalconeruk, line 14: noticed the braces location change - was that intentional? just checking to see if this is consistent with whats there in the code?

johnnolan: Yeah I was just updating it due to php storm feedback. I may look at some point to add in JS linting if that is wrong so we all get the same across our desktops. Looking now as well to double check

johnnolan: Yeah seems to be the pattern as set in other files for functions

PR #438, opened 2020-07-15
UML-932 - Users should start their journey from the gov.uk start page

andrewpearce-digital:
## Purpose
Users should start their journey from the gov.uk start page

Fixes UML-932

## Approach

Add rules to the Application Load Balancer

## Learning

Using referrers may be inconsistent
http://nginx.org/en/docs/http/ngx_http_referer_module.html
So instead we are performing a rewrite on the start page of the service.

http://nginx.org/en/docs/http/ngx_http_rewrite_module.html
https://www.liquidweb.com/kb/redirecting-urls-using-nginx/

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1014,7 +1014,7 @@ jobs:
       - slack/notify:
           title: "Use a Lasting Power of Attorney Development Environment Ready"
           color: "#9933ff"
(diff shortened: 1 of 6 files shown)


andrewpearce-digital:
It's even got tests!
image
image

PR #445, opened 2020-07-17
UML-962 - check state of user in login on actor.

williamfalconeruk:
## Purpose
To Fix UML-962

## Approach
- Actor Triage page now detects if logged in and redirects to Lpa dashboard or lpa add accordingly.
- follows the same approach as login code.
- viewer home page has not changed

## Checklist


service-front/app/features/context/UI/AccountContext.php
```diff
@@ -214,6 +214,37 @@ public function iEnterCorrectCredentials()
         $this->ui->pressButton('Sign in');
     }
 
+    /**
```
cooperaj, line 217: I'll be honest. I'm very puzzled by the existence of this new step definition

cooperaj: Sorry, just jammed this in before it got merged. This ^ needs some discussion/rationale.

service-front/app/src/Actor/src/Handler/ActorTriagePageHandler.php
```diff
@@ -49,6 +55,15 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             return $this->handlePost($request);
         }
 
+        $user = $this->getUser($request);
```
cooperaj, line 58: This replicates functionality we already have, but I can see this PR doesn't remove that? So do we need both?

service-front/app/features/context/UI/AccountContext.php
cooperaj: > I'll be honest. I'm very puzzled by the existence of this new step definition

To clarify, this step is pretty much identical to the one defined above it bar one change to the return. Since I'm going through all the Context files right now I've become painfully aware of (and in some cases, have removed) duplicate steps with different names. This is probably a case where I would have done that.

GemTay: I've removed this step

service-front/app/src/Actor/src/Handler/ActorTriagePageHandler.php
```diff
@@ -49,6 +55,10 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             return $this->handlePost($request);
         }
 
+        if (! is_null($this->getUser($request))) {
```
johnnolan, line 58: Approved but is this a php thingy where there is a space here `if (! is_null` between the ! and is_null? Looks so alien to me

GemTay: @johnnolan It doesn't make any difference, I tend to write it without a space but Billy wrote that and it works just the same :))

cooperaj: I think the PHPCS tool (that we should probably all be using) will grumble about the space.

PR #446, opened 2020-07-17
Uml 879 add a lpa pact

cooperaj:
## Purpose
Utilises Chris' work to implement Pact mock testing of the Add an LPA to an account feature.

Fixes UML-879

## Approach
I moved all the LPA context steps to a new context file and added a trait that implements pact testing features. That trait is used in the LPA context file by the 2 functions needed to test this part of the ticket.

It does need an alteration to the underlying pact behat extension that has been submitted as an upstream change (https://github.com/Raffers/pact-behat-extension/pull/1).

We will need to pull the above repo into the MoJ account and setup the code to use that fork once that is in place.

## Learning
Upstream code does some weird things sometimes.

## Checklist


service-api/app/features/bootstrap/config.php
```diff
@@ -2,15 +2,72 @@
...
+            'requestTracingProcessor' => [
+                'type' => \App\Service\Log\RequestTracingLogProcessorFactory::class,
+                'options' => [], // No options
+            ],
+        ],
+    ],
+
+    'repositories' => [
+        'dynamodb' => [
+            'actor-codes-table' => 'actor-codes',
```
GemTay, line 53: Can the actor codes table be removed since it's no longer used?

cooperaj: Yes. Yes it can

cooperaj: I think we might have to leave it and do a further cleanup ticket where we remove the legacy code stuff. Taking it out just causes all sorts of weird breakage.

PR #460, opened 2020-07-27
Adds Custom Email component for better control of validation

GemTay:
## Purpose
Prevents the laminas email validator regex error from being displayed to the user if they do not enter a valid email address

Fixes UML-933

## Approach

Creates a custom email element to attach the custom email validator and display the correct error messages to the user

## Checklist

### DIFF
--- a/service-front/app/src/Common/src/Form/Element/Email.php
+++ b/service-front/app/src/Common/src/Form/Element/Email.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Common\Form\Element;
+
+use Laminas\Filter\StringTrim;
+use Laminas\Form\Element\Email as LaminasEmail;
+
+/**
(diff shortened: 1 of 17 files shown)


service-front/app/src/Common/src/Form/Element/Email.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function getInputSpecification()
+    {
+        return [
+            'name' => $this->getName(),
+            'required' => true,
+            'filters' => [
+                ['name' => StringTrim::class],
+            ],
+            'validators' => [
```
cooperaj, line 35: By not having validators here the functionality is still covered by the forms validation?

GemTay: Yes, the validators can either be defined here in the email element or inside each form. It's made me wonder which is best to go with...

PR #476, opened 2020-08-04
Uml 993 configure translator factory

cooperaj:
## Purpose
Create a factory to load the translations into our i18n laminas translator

Fixes UML-993

## Approach

Simple factory and tests. Needed to create a basic Welsh translation POT, PO and MO file for it to load and use.

Built on top of routing PR. Only the last commit is worth looking at.

## Learning

The multilingual documentation on the laminas site leaves a lot to be desired.

## Checklist


service-front/app/languages/cy.po
```diff
@@ -0,0 +1,46 @@
...
+"X-Generator: Poedit 2.4\n"
+"Last-Translator: \n"
+"Plural-Forms: nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n==3 ? "
+"3 : n==6 ? 4 : 5);\n"
+"Language: cy\n"
+
+msgid "Enter code"
+msgstr "Rhowch god"
+
+msgid "Back"
```
mattmachell, line 19: Just a word of warning. Welsh translations can vary based on context (particularly with things like numbers from work I've done before), so one English phrase can be potential translated into two or more different Welsh phrases depending on context. So if your translation keys are the English phrase, rather than a key for the page context, you set yourself up for issues where that one phrase has a one to many mapping and the two Welsh phrases for the English term should be different. 

There's also potential issues in situations where you reuse a translation in multiple places, but later change it for one and that doesn't work for across all existing instances. Would strongly recommend a template.context.purpose key method, i.e. login_page.main_form.email_label for keys.

cooperaj: The gettext translation formats support the addition of a [msgctxt](http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html#sec-podisamb) tag for that specific purpose:

```
#: tools/observinglist.cpp:700
msgctxt "First letter in 'Scope'"
msgid "S"
msgstr ""
⁠
#: skycomponents/horizoncomponent.cpp:429
msgctxt "South"
msgid "S"
msgstr ""
```

There is an initial ticket to generate the template files that are used to create the translation files (like the one this comment is on) that should deal with the complexities of this and help us to decide on the best solution.

With an appropriately generated file the gettext application that we'll be using will also be intelligent enough to handle the changing of keys (i.e. English in templates) when it does it's merges and the poedit program will flag changes to an editor. Although there are some sticking points by not using 'template.context.purpose' I think the benefits currently outweigh the downsides *if* we use the tooling right. 

Namely:
 1) The templates remain readable by devs and code reviewers without having to constantly refer to a translation lookup sheet.
 2) Given that, the translations can contain (small amount of) [markup](http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html#sec-pomarkup) reducing the amount of translation tagging we have to do and increasing template readability further.

The intention is to have a pipeline step that fails if translation conditions aren't right; either there are new untranslated strings or changed strings. This should be fairly easy to achieve and will stop bad text sneaking through.

PR #477, opened 2020-08-05
Uml 901 change error messages

kate-49:
## Purpose
https://opgtransform.atlassian.net/browse/UML-901?atlOrigin=eyJpIjoiMjdhNWU1ZWZmMDIzNDUwYmIyNTlkZGU3YTY1ODhhYzUiLCJwIjoiaiJ9 
https://docs.google.com/document/d/1tcbEr1OWGg7LgGUYpqcxbl95AWdthjByMhUtmS8K8bQ/edit?ts=5ef9c73e#

Fixes UML-901

## Approach
Changes / standardises error messages across 7 pages. 
Some content changes for standardisation.
Raises a specific error message for LPA reference number being too long or too short. 
Raises a specific error message depending on what field of date is missing - day, month or year. I created 3 new types of date/ dob validation to do this.
Amended tests to check for new content and added tests to cover the specific day/ month/ year incomplete validation. This also now checks that an empty date error is thrown if user doesn't enter anything into these fields.

## Checklist


service-front/app/src/Common/src/Validator/DateValidator.php
```diff
@@ -14,14 +14,18 @@ class DateValidator extends AbstractValidator
...
+    const DAY_INCOMPLETE      = 'dayIncomplete';
+    const MONTH_INCOMPLETE    = 'monthIncomplete';
+    const YEAR_INCOMPLETE     = 'yearIncomplete';
 
     protected $messageTemplates = [
         self::DATE_INVALID_FORMAT => 'Date value must be provided in an array',
         self::DATE_EMPTY          => 'Enter a date',
-        self::DATE_INCOMPLETE     => 'Date must include a day, month and year',
         self::DATE_INVALID        => 'Enter a real date',
+        self::DAY_INCOMPLETE         => 'Date must include a day, month and year',
```
williamfalconeruk, line 26: I wonder if these 3 message templates defaults are correct. probably be more specific in the intent of the messages e.g (like the dob validator, but just saying something like: `Date must include a day` for example...

service-front/app/test/CommonTest/Validator/DateValidatorTest.php
```diff
@@ -111,7 +111,7 @@ public function testIsValidIncompleteFormat()
         ]);
 
         $this->assertEquals([
-            DateValidator::DATE_INCOMPLETE => 'Date must include a day, month and year',
+            DateValidator::DAY_INCOMPLETE => 'Date must include a day, month and year',
```
williamfalconeruk, line 114: referring to my previous comment on this, if we change the message template then change this test.

williamfalconeruk: you could also cover the `MONTH_INCOMPLETE` and `YEAR_INCOMPLETE` test cases here too

service-front/app/src/Common/src/Validator/DateValidator.php
kate-49: So because i was removing the DATE_INCOMPLETE validator I left all 3 returning that message in the validator and only changed it to the specific validator under dob. I wasn't sure if the DATE_INCOMPLETE was used throughout the program and I was worried I might be making error changes to other pages that weren't specified on the ticket. I'm happy to change them to more specific ones if you think that's a better move.

service-front/app/test/CommonTest/Validator/DateValidatorTest.php
kate-49: That makes sense

service-front/app/src/Common/src/Validator/DateValidator.php
```diff
@@ -14,14 +14,18 @@ class DateValidator extends AbstractValidator
 {
     const DATE_INVALID_FORMAT = 'dateInvalidFormat';
     const DATE_EMPTY          = 'dateEmpty';
-    const DATE_INCOMPLETE     = 'dateIncomplete';
```
williamfalconeruk, line 17: this has been removed, so i assume that this doesn't exist elsewhere...worth doing a search to check

service-front/app/src/Common/src/Validator/DateValidator.php
```diff
@@ -14,14 +14,18 @@ class DateValidator extends AbstractValidator
...
     const DATE_INVALID        = 'dateInvalid';
+    const DAY_INCOMPLETE      = 'dayIncomplete';
+    const MONTH_INCOMPLETE    = 'monthIncomplete';
+    const YEAR_INCOMPLETE     = 'yearIncomplete';
 
     protected $messageTemplates = [
         self::DATE_INVALID_FORMAT => 'Date value must be provided in an array',
         self::DATE_EMPTY          => 'Enter a date',
-        self::DATE_INCOMPLETE     => 'Date must include a day, month and year',
```
williamfalconeruk, line 23: this has been removed, so i assume that this doesn't exist elsewhere...worth doing a search to check...

service-front/app/src/Common/src/Validator/DateValidator.php
williamfalconeruk: i'm coming from the point of view of the end user in terms of consistency - , so i'd check with the business just to see if they would be happy as is or a specific message for each.

kate-49: I'll check with lewis, thanks Billy

PR #482, opened 2020-08-06
Fix to identity generation

cooperaj:
## Purpose
The algorithm that generated the identity key that is given to users was incorrect and quite fundamentally broken. This PR addresses that by reworking the algorithm to actually output a useful value.

## Approach
The original algorithm used `implode` correctly on an incorrectly structured array. Resulting in a hash of the string "ArrayArrayArrayArrayArray". The array is now structured so that it contains the actual values. Something like;

`text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8gzip, deflateen-GB,en;q=0.5Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:78.0) Gecko/20100101 Firefox/78.0172.18.0.1`

## Learning
Always verify string inputs to functions.

**Although the tests have been updated to get this PR to pass additional tests need writing to check that the expected header values are a part of the hash. This can be done by listening to the logged out debug value in the test.**

## Checklist

### DIFF
--- a/service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
+++ b/service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
@@ -31,22 +31,27 @@ class UserIdentificationService
     public function id(\Psr\Http\Message\ServerRequestInterface $request): string
     {
         $headersToHash = [
-            'accept',
-            'accept-encoding',
-            'accept-language',
-            'user-agent',
-            'x-forwarded-for'
+            'accept' => '',
+            'accept-encoding' => '',
+            'accept-language' => '',
+            'user-agent' => '',
+            'x-forwarded-for' => ''
         ];
 
         // pull each header value out (if it exists)
-        foreach ($headersToHash as $header) {
+        foreach ($headersToHash as $header => $value) {
             $headersToHash[$header] =
                 $request->hasHeader($header)
-                    ? $request->getHeader($header)
+                    ? $request->getHeader($header)[0]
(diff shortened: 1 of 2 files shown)


service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
```diff
@@ -31,23 +31,28 @@ public function __construct(LoggerInterface $logger)
...
+            'x-forwarded-for' => ''
         ];
 
         // pull each header value out (if it exists)
-        foreach ($headersToHash as $header) {
+        foreach ($headersToHash as $header => $value) {
             $headersToHash[$header] =
                 $request->hasHeader($header)
-                    ? $request->getHeader($header)
+                    ? $request->getHeader($header)[0]
```
cooperaj, line 45: This is the key bit. Everything else is tidying up

service-front/app/src/Common/src/Service/Security/UserIdentificationService.php
```diff
@@ -31,23 +31,28 @@ public function __construct(LoggerInterface $logger)
...
                     : $header;
         }
 
         $this->logger->debug('Identity of incoming request built', $headersToHash);
 
+        $this->logger->debug(
+            'Identity of incoming request built',
+            ['prehash_id' => implode('', $headersToHash)]
```
cooperaj, line 53: A new test should be written that verifies that this value contains header values

PR #484, opened 2020-08-10
Uml 892 first time user

nickdavis2001:
## Purpose
_As a first time user I want to know what I can do with the service. So the What can I do with my LPAs should be Open unless they've already requested an Access code_

## Approach

_Pass the number of access codes created into the twig. In the twig, if its zero then open the Details. _

## Learning

_Ran into some difficulties due to the fact that steps have side-effects . For testing, tried originally using I Am On The Dashboard step, but this step contains code to tell the mock to return empty access codes. Learned from this to be careful trying to use existing steps as they may do unwanted things. Need to learn more about the mocking framework , had naively assumed that if you instruct a mock to behave in a certain way, it would always behave like that unless told otherwise_

## Checklist


service-front/app/features/context/UI/AccountContext.php
```diff
@@ -3478,7 +3480,30 @@ public function iAccessTheAccountCreationPage()
...
-         // Not needed for this context
+        // Not needed for this context
+    }
+
+    public function elementisOpen(string $searchStr)
+    {
+        $page = $this->ui->getSession()->getPage();
+        $element = $page->find('css', $searchStr);
+        $elementHtml = $element->getOuterHtml();
+        return str_contains($elementHtml, ' open>');
```
williamfalconeruk, line 3491: probably worth removing the closing angle bracket....

service-front/app/features/actor-dashboard.feature
```diff
@@ -39,6 +39,18 @@ Feature: The user is able to see correct information on their dashboard
...
+  @ui
+  Scenario: As a new user What Can I do with my LPAs is open
+    Given I have added an LPA to my account
+    When I am on the dashboard page
+    Then I can see that the What I can do link is open
+
+  @ui
+  Scenario: As a user who has created access codes What Can I do with my LPAs is closed
+    Given I have added an LPA to my account
+    When I can see that my LPA has 2 active codes with expiry dates 2021-04-01T23:59:59+01:00 2021-04-01T23:59:59+01:00
```
cooperaj, line 51: What happens in the future when these dates are now in the past?

Is this test more that "I can see my LPA has 2 active codes that expire in the future"

service-front/app/src/Actor/templates/actor/lpa-blank-dashboard.html.twig
```diff
@@ -18,7 +18,7 @@
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-two-thirds">
                     <p class="govuk-body-l">Any lasting power of attorney (LPA) you add to your account will show on this page.</p>
-                    <details class="govuk-details" data-module="govuk-details">
+                    <details class="govuk-details" data-module="govuk-details" open>
```
cooperaj, line 21: Is this valid html?

service-front/app/src/Actor/src/Handler/LpaDashboardHandler.php
```diff
@@ -80,14 +81,15 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
             $lpas[$lpaKey]['actorActive'] = $lpaData['actor']['type'] === 'donor' || $lpaData['actor']['details']->getSystemStatus();
         }
-
         $lpas = $this->lpaService->sortLpasByDonorSurname($lpas);
 
         return new HtmlResponse($this->renderer->render('actor::lpa-dashboard', [
             'user' => $user,
-            'lpas' => $lpas
+            'lpas' => $lpas,
+            'accesscodes' => $totalCodes,
```
cooperaj, line 92: I think this should have a different name. 'accesscodes' is not descriptive enough to tell me what it is, which I think is the total number of active codes?

service-front/app/src/Actor/templates/actor/lpa-dashboard.html.twig
```diff
@@ -18,7 +18,7 @@
         <div class="govuk-grid-row">
             <div class="govuk-grid-column-two-thirds">
                 <p class="govuk-body-l">Any lasting power of attorney (LPA) you add to your account will appear on this page.</p>
-                <details class="govuk-details" data-module="govuk-details">
+                <details class="govuk-details" data-module="govuk-details" {% if accesscodes == 0 %} open {% endif %} >
```
cooperaj, line 21: Please use '===' in comparisons. The nature of PHP is that 0 == false == ''.

nickdavis2001: OK. :) thats kind of amusing, once you learn enough languages you can get mixed up with some having == as equality, and some having =,  now PHP adds === into the mix just for fun. ;)

service-front/app/features/context/UI/AccountContext.php
nickdavis2001: Sure.  I left the closing bracket just thinking, could 'open' appear elsewhere for some reason, but, seems unlikely, so , changing it now

service-front/app/src/Actor/src/Handler/LpaDashboardHandler.php
nickdavis2001: Ok, am renaming to active_access_codes_count,   on the assumption that twig doesn't allow camelcase

service-front/app/src/Actor/templates/actor/lpa-blank-dashboard.html.twig
nickdavis2001: How I came up with this was simply, if you go to the page and open up the details by hand, with inspector running in Chrome, then open appears just before closing tag. I therefore replicated the behaviour that was already there, explicitly, to force it open. Looks to be valid. See https://www.w3schools.com/tags/att_open.asp

cooperaj: Learn something new every day :)

service-front/app/src/Actor/templates/actor/lpa-dashboard.html.twig
cooperaj: PHP's roots mean that a lot of operations by default are less strict. If you want more strictness it's something you have to specify, in this case '===' is a strict comparison (value and type) whereas '==' is just a value comparison. There's quite the number of idiosyncrasies, fortunately this one will have a better fix in version 8 https://wiki.php.net/rfc/string_to_number_comparison

nickdavis2001: Thanks for helpful clarification :)

service-front/app/features/actor-dashboard.feature
nickdavis2001: This is more that given there are 2 active codes, we see that the info link has closed. I've now changed this to 2041 to stop this test from failing next year. ;) And changed another tests, from which I stole the idea, to 2041 as well so that that one doesn't break either.  Now pushing this change, but having issues after merging in master to this branch

nickdavis2001: Will make a story on backlog to tidy up hard-coded dates

PR #490, opened 2020-08-11
Uml 292 add lpa flash message

GemTay:
## Purpose
Displays a flash message on the dashboard page after the user has added an LPA

Fixes UML-292

## Approach

Installed and included mezzio flash middleware. Added the donor's full name and lpa type to the session so that it can be used in the post request to complete the adding of the lpa. The flash message is set the check lpa handler on POST request before the user is redirected to the dashboard page. The dashboard page handler takes the flash object and passes it to the twig template where the message is rendered if the flash message key is set

PENDING Behat test

## Checklist

### DIFF
--- a/.idea/php.xml
+++ b/.idea/php.xml
@@ -5,6 +5,18 @@
       <PhpCSConfiguration beautifier_path="$PROJECT_DIR$/service-front/app/vendor/squizlabs/php_codesniffer/bin/phpcbf" standards="MySource;PEAR;PSR1;PSR12;PSR2;Squiz;Zend" tool_path="$PROJECT_DIR$/service-front/app/vendor/squizlabs/php_codesniffer/bin/phpcs" />
     </phpcs_settings>
   </component>
+  <component name="PhpIncludePathManager">
+    <include_path>
+      <path value="$PROJECT_DIR$/vendor/composer" />
+      <path value="$PROJECT_DIR$/vendor/psr/http-message" />
+      <path value="$PROJECT_DIR$/vendor/psr/http-server-middleware" />
+      <path value="$PROJECT_DIR$/vendor/psr/http-server-handler" />
+      <path value="$PROJECT_DIR$/vendor/laminas/laminas-zendframework-bridge" />
+      <path value="$PROJECT_DIR$/vendor/mezzio/mezzio-flash" />
+      <path value="$PROJECT_DIR$/vendor/mezzio/mezzio-session" />
+      <path value="$PROJECT_DIR$/vendor/psr/container" />
+    </include_path>
+  </component>
   <component name="PhpInterpreters">
     <interpreters>
       <interpreter id="d0145a1b-f903-43c7-8428-a985293fe8b9" name="api-app" home="docker-compose://DATA" debugger_id="php.debugger.XDebug" debugger_extension="/usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so">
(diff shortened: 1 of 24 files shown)


.idea/php.xml
```diff
@@ -5,6 +5,18 @@
...
+      <path value="$PROJECT_DIR$/vendor/composer" />
+      <path value="$PROJECT_DIR$/vendor/psr/http-message" />
+      <path value="$PROJECT_DIR$/vendor/psr/http-server-middleware" />
+      <path value="$PROJECT_DIR$/vendor/psr/http-server-handler" />
+      <path value="$PROJECT_DIR$/vendor/laminas/laminas-zendframework-bridge" />
+      <path value="$PROJECT_DIR$/vendor/mezzio/mezzio-flash" />
+      <path value="$PROJECT_DIR$/vendor/mezzio/mezzio-session" />
+      <path value="$PROJECT_DIR$/vendor/psr/container" />
+    </include_path>
+  </component>
```
Raffers: To remove all of this, you can do it via preferences > languages > php. As if you even remove it via the file (even with php storm shutdown) it'll still re-add these.

service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
```diff
@@ -117,6 +121,12 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                         );
 
                         if (!is_null($actorCode)) {
+                            /** @var FlashMessagesInterface $flash */
+                            $flash = $request->getAttribute(FlashMessageMiddleware::FLASH_ATTRIBUTE);
+                            $donor = $session->get('donor_name');
+                            $lpaType = $session->get('lpa_type');
+                            $flash->flash(self::ADD_LPA_FLASH_MSG, 'You\'ve added ' . $donor . '\'s ' . $lpaType . ' LPA' );
```
cooperaj, line 128: This should probably look like either:
`"You've added $donor 's $lpaType LPA"`
or
`sprintf('You\'ve added %s\'s %s LPA`, $donor, $lpaType);`

cooperaj: Either will help the translation effort when we get to these sorts of things

service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
```diff
@@ -155,6 +165,11 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                             'uId' => $referenceNumber
                         ]
                     );
+
+                    // data to be used in flash message
+                    $session->set('donor_name', $lpa->getDonor()->getFirstname() . ' ' . $lpa->getDonor()->getSurname());
```
cooperaj, line 170: Feels like there should be a nicer way to get this information when needed - rather than using the session

service-front/app/src/Actor/templates/actor/lpa-dashboard.html.twig
```diff
@@ -9,6 +9,16 @@
     {{ include('@partials/account-bar.html.twig') }}
 
     <main class="govuk-main-wrapper" id="main-content" role="main">
+
+        {% if flash.getFlash(constant('Actor\\Handler\\CheckLpaHandler::ADD_LPA_FLASH_MSG')) %}
```
cooperaj, line 13: is it possible to generalise this in it's own partial? flash messages, if we use them as intended, can appear on any page.

PR #494, opened 2020-08-12
UML-1027 - Upgrade Account Terraform

andrewpearce-digital:
## Purpose

Upgrade environment configuration to Terraform 0.13.0

Fixes UML-1027

## Approach

Follow the upgrade guides for the new version.
Improve the way that providers are sourced for environment level terraform
Remove private beta links from Slack messages (these are still available as outputs of the environment apply jobs)
Add Terraform outputs for environment container_version and workspace_name so that they are easier to get from CircleCI job output.

## Learning

https://www.terraform.io/upgrade-guides/0-13.html

## Checklist


terraform/account/credentials.tf
```diff
@@ -7,6 +9,21 @@ terraform {
     role_arn       = "arn:aws:iam::311462405659:role/opg-use-an-lpa-ci"
     dynamodb_table = "remote_lock"
   }
+
+  required_providers {
+    aws = {
+      source  = "-/aws"
```
williamfalconeruk, line 15: just interested... the namespace has been taken out, what's the reason for this?

andrewpearce-digital: Hashicorp have updated the way that providers are namespaced. I noticed during the account changes that tf init installed both v2.70.0 and v3.1.0 providers because required_providers and providers both called for aws. This change ensures only one is downloaded and used.

PR #505, opened 2020-08-17
Uml 995 twig extension

cooperaj:
## Purpose
This change implements the first part of the twig extension we need to translate pages. It adds translations for the start pages of both services

Fixes UML-995

## Approach
I realised we already have access to a factory to create the translation component so that was removed.

This mostly pulls code from the twig extension that is a part of the symfony/twig-bridge package with some minor alterations to get it working with our translation packages.

## Learning
The code to read twig tags is simple but quite verbose.

## Checklist


service-front/app/src/Common/src/Middleware/I18n/SetLocaleMiddleware.php
```diff
@@ -10,20 +10,23 @@
...
 
-    public function __construct(UrlHelper $helper, string $defaultLocale = null)
+    public function __construct(UrlHelper $helper, Translator $translator, string $defaultLocale = null)
     {
         $this->helper = $helper;
         $this->defaultLocale = $defaultLocale;
+        $this->translator = $translator;
     }
 
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
GemTay, line 32: Does this middleware need to be added to the pipeline? I can't see where it's added in here

cooperaj: That was done in the initial PR ;)

PR #508, opened 2020-08-18
changed aria attribute, now doesn't get flagged by ax extension

kate-49:
# Purpose

Fixes UML-939
https://opgtransform.atlassian.net/browse/UML-839?atlOrigin=eyJpIjoiNjY0ZjNjNzA1MTNlNDBhM2E3OTQ1YWQ0MDZjMzQ2ZjkiLCJwIjoiaiJ9

## Approach

Added an aria attribute to the organisation, this means that the access code name is associated with aria-controls="accordion-default-content-4".

I have tested this with the ax chrome extension. The attached screenshot shows the error is no longer showing. 

## Checklist

* [X ] I have performed a self-review of my own code
Screenshot 2020-08-18 at 11 04 52


service-front/app/src/Actor/templates/actor/partials/check-code-details.html.twig
```diff
@@ -1,7 +1,7 @@
 <div class="govuk-accordion__section ">
     <div class="govuk-accordion__section-header">
         <h2 class="govuk-accordion__section-heading">
-        <span class="govuk-accordion__section-button" id="accordion-default-heading-4">
+        <span class="govuk-accordion__section-button" id="accordion-default-heading-4" aria-controls="accordion-default-content-4">
```
johnnolan, line 4: I am not sure this is the right fix, I think it is more to do with the naming of the id. This refers to line 9 in the file and there is a chance that the id is duplicated elsewhere. Ideally it would have some sort of unique id per instance of the partial or a count that increments the number. Probably worth a quick call to discuss

johnnolan, line 4: Can we change this so it doesn't have a number on, so its just `accordion-default-heading` and `accordion-default-content` throughout the file?

johnnolan: In fact https://opgtransform.atlassian.net/browse/UML-841 will probably cover this as we need to pass a new parameter in to prevent the clashing of duplicate ids so ignore for now

kate-49: @johnnolan should I merge it or just ditch this work? I looked at the other tickets but I think they were a bit above my level tbh.

johnnolan: Definitely keep it, all approved from me, it solves the issue in the ticket and then I am happy to pair if you want on the other ones

PR #510, opened 2020-08-19
Created twig extension to populate currentLocale on default template.

Raffers:
# Purpose

Add relevant meta information for additional & default language into default twig template. Auto-populating html lang attribute based on language site is accessed in.

Fixes UML-1002

## Approach

Created twig extension to populate twig global var to set html lang attribute. 

## Learning

## Checklist

### DIFF
--- a/service-front/app/src/Common/src/ConfigProvider.php
+++ b/service-front/app/src/Common/src/ConfigProvider.php
@@ -92,6 +92,7 @@ class ConfigProvider
                     View\Twig\SymfonyTranslationExtensionFactory::class,
 
                 View\Twig\JavascriptVariablesExtension::class => View\Twig\JavascriptVariablesExtensionFactory::class,
+                View\Twig\GenericGlobalVariableExtension::class => View\Twig\GenericGlobalVariableExtensionFactory::class,
             ],
 
             'delegators' => [
--- a/service-front/app/src/Common/src/ConfigProvider.php
+++ b/service-front/app/src/Common/src/ConfigProvider.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Common;
 
+use Common\View\Twig\LanguageLocaleVariablesExtension;
+
 /**
  * The configuration provider for the Common module
  *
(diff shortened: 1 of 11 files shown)


williamfalconeruk (approved):
👍

service-front/app/src/Common/src/ConfigProvider.php
```diff
@@ -126,7 +128,8 @@ public function getTwig(): array
                 View\Twig\GovUKLaminasFormErrorsExtension::class,
                 View\Twig\GovUKLaminasFormExtension::class,
                 View\Twig\JavascriptVariablesExtension::class,
-                \Symfony\Bridge\Twig\Extension\TranslationExtension::class
+                View\Twig\LanguageLocaleVariablesExtension::class,
```
cooperaj, line 131: I'd possibly say that this could be made more generic. Like a "GlobalVariableExtension" since we may end up needing more global things and this would be a good place. Other than that looks solid 👍

cooperaj: That works. I'd not worry so much about dumping things in here, I don't think we're going to need too many global values but we want to minimise the extensions possible so that we don't slow down the twig too much.

service-front/app/src/Common/templates/layout/default.html.twig
```diff
@@ -16,6 +16,10 @@
     <link rel="apple-touch-icon" href="{{ asset('assets/images/govuk-apple-touch-icon.png') }}">
     <link rel="stylesheet" href="{{ asset('stylesheets/main.css') }}">
 
+    <link rel="alternate" hreflang="en-gb" href="https://use-lasting-power-of-attorney.service.gov.uk/home" />
+    <link rel="alternate" hreflang="cy-gb" href="https://use-lasting-power-of-attorney.service.gov.uk/cy/home" />
```
johnnolan, line 20: Will the use/view url part need to be dynamic too?

williamfalconeruk: ooh good spot!

johnnolan (approved):
Glorious.

williamfalconeruk (approved):

PR #511, opened 2020-08-20
Uml-292 Added lpa flash message (2)

GemTay:
# Purpose

Displays a flash message on the dashboard page after the user has added an LPA

Fixes UML-292

## Approach

Installed and included mezzio flash middleware. Added the donor's full name and lpa type to the session so that it can be used in the post request to complete the adding of the lpa. The flash message is set the check lpa handler on POST request before the user is redirected to the dashboard page. The dashboard page handler takes the flash object and passes it to the twig template where the message is rendered if the flash message key is set

Also refactored the CheckLpaHandler, splitting the get and post requests into separate handle functions

## Checklist

### DIFF
--- a/service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
+++ b/service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
@@ -83,6 +83,24 @@ class CheckLpaHandler extends AbstractHandler implements CsrfGuardAware, UserAwa
      */
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
+        $this->session = $this->getSession($request, 'session');
+
+        $this->form = new LpaConfirm($this->getCsrfGuard($request));
+
+        $this->user = $this->getUser($request);
+        $this->identity = (!is_null($this->user)) ? $this->user->getIdentity() : null;
+
+        $this->passcode = $this->session->get('passcode');
+        $this->referenceNumber = $this->session->get('reference_number');
+        $this->dob = $this->session->get('dob');
+
+        if (!isset($this->identity) || !isset($this->passcode) || !isset($this->referenceNumber) || !isset($this->dob)) {
+            // We don't have a code so the session has timed out
+            // TODO this can be reached if the session is still perfectly valid but the lpa search/response
+            //      failed in some way. Make this better.
+            throw new SessionTimeoutException();
+        }
+
         switch ($request->getMethod()) {
             case 'POST':
                 return $this->handlePost($request);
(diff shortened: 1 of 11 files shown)


GemTay:
image

service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
```diff
@@ -83,6 +82,18 @@ public function __construct(
...
+            case 'POST':
+                return $this->handlePost($request);
+            case 'GET':
+                return $this->handleGet($request);
+            default:
+                return $this->handleGet($request);
+        }
+    }
+
+    public function handleGet(ServerRequestInterface $request): ResponseInterface
```
GemTay, line 96: This block of code creating the variables and checking they are set is repeated in both the handleGet and handlePost functions. Currently looking into finding a neater solution for both functions to access these variables

Raffers: I think moving this block of code into the main handle function would clean up a lot of it. Setting them as properties?

```
        $session = $this->getSession($request, 'session');

        $form = new LpaConfirm($this->getCsrfGuard($request));

        $user = $this->getUser($request);
        $identity = (!is_null($user)) ? $user->getIdentity() : null;

        $passcode = $session->get('passcode');
        $referenceNumber = $session->get('reference_number');
        $dob = $session->get('dob');

        if (!isset($identity) && !isset($passcode) && !isset($referenceNumber) && !isset($dob)) {
            // We don't have a code so the session has timed out
            // TODO this can be reached if the session is still perfectly valid but the lpa search/response
            //      failed in some way. Make this better.
            throw new SessionTimeoutException();
        }
```

Also you could remove the return below get & still have it cascade down to `handleGet` within default.

Raffers (approved):

SeemaMenon (approved):

service-front/web/src/scss/_lpa-flash-message.scss
```diff
@@ -0,0 +1,18 @@
+.lpa-flash__box {
```
johnnolan, line 1: This can also be written like this taking advantage of SCSS

`
.lpa-flash {
  &__box {
    border:  govuk-spacing(1) solid govuk-colour('green');
    padding: 10px;
    margin-bottom: 30px;
  }

  &__message {
    display: inline-block;
    vertical-align: top;
    margin: 0px 0px 0px 0px;
  }

  &__icon {
    fill: govuk-colour("green");
  height: 25px;
  width: 25px;
  margin-right: 10px;
  }
}`

Raffers (approved):

SeemaMenon (approved):

PR #515, opened 2020-08-21
Uml 1000 add twig extension code

williamfalconeruk:
# Purpose

This is a DRAFT PR that covers the translation ticket UML-1000.

There is additional code added to the error extension which maps error validation messages, that can be translated added into twig files at the top.
The format aligns roughly so far with what is seen in the error messages that come out of the `$form->getMessages` call.

I ended up writing a custom replacement function to get the strings to map over. This could possibly be improved on, but it mostly works, though YMMV, as it's incomplete, and fails 4 tests right now.

The idea is that we place the validation messages inside the twig files, so we can extract the automatically as part of the translation file generation.
 
To add validation see the example twigs. the format is roughly:

```twig
{{
translate_validation_messages(form, 
 {
            'field_name1': // fieldname
            { //all contained validation messages..
                'validation1'      :   'this is my error 1' | trans,
                'someOtherValidation' : 'this is another error' | trans
            },
            'field_name2':
            {
                'isEmpty'     : 'field 2 is empty' | trans
            }
//..etc
        }
}}
```

## Still to be done:
- fix any test breaks
- Use class constants to reference the messages  and introspection/reflection to work the keys out.
- Further work as mentioned previously to go through each form and place the validation messages in the prescribed format onto the twig files. the examples below are incomplete but should be functional enough.
- further testing on different forms.

## Checklist

### DIFF
--- a/service-front/app/src/Actor/src/Form/ChangeEmail.php
+++ b/service-front/app/src/Actor/src/Form/ChangeEmail.php
@@ -74,11 +74,6 @@ class ChangeEmail extends AbstractForm implements InputFilterProviderInterface
                     [
                         'name'                   => EmailAddressValidator::class,
                         'break_chain_on_failure' => true,
-                        'options'                => [
-                            'messages' => [
-                                EmailAddressValidator::INVALID => 'Enter an email address in the correct format, like name@example.com',
-                            ],
-                        ],
                     ]
                 ],
             ],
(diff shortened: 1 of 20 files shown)


SeemaMenon (approved):

SeemaMenon (commented):
LGTM....will request another 2 pair of eyes to go through.

PR #517, opened 2020-08-24
UML-984 - I need access to session endpoints that allow me monitor my sessions

Raffers:
# Purpose

Creates endpoints to check time remaining on session & refresh the session without navigating away from page.

Fixes UML-984

## Approach

Adjusts initializeSessionFromRequest & persistSession on EncryptedCookiePersistence. It first stores any session information along with the request path. We then look at if the session is being checked & use the original session time if we are targeting the check endpoint.

## Learning

## Checklist


service-front/app/src/Common/src/Handler/SessionCheckHandler.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    use SessionTrait;
+
+    private ContainerInterface $container;
+
+    /**
+     * SessionCheckHandler constructor.
+     * @param ContainerInterface $container
+     */
+    public function __construct(ContainerInterface $container)
```
cooperaj, line 26: Please don't inject the container anywhere. If access to things from the container is needed then inject those values from a factory class that pull from the container. As a general rule the only classes with access to the container are factory classes.

service-front/app/src/Common/src/Service/Session/EncryptedCookiePersistence.php
```diff
@@ -232,8 +239,15 @@ public function initializeSessionFromRequest(ServerRequestInterface $request): S
 
     public function persistSession(SessionInterface $session, ResponseInterface $response): ResponseInterface
     {
-        // Record the current time.
-        $session->set(self::SESSION_TIME_KEY, time());
+        // Checking session or setting current time
+        $time = isset($this->originalSessionData[self::SESSION_TIME_KEY])
```
cooperaj, line 243: Would prefer to see this check in a separate method.

I think the check of the url needs figuring out a different way as well. Is the request url not available in the ResponseInterface? Can we not store the session time instead of the entire session? Is it possible to avoid the persistSession  method call entirely when accessing this endpoint?

cooperaj (changes requested):

service-front/app/src/Common/src/Service/Session/EncryptedCookiePersistence.php
Raffers: It's only available as a header, which would need additional formatting to get it to the same state. But it's already in the required state in the request object. 
I was thinking in case anything else was needed for the session, but I've changed that now. 
I did initially look at trying to bypass persistSession entirely but couldn't find a nice way to do it without going too far down into the framework.

service-front/app/test/CommonTest/Handler/SessionCheckHandlerTest.php
```diff
@@ -0,0 +1,76 @@
+<?php
```
cooperaj, line 1: Handlers we typically mark as @codecoverageignore and rely on behat to test. If you think there's benefit in having this thats also good (due to the json based nature).

service-front/app/src/Common/src/Service/Session/EncryptedCookiePersistence.php
```diff
@@ -360,4 +367,14 @@ private function getCookieLifetime(SessionInterface $session): int
 
         return $lifetime > 0 ? $lifetime : 0;
     }
+
+    private function sessionTime(): int
+    {
+        // Checking session or setting current time
+        return isset($this->originalSessionTime)
```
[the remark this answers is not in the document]
Raffers: I was getting Property access error's for some reason before, I've just removed it after doing null coalescing check & it's now passing.

service-front/app/test/CommonTest/Handler/SessionCheckHandlerTest.php
Raffers: Yer, that was my thinking. As we use Behat as if it was a user performing the action, I thought it best to do Check & Refresh as unit tests due to more being an endpoint that is checked automatically.

williamfalconeruk (dismissed):

GemTay (approved):

cooperaj (approved):

PR #534, opened 2020-08-28
UML-974 Translation switch - DO NOT MERGE YET

GemTay:
# Purpose

Adds a UI switch to enable users to switch between english and welsh

Fixes UML-974

## Approach

Created a common partial twig template to match the prototype. The switch checks whether the current url contains the welsh prefix, and if so displays "Cymraeg" as text rather than a link - and vice versa for the english link. It adds a /cy prefix to the welsh link and removes it from the english link

## Checklist


service-front/app/src/Common/templates/partials/welsh-switch.html.twig
```diff
@@ -0,0 +1,21 @@
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        <ul class="trans-switch">
+            <li class="govuk-list trans-switch__list govuk-body-s">
+                {% if path() matches '/(\\/cy.*)/' %}
```
cooperaj, line 5:
```suggestion
                {% if path() matches '/(^\\/cy.*)/' %}
```
Might need this or something like /lpa/cymraeg.html will match.

GemTay: Good spot - fixed :)

Raffers (dismissed):

williamfalconeruk:
I thin this would have been better to be put into a base layout page, rather than as an include on each of the pages in the service. Or this might just be my understanding of twig...

williamfalconeruk (dismissed):

SeemaMenon (approved):
LGTM

SeemaMenon:
> I think this would have been better to be put into a base layout page, rather than as an include on each of the pages in the service. Or this might just be my understanding of twig...

I too did think the same actually !

cooperaj (approved):
LGTM

SeemaMenon (approved):

PR #539, opened 2020-09-02
UML-1045 banner message that only LPAs after a certain date can use the service

nickdavis2001:
# Purpose

_UML-1045 banner message that only LPAs after a certain date can use the service_

## Approach

_Add a new twig with the banner, include this in the triage page, behat test_

## Learning

__

## Checklist


service-front/web/src/scss/_banner.scss
```diff
@@ -0,0 +1,51 @@
...
+
+.moj-banner__message h2:last-child,
+.moj-banner__message p:last-child {
+  margin-bottom: 0;
+}
+
+/* Style variants
+   ========================================================================== */
+
+.moj-banner--success {
```
Raffers, line 42: shouldn't thie be `__` in place of `--`

Also, have a read of John's comment on a previous PR. The structure of this can be changed - https://github.com/ministryofjustice/opg-use-an-lpa/pull/511#discussion_r475496095

Raffers (changes requested):

service-front/web/src/scss/_banner.scss
nickdavis2001: Thanks for feedback. :) This was a direct copy from Make An LPA, so, while there may well be flaws in this, its tried and tested in production already elsewhere. I'm a little disinclined to fix this at this point in time, since its already passed UAT and we want to get Welsh translation out. How about if I make another story on the backlog to revisit this? (open to arguments to the contrary.  ;). )

Raffers (dismissed):

service-front/web/src/scss/_banner.scss
johnnolan: `--` is right but as Chris said, keeping it line with the other style files would be good. Given the conversation in standup, carry on deploying and create a little ticket to tidy this up. Happy to pair to show how you can restructure it and should be a 1 pointer

johnnolan (dismissed):

williamfalconeruk (approved):

PR #550, opened 2020-09-04
Uml 443 show password option

GemTay:
# Purpose

Allows users to toggle between entering their hidden password twice to confirm or entering it once and being able to see it as text

Fixes UML-443

## Approach

Most likely a poor attempt at writing javascript, but hey it works! The data-hideConfirmPassword variable is set to true so that the page is loaded with both password boxes visible. The data-for="password" is checked to ensure that we are making the correct element visible/hidden. 

I based this approach off the code in the make an LPA code base:
application.js from line 23531
application->general->register->index.twig line 96

## Learning

I learnt that I don't fully understand javascript that well and need to learn it haha :)

## Checklist


GemTay:
Screenshot 2020-09-04 at 13 51 32
Screenshot 2020-09-04 at 13 51 47

service-front/web/src/javascript/showHidePassword.js
```diff
@@ -0,0 +1,40 @@
+const showHidePassword = () => {
```
cooperaj, line 1: If we're going to be using this on just the one element (#showHidePasssword) it might be nicer to pass it in to a constructor like is done with cookieConsent?

service-front/web/src/javascript/showHidePassword.js
```diff
@@ -0,0 +1,40 @@
...
+            } else {
+                if (hideConfirm) {
+                    pwdConfirmParent.hidden = false;
+                    passwordElement.setAttribute('type', 'password');
+                    this.setAttribute('data-for', 'password');
+                }
+            }
+
+            //  Change the link text
+            link.textContent = (isShowing ? 'Hide password' : 'Show password');
```
cooperaj, line 33: This will not work with translation. Look at Johns recent PR as to how you can pass in strings to show using data attributes

GemTay:
Show/Hide button translates to welsh 
image
image

cooperaj (dismissed):

service-front/app/src/Common/templates/partials/govuk_form.html.twig
```diff
@@ -31,11 +31,26 @@
 
         {% endif %}
 
-        <input class="govuk-input {{- block('input_extra_class') -}} {{- block('input_error_class') -}}"
-                id="{{ element.getName() }}" name="{{ element.getName() }}" type="{{ type }}" value="{{ value }}"
-                {% if inputmode is defined %}inputmode="{{ inputmode }}"{% endif %}
-                {% if spellcheck is defined %}spellcheck="{{ spellcheck }}"{% endif %}
-                {% if autocomplete is defined %}autocomplete="{{ autocomplete }}"{% endif %} />
+        {% if element.getName() == 'show_hide_password' %}
```
GemTay, line 34: I don't particularly like this solution however it was the only way I could wrap the input and button inside a div with the class "moj-password-reveal" so that the html layout replicated the moj prototype https://moj-design-system.herokuapp.com/components/password-reveal/examples/default . Without this layout, the show/hide button is not rendered side by side with the input on all sized devices

cooperaj: You could make a custom type?

service-front/app/test/ActorTest/Handler/CreateAccountHandlerTest.php
```diff
@@ -129,11 +129,10 @@ public function testFormSubmittedCreateAccount()
             ->willReturn('POST');
         $this->requestProphecy->getParsedBody()
             ->willReturn([
-                '__csrf'           => self::CSRF_CODE,
-                'email'            => 'a@b.com',
-                'password'         => 'P@55word',
-                'password_confirm' => 'P@55word',
-                'terms'            => '1',
+                '__csrf'                => self::CSRF_CODE,
```
cooperaj, line 132: The handlertest should probably be removed (and @codecoverageingore added to the handler) I believe all these cases should be handled by the behat?

service-front/app/src/Common/templates/partials/govuk_form.html.twig
cooperaj: Or, at least, use that as a basis for the refactoring.

SeemaMenon (approved):

service-front/app/src/Common/templates/partials/govuk_form.html.twig
```diff
@@ -189,3 +179,33 @@
         </span>
     {% endif %}
 {%- endblock input_hint -%}
+
+{%- block form_input -%}
+    {% if element.getName() == 'show_hide_password' %}
```
cooperaj, line 184: I would say, in order to reduce repetition, that you could split this up so that the show_hide_password checks are done twice?

```
if show_hide_password
     div.moj-password-reveal

input

if show_hide_password
     button
     enddiv.moj-password-reveal
```

cooperaj (commented):
Something like this?

cooperaj (approved):
LGTM

SeemaMenon (approved):

PR #566, opened 2020-09-17
UML-1064: Removed inactive code duplication

SeemaMenon:
# Purpose

https://opgtransform.atlassian.net/browse/UML-1064

Fixes UML-1064

## Approach

## Learning

## Checklist

### DIFF
--- a/service-front/app/src/Actor/templates/actor/partials/check-code-details.html.twig
+++ b/service-front/app/src/Actor/templates/actor/partials/check-code-details.html.twig
@@ -40,7 +40,9 @@
                 {% endif %}
             </div>
 
-            {% if (check_if_code_has_expired(code.Expires) and check_if_code_is_cancelled(code)) %}
+            {# Checking for cancelled status first.
+            This would reduce logic to check and display cancelled status for codes that is cancelled and expired #}
+            {% if check_if_code_is_cancelled(code)  %}
                 <div class="govuk-summary-list__row">
                     <dt class="govuk-summary-list__key">{% trans %}Status{% endtrans %}</dt>
                     <dd class="govuk-summary-list__value">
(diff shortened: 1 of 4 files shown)


cooperaj (approved):

service-front/app/src/Actor/templates/actor/partials/check-code-details.html.twig
```diff
@@ -40,22 +40,27 @@
                 {% endif %}
             </div>
 
-            {% if check_if_code_has_expired(code.Expires) %}
+            {% if (check_if_code_has_expired(code.Expires) and check_if_code_is_cancelled(code)) %}
```
williamfalconeruk, line 43: I wonder if you need both of these? as you have a check code is cancelled already in the `elseif` part...

williamfalconeruk: possibly swap the logic around: so you have the cancelled check first, then do the expiry check as the last thing. that might be simpler - unless i've missed something...

williamfalconeruk (changes requested):
think the logic needs to be simplified here.

Requirement states:
Screenshot 2020-09-21 at 13 21 36

service-front/app/src/Actor/templates/actor/partials/check-code-details.html.twig
SeemaMenon: @williamfalconeruk - I had to had in this logic because the requirement says;

**For an inactive code:** 
1. if an access is both cancelled and expired, then the status in the check access code page should show the status as CANCELLED.
2. If an an access code is cancelled, show the status as CANCELLED
3. If an access code is expired, show the status as EXPIRED.

Had to keep the conditions in the else if explicitly because otherwise the active code too will show a status [which is not needed and correct]

williamfalconeruk: to work this out i did a small bit of logic deduction on this. Looking at the rules: 

- `CANCELLED` will need to appear whether or not your code is expired, so you just could put `check_if_code_is_cancelled` as your first logic check. 
-the `elseif` part then will catch all non cancelled and then you can `check_if_code_is_expired` at that point. 

Hope this clarifies 🙂

williamfalconeruk (approved):
LGTM 👍

nickdavis2001 (approved):

PR #567, opened 2020-09-17
Uml 543 print styling 2

DanGuyMOJ:
# Purpose

Add print stylesheet. Add image to temporary directory because the correct directory wasn't showing up locally so this needs to be moved to the correct folder.

Fixes UML-543

## Checklist


service-front/web/src/scss/_lpa-print-styling.scss
```diff
@@ -0,0 +1,104 @@
...
+        margin-bottom: 0;
+      }
+      .govuk-panel__body p {
+        color: #000000;
+        font-size: 14pt !important;
+        margin-bottom: 5px !important;
+      }
+  }
+  .govuk-panel--confirmation {
+    border: 4px solid #00703c;
```
johnnolan, line 66: Instead of using the hex value throughout the document you can use the built in GDS ones. See other scss files for an example. That way updates in the future will auto update this file.

Also may be worth looking at the margin and padding values which have their own settings

DanGuyMOJ: Cheers @johnnolan. I'll swap those out.

johnnolan (approved):

SeemaMenon (approved):

PR #596, opened 2020-10-19
Uml1105 head request uses activation token

cooperaj:
# Purpose

The Microsoft SafeURL service sends a preemptive HEAD request to a url to check that it is "safe" before sending its users on. The previous piece of code that we had did not distinguish between HEAD and GET and so was activating the account before the user actually visited the activation url. This mean't they would get the failure message

This ensures we respond to HEAD requests appropriately.

Fixes UML-1105

## Approach

There is a Mezzio middleware that is responsible for managing head requests that this request hits. This middleware attaches and attribute to the request before forwarding it on to the handler. It then strips the body content away before allowing the final dispatch to the user. By using that attribute we can stop the service from carrying out actions when accessed with a HEAD request.

## Learning

This will likely impact any part of the application that can be deep-linked to that carries out an action (like changing of passwords. This will need fixing for all those cases.

## Checklist

### DIFF
--- a/service-front/app/features/context/UI/AccountContext.php
+++ b/service-front/app/features/context/UI/AccountContext.php
@@ -31,6 +31,7 @@ use Psr\Http\Message\RequestInterface;
  * @property $organisation
  * @property $newUserEmail
  * @property $userEmailResetToken
+ * @property $activationToken
  */
 class AccountContext implements Context
 {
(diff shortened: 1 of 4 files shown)


cooperaj:
Have added the HEAD request checking to the email change handler. It is not needed for the password change as we do not make any changes to data on access to that page.

GemTay (approved):

SeemaMenon (approved):

PR #598, opened 2020-10-19
Lewis-Cummings-content-fix

LewisCummings:
Changed "if you're names" to "If you're named"

# Purpose

Content was incorrect and picked up by a user


cooperaj:
This PR will cause Welsh translations of that phrase to fail. Additional work to update the language files will be needed.

PR #615, opened 2020-11-04
Add privacy notice link

cooperaj:
# Purpose

Fixes UML-1087

## Approach

Add link to footer and adjust order of existing links.

## Learning

This introduces some issues with the back link on the privacy notice page (where does it go back to, it's been opened in a new tab)

## Checklist

### DIFF
--- a/service-front/app/src/Common/templates/layout/default.html.twig
+++ b/service-front/app/src/Common/templates/layout/default.html.twig
@@ -102,12 +102,12 @@
                 <h2 class="govuk-visually-hidden">{% trans %}Support links{% endtrans %}</h2>
                 <ul class="govuk-footer__inline-list">
                     <li class="govuk-footer__inline-list-item">
-                        <a class="govuk-footer__link" href="{{ path('terms-of-use') }}" target="_blank">
+                        <a class="govuk-footer__link" href="{{ path('terms-of-use') }}">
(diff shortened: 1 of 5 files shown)


service-front/app/src/Common/templates/layout/default.html.twig
```diff
@@ -107,20 +107,25 @@
                         </a>
                     </li>
                     <li class="govuk-footer__inline-list-item">
-                        <a class="govuk-footer__link" href="{{ path('cookies') }}">
-                            {% trans %}Cookies{% endtrans %}
+                        <a class="govuk-footer__link" href="{{ path('privacy-notice') }}" target="_blank">
```
GemTay, line 110: From what i've read, accessibility requirements mention the need to warn users that the link will open in a new tab. On the create account page we tell the user with the link text "terms of use (opens in new tab)". Might be something to ask Sarah/Lewis/Dan about

GemTay (commented):
Just a small accessibility issue to fix, everything else looks great :)

service-front/app/src/Common/templates/layout/default.html.twig
cooperaj: Mmm, this is true. We dont yet do it for terms of use.

GemTay (approved):

PR #631, opened 2020-11-26
Uml 1125 user cannot add same lpa twice

GemTay:
# Purpose

Prevents users from adding the same LPA more than once

Fixes UML-1125

## Approach

Fetch all LPA's the user has added to their account, and loop over each LPA to check if its Uid is the same as the one they are attempting to add. If so, the user is displayed the new content to tell them that they have already added that LPA 

## Checklist


service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
```diff
@@ -139,6 +139,31 @@ public function handleGet(
             $lpa = $lpaData['lpa'];
             $actor = $lpaData['actor']['details'];
 
+            // check if they have already added this LPA
+            $lpasAdded = $this->lpaService->getLpas($this->identity);
```
cooperaj, line 143: Mmm. I think I would prefer to see this in the service itself. The ability to add an LPA (or not) should be determined there and the result of that handled in the handler. A handler should marshal various services but should never really do any *work*. Perhaps the service can have a function like

```
public function existsInAccount(string $referenceNumber, string $identity): bool;
```

cooperaj: btw, the name of the function is just a suggestion. I can come up with better ones already and I just wrote it.

GemTay: Thats true, good point - i've moved it into a new function in the service 👍

service-front/app/src/Common/src/Service/Lpa/LpaService.php
```diff
@@ -365,6 +365,33 @@ public function sortGroupedDonorsLpasByTypeThenAddedDate(ArrayObject $donors): A
...
+                    'Account with Id {id} has attempted to add LPA {uId} which already exists in their account',
+                    [
+                        'id' => $identity,
+                        'uId' => $referenceNumber
+                    ]
+                );
+                return $userLpaActorToken;
+            }
+        }
+        return '';
```
GemTay, line 392: @cooperaj I can't decide what the best return type should be here, and I don't really like to return an empty string, but it gives a consistent return type? Maybe null would be better? What do you think?

GemTay (commented):
Question...

service-front/app/src/Common/src/Service/Lpa/LpaService.php
cooperaj: I'd just make this a boolean - as it's a simple yes or no question that we're asking.

cooperaj: Because of that i'd call the function `isLpaAlreadyAdded`

service-front/app/src/Actor/src/Handler/CheckLpaHandler.php
```diff
@@ -139,6 +139,21 @@ public function handleGet(
             $lpa = $lpaData['lpa'];
             $actor = $lpaData['actor']['details'];
 
+            // check if they have already added this LPA
+            $lpaTokenAdded = $this->lpaService->isLpaAlreadyAdded($referenceNumber, $this->identity);
+
+            if ($lpaTokenAdded) {
```
cooperaj, line 145:
```suggestion
             if (false !== $lpaTokenAdded = $this->lpaService->isLpaAlreadyAdded($referenceNumber, $this->identity)) {
```

cooperaj: Think this should work.
1) nails down the the false check (in php 7 "fooBar" == 0 == false)
2) removes a line

cooperaj (approved):

PR #633, opened 2020-11-27
Uml 1104 codes content change

cooperaj:
# Purpose
Adds additional textual context to viewer code names for accessibility reasons

Fixes UML-1104

## Approach
Uses the govuk design systems accordion subtitle field to provide additional information. In order to get the date formatted correctly I had to play around with the LpaExtension and it's date handling.

I also fixed up the database seeding which was actually breaking fresh development environments with bad data.

## Learning
There're some dark corners of our app that need some TLC. For instance, creation timestamps against records in the database use a different format to basically every other date we store.

## Checklist

* [ ] ~I have added relevant logging with appropriate levels to my code~
* [ ] ~I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~


service-front/app/src/Common/src/Service/Lpa/ViewerCodeService.php
```diff
@@ -107,16 +107,13 @@ private function getNumberOfActiveCodes(ArrayObject $shareCodes): ?ArrayObject
...
-
-                if (new DateTime($code['Expires']) >= (new DateTime('now'))->setTime(23,59,59) &&
-                    !(array_key_exists("Cancelled",$code)) &&
-                    (!empty($code['UserLpaActor']))) {
-
-                    $counter += 1;
-                }
+        foreach ($shareCodes as $code) {
+            if (
+                new DateTime($code['Expires']) >= (new DateTime('now'))->setTime(23, 59, 59)
```
GemTay, line 112: How come we compare the code expiry to the end of the current day? Is it so that codes expire 30 days after they were created, regardless of the time of day? If so we'll have to do the same comparison in the checkAccessCodesHandler so that the number of active codes displayed on the dashboard matches the number on the check access codes page

cooperaj: I'm guessing so, I just tidied it up though

service-front/web/src/scss/lpa.scss
```diff
@@ -1,8 +1,6 @@
-$govuk-typography-use-rem: false;
```
GemTay, line 1: This was added by john to fix the bug that users were seeing on iOS 10.3 Safari, maybe best to keep it?

service-front/web/src/scss/lpa.scss
```diff
@@ -1,8 +1,6 @@
...
 @import "../../node_modules/govuk-frontend/govuk/settings/all";
 @import "../../node_modules/govuk-frontend/govuk/tools/all";
 @import "../../node_modules/govuk-frontend/govuk/helpers/all";
-@import "../../node_modules/govuk-frontend/govuk/core/all";
```
GemTay, line 5: How come this was removed?

GemTay (changes requested):
Few queries

service-front/web/src/scss/lpa.scss
cooperaj: Complex reasons. I guess I need to run this and the above past you and John. Monday?

Briefly, because of the build order of things, and that gds.css is included already (which contains all the above imports anyway) it was subtly breaking accordions with odd padding and margin values.

service-front/web/src/scss/lpa.scss
cooperaj: See below :)

service-front/app/src/Common/src/Service/Lpa/ViewerCodeService.php
GemTay: @cooperaj No problem - I just checked and replicated the issue locally, I'll raise a bug ticket, as if a code is expiring on the current day, the dashboard displays the LPA as having 1 active code, however when you then click to check access codes, the code is in the inactive section. So theres an inconsistency in the code

cooperaj: Assign it straight to me. I've already started work on the fix :)

service-front/web/src/scss/lpa.scss
GemTay: Makes sense 👍

GemTay (dismissed):

service-front/app/src/Common/src/Service/Lpa/ViewerCodeService.php
cooperaj: Spent Friday refactoring all this stuff into a ViewerCode object so we don't spread this kind of thing about.

service-front/web/src/scss/lpa.scss
cooperaj: @johnnolan this need looking at?

johnnolan: @cooperaj that needs to stay so we can support our use of the specific components and browser bugs unfortunately.

johnnolan: After chatting with @cooperaj I am happy for this to be removed. The main place we care about this being set is here https://github.com/ministryofjustice/opg-use-an-lpa/blob/2443a622a600e9f5787714095330da462e45053a/service-front%2Fweb%2Fsrc%2Fgds.scss which removes the rems correctly for the iOS bug

andrewpearce-digital (approved):
docker-compose.yml
service-api/docker/seeding/start.sh
service-api/seeding/dynamodb.py

All these look good to me. Someone else should approve the other files.

PR #634, opened 2020-12-01
UML-1191 Feature flag older LPA flow

GemTay:
# Purpose

Creates older LPA journey routes with a feature flag to turn the journey off until we say so

Fixes UML-1191

## Approach

Created comparison check against feature flag. If it is not true, the routes are not created. I created a handler, route and twig file for the triage page as a way to test the functionality works

## Checklist

### DIFF
--- a/service-front/app/config/routes.php
+++ b/service-front/app/config/routes.php
@@ -167,7 +167,7 @@ $actorRoutes = function (Application $app, MiddlewareFactory $factory, Container
     ], ['GET', 'POST'], 'add.by-code');
 
     // Older LPA journey
-    if ($container->get('config')['feature_flags']['use_older_lpa_journey'] === "true") {
+    if ($container->get('config')['feature_flags']['use_older_lpa_journey']) {
         $app->get('/add', [
             Mezzio\Authentication\AuthenticationMiddleware::class,
             Actor\Handler\AddLpaTriageHandler::class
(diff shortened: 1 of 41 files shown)


cooperaj (changes requested):
Looks pretty solid but I think we need to work the old 'lpa.add' links slightly differently - though maybe as another ticket.

If the feature is disabled all the codes should continue to work as it currently does, if we enable it then I think that any location that originally went to 'lpa.add' should now go to the triage page? But we wouldn't want to change the routes as the feature flag will handle what that does. I think that the 'lpa.add' route should continue to exist but the feature flagged routing should decide whether that goes to the current location or to the new triage page.

service-front/app/config/routes.php
```diff
@@ -163,6 +159,20 @@
...
     ], 'lpa.death-notification');
+
+    // Adding LPA
+    $app->route('/add/by-code', [
+        Mezzio\Authentication\AuthenticationMiddleware::class,
+        Actor\Handler\LpaAddHandler::class
+    ], ['GET', 'POST'], 'add.by-code');
+
+    // Older LPA journey
+    if ($container->get('config')['feature_flags']['use_older_lpa_journey'] === "true") {
```
cooperaj, line 170: Instead of adding a new route if the flag is enabled we should toggle between the original routes handler (maybe with the new name) and the triage handler.

if (flag off) {
lpa.add => LpaAddHandler
else {
lpa.add => TriageHandler
lpa.add.by-code => LpaAddHandler
lpa.add.by-paper => LpaAddOlderHandler (this won't exist yet)

cooperaj: Doing it this way means the code for all this can exist on live without being enabled by having a route hooked up to it.

cooperaj (approved):

PR #639, opened 2020-12-03
Uml 1191 fixed feature flag

GemTay:
# Purpose

Implements the feature flag correctly this time. Creates older LPA journey routes with a feature flag to turn the journey off until we say so

Fixes UML-1191

## Approach

The issue was caused as the variable was a string eg "false" which evaluated to true and passed the if statement. I've converted the variable to a boolean to fix the issue

## Learning

- Check what the variable is i'm evaluating against rather than assuming in future
- Get static analysis setup to catch errors like this

## Checklist

### DIFF
--- a/service-front/app/config/autoload/features.global.php
+++ b/service-front/app/config/autoload/features.global.php
@@ -4,6 +4,6 @@ declare(strict_types=1);
 
 return [
     'feature_flags' => [
-        'use_older_lpa_journey' => getenv('USE_OLDER_LPA_JOURNEY') ?: false,
+        'use_older_lpa_journey' => filter_var(getenv('USE_OLDER_LPA_JOURNEY'), FILTER_VALIDATE_BOOL) ?: false,
     ]
 ];
--- a/service-front/app/config/autoload/features.global.php
+++ b/service-front/app/config/autoload/features.global.php
@@ -4,6 +4,6 @@ declare(strict_types=1);
 
 return [
     'feature_flags' => [
-        'use_older_lpa_journey' => getenv('USE_OLDER_LPA_JOURNEY') ?: 'false',
+        'use_older_lpa_journey' => getenv('USE_OLDER_LPA_JOURNEY') ?: false,
     ]
 ];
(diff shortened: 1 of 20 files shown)


service-front/app/config/autoload/features.global.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+return [
+    'feature_flags' => [
+        'use_older_lpa_journey' => getenv('USE_OLDER_LPA_JOURNEY') ?: false,
```
cooperaj, line 7: I think I would prefer to see conversion to a bool here. Since we're already defaulting the value to `false` if it doesn't exist it makes sense to ensure that if it does exist then it would be a boolean

service-front/app/config/routes.php
```diff
@@ -163,6 +159,26 @@
         Mezzio\Authentication\AuthenticationMiddleware::class,
         Actor\Handler\DeathNotificationHandler::class
     ], 'lpa.death-notification');
+
+    // Older LPA journey
+    if (filter_var($container->get('config')['feature_flags']['use_older_lpa_journey'], FILTER_VALIDATE_BOOL)) {
```
cooperaj, line 164: Which would mean this line would just a be a normal boolean check.

service-front/app/features/bootstrap/behat.config.php
```diff
@@ -25,6 +25,10 @@
         //'debug'   => true,
     ],
 
+    'feature_flags' => [
+        'use_older_lpa_journey' => 'true',
```
cooperaj, line 29:
```suggestion
        'use_older_lpa_journey' => true,
```

service-front/app/config/autoload/features.global.php
GemTay: Yeah thats a much better place to have it, changes made 👍

cooperaj (approved):
Am happy with this. Hopefully we cant figure out a way to test it (preprod?)

PR #651, opened 2020-12-21
UML-1189 - Deploy images without dev dependencies

andrewpearce-digital:
# Purpose

Build and test the service, and deploy an environment that does not contain development dependencies.

Fixes UML-1189

## Approach

- Use docker build targets to create development images for unit and integration testing, and production images to be deployed to an environment.
- Rename pipeline jobs to remove redundant naming and make things neater
- Fix filter_var argument

image

* [x] Check docker caching behaviour on CircleCI
    * caches are used and available during rebuilds
* [x] Explore docker build targets option
    * Build targets work well on the pipeline for API App
    * Local build times are within an appropriate range of master having changed one Dockerfile

## Learning

https://docs.docker.com/develop/develop-images/multistage-build/

## Checklist


andrewpearce-digital:
Definite benefits from caching on docker builds
image

andrewpearce-digital:
A targetted build (which ran a composer install twice, once for dev deps and again for no dev deps but only copied in non-dev deps) completed in similar time to previous builds
image
It was expected at this stage to fail at unit testing as those dependencies were not available

andrewpearce-digital:
Building a production target after a development target uses a cache, shows good performance
image

andrewpearce-digital:
A local no-cache rebuild of the service with the multistage file took 31 minutes on my machine
`109.63s user 141.35s system 13% cpu 31:17.49s total`
Compared to master
`97.44s user 124.83s system 12% cpu 28:37.88s total`

andrewpearce-digital:
```
Step 3/18 : RUN composer install --prefer-dist --no-dev --no-suggest --no-interaction --no-scripts --optimize-autoloader --ignore-platform-reqs
 ---> Running in 7bb9fbcb82e3
You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.
```
Removing the --no-suggest option as per the warning

andrewpearce-digital:
Without dev dependencies, the build would return only 
image

Is this due to FILTER_VALIDATE_BOOL being recognised with dev dependencies where it should be FILTER_VALIDATE_BOOLEAN?

andrewpearce-digital:
Improved naming of jobs on the pipeline

service-front/app/config/autoload/features.global.php
```diff
@@ -4,6 +4,6 @@
 
 return [
     'feature_flags' => [
-        'use_older_lpa_journey' => filter_var(getenv('USE_OLDER_LPA_JOURNEY'), FILTER_VALIDATE_BOOL) ?: false,
+        'use_older_lpa_journey' => filter_var(getenv('USE_OLDER_LPA_JOURNEY'), FILTER_VALIDATE_BOOLEAN) ?: false,
```
cooperaj, line 7: Looks to be a php 8 thing https://github.com/php/php-src/commit/f13d0a72d5cf92785c91ffc33c27df3df3f8e96e

Not sure why it works as a dev install though... maybe a library we pull in for dev is adding a shim.

cooperaj: Yep symfony/polyfill-php80. 

I'll take a look to see if it offers anything beyond this gives us some forward compatibility.

service-front/docker/app/Dockerfile
```diff
@@ -21,5 +27,11 @@ COPY service-front/docker/app/fpm-pool.ini /usr/local/etc/php-fpm.d/zz-logging.c
 
 WORKDIR /app
 
+CMD php-fpm
+
+FROM production as development
+
+COPY --from=composer-dev /app/vendor /app/vendor
+
 CMD ([[ -z "${ENABLE_XDEBUG}" ]] || docker-php-ext-enable xdebug) \
```
cooperaj, line 36: Having this at the bottom means we can only run debug on development images, ok but intentional?

service-front/app/config/autoload/features.global.php
andrewpearce-digital: That's what I thought, I found something in a symphony package that was doing that and thought one of the laminas packages was doing the same.

cooperaj (approved):
Few comments but looks great

service-front/app/config/autoload/features.global.php
cooperaj: The symfony package will have been brought in by the behat testing libraries I imagine.

service-front/docker/app/Dockerfile
andrewpearce-digital: intentional. I'd be happy to work out having it enabled in the production images or ticketing up the changes to have production images do what's needed without debugging enabled.

service-pdf/docker/app/Dockerfile
```diff
@@ -29,8 +29,12 @@ WORKDIR /src/app
...
-RUN npm install
+RUN npm install --only=production
 
 EXPOSE 80
 
 CMD [ "node", "-r", "esm", "src/server.js"]
+
+FROM production AS development
+
+RUN npm install
```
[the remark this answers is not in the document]
andrewpearce-digital: I think I observed it just installing the difference given (missing at that stage) between what was in production and development, so left it as is.

service-front/docker/app/Dockerfile
cooperaj: To be honest it's probably not necessary nor even configured properly to work when the images are running in ECS so best this way. Making sure we're running development images locally though is something to verify.

andrewpearce-digital: Without specifying a build target during docker build, you'll get a development image, so there should be no changes required to local build steps to get a dev image.

PR #672, opened 2021-01-28
Changed healthcheck endpoint used for sirius_api and changed a key in…

SeemaMenon:
… the json return

# Purpose

https://opgtransform.atlassian.net/browse/UML-1243

Fixes UML-1243

## Approach

## Learning

## Checklist


service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -96,22 +96,24 @@ protected function checkApiEndpoint(): array
...
 
-            // when $data == null a 404 has been returned from the api
-            if (is_null($data)) {
+            if ($response->getStatusCode() === 200) {
                 $data['healthy'] = true;
             } else {
                 $data['healthy'] = false;
             }
-        } catch (Exception $e) {
+        } catch (GuzzleException $ge) {
```
cooperaj, line 112: This class shouldn't be aware that Guzzle is being used. There is a better exception to use in this case.
I think the possibility is that it either an ApiException or a ClientExceptionInterface would be thrown so maybe better to leave it as an Exception.

cooperaj (changes requested):
Quick fix and looks good

service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -96,22 +88,24 @@ protected function checkApiEndpoint(): array
...
+        $request = new Request('GET', $url);
+        $request = $this->awsSignature->sign($request);
+
         try {
-            $data = $this->lpaInterface->get("700000000000");
+            $response = $this->httpClient->send($request);
 
-            // when $data == null a 404 has been returned from the api
-            if (is_null($data)) {
+            if ($response->getStatusCode() === 200) {
```
cooperaj, line 99: Mmm. Thinking about it we may need to check to see the output of the healthcheck. I'm fairly certain the API returns 200 for pretty much everything and only an error when a genuine error is encountered on their end (i.e. it's not fully RESTful).

Going forward either way it's best to use StatusCodeInterface::STATUS_OK instead of magic numbers.

SeemaMenon: Mmm...ok, changed on both checks.

cooperaj (commented):
Looks good, just need to check the api healthcheck endpoint doesn't return a 200 response if they fail (like I think they do)

service-api/app/src/App/src/Handler/HealthcheckHandler.php
SeemaMenon: Changed.

SeemaMenon:
> Looks good, just need to check the api healthcheck endpoints don't return a 200 response if they fail (like I think they do)

Screenshot 2021-01-28 at 17 42 30

cooperaj:
I believe that this won't work. It's quite possible that the sirius_api and lpa_codes_api healthcheck endpoints will return a 200 response even if they have failed. The reason for the failure will be a part of the body of the response. I think this because that's how the rest of the endpoints work. The "not found" response for an lpa code is a reponse with 200 status code and body of "{actor: null}".

service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -76,7 +76,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
     {
         return new JsonResponse([
             "version" => $this->version,
-            "api" => $this->checkApiEndpoint(),
+            "sirius_api" => $this->checkApiEndpoint(),
```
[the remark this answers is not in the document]
SeemaMenon: @cooperaj , @GemTay  - OK with lpa_api?

cooperaj (approved):

PR #697, opened 2021-02-11
Uml 1284 frontend deps update

johnnolan:
# Purpose

We require an upgrade of our dependencies in our front end applications as they are out of date.

Fixes UML-1284

## Approach

Ran in each application a `npm-upgrade` and bumped the packages to the latest versions.

Upon upgrading the PDF Service I found multiple breaking changes that needed fixing.

 * The latest version of node swaps out `Uint8Array` for a new `ArrayBuffer` so the code handling the pdf output needed to be amended to pass and test the correct type
 * Updated the deprecated `emulateMedia` to `emulateMediaType` for puppeteer
 * Updated the Dockerfile and circleci images to use the latest versions compatible with version 7 of Puppeteer and Chrome
 * Refactored some of the code to match more closely with the `opg-pdf-service`

## Learning

Thinking when we can it would be good to pull the `opg-pdf-service` repo instead of maintaining our own code.

## Checklist


cooperaj (approved):
Looks good. I agree with moving to the opg service and have mentioned it to Andrew recently.

PR #718, opened 2021-02-22
Fix additional data on exceptions

GemTay:
# Purpose

Allows additional data to be accessed on exceptions

## Checklist

### DIFF
--- a/service-front/app/src/Common/src/Exception/ApiException.php
+++ b/service-front/app/src/Common/src/Exception/ApiException.php
@@ -47,13 +47,12 @@ class ApiException extends AbstractApiException
     public function __construct(
         string $message,
         int $code = self::DEFAULT_ERROR,
-        ResponseInterface $response = null,
-        array $additionalData = [],
+        ?ResponseInterface $response = null,
+        ?array $additionalData,
         Throwable $previous = null
     ) {
         $this->response = $response;
         $this->code = $code;
-        $this->additionalData = $additionalData;
 
         parent::__construct(self::DEFAULT_TITLE, $message, $additionalData, $previous);
     }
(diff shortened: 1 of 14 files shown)


service-front/app/src/Common/src/Exception/ApiException.php
```diff
@@ -88,6 +101,6 @@ public static function create(string $message = null, ResponseInterface $respons
             }
         }
 
-        return new self($message, $code, $response, $previous);
+        return new self($message, $code, $response, $additionalData, $previous);
```
cooperaj, line 104: Think we might need to pair on this. I'm struggling to remember why there is the distinction between a factory built ApiException and a 'new'd one but if we're going to pass this here then we don't actually need the `getAdditionalData` function in this class.

service-front/app/test/CommonTest/Service/ApiClient/ClientTest.php
```diff
@@ -292,7 +374,16 @@ public function client_throws_error_with_delete_request()
     {
         $exceptionProphecy = $this->prophesize(HttpException::class);
         $exceptionProphecy->getResponse()
-            ->willReturn($this->setupResponse('', StatusCodeInterface::STATUS_INTERNAL_SERVER_ERROR)->reveal());
+            ->willReturn(
+                $this->setupResponse(
+                    json_encode([
+                        'title' => 'An API exception has occurred',
```
cooperaj, line 380: Yeah, I think unless we're going to be asserting on any of this fixture data we need to think about if we really need it?

service-front/app/src/Common/src/Exception/ApiException.php
GemTay: Yeah happy to pair on this, i'm still trying to understand it's original setup, as getAdditionalData didnt work, so best to go through it together since this could break a lot of things

cooperaj (approved):

cooperaj (approved):

PR #719, opened 2021-02-23
Uml 1222 requesting activation key within 14 days

GemTay:
# Purpose

Displays a different page to users who request an activation key again within 14 days, to inform them of when their letter should arrive by.

Fixes UML-1222

## Approach

Setup date comparison on the code created date returned from lpa codes api. If the difference is within 2 weeks, it throws an exception and attaches the created date as additional data on the exception. I then added a new switch case to the AddOlderLpa and CheckYourAnswersHandler to match the new exception message, and render the new twig file.

## Checklist

### DIFF
--- a/service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
+++ b/service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
@@ -9,6 +9,8 @@ use App\DataAccess\Repository\LpasInterface;
 use App\Exception\ApiException;
 use App\Exception\BadRequestException;
 use App\Exception\NotFoundException;
+use App\Middleware\ProblemDetailsMiddleware;
+use DateTime;
 use Psr\Log\LoggerInterface;
 
 class OlderLpaService
@@ -156,25 +158,37 @@ class OlderLpaService
      *
      * @param string $lpaId
      * @param string $actorId
-     * @return bool
+     * @return DateTime|null
      */
-    public function hasActivationCode(string $lpaId, string $actorId): bool
+    public function hasActivationCode(string $lpaId, string $actorId): ?DateTime
     {
         $response = $this->actorCodes->checkActorHasCode($lpaId, $actorId);
 
         if (is_null($response->getData()['Created'])) {
-            return false;
+            return null;
         }
 
-        $this->logger->notice(
-            'Activation key request denied for actor {actorId} on LPA {lpaId} as they have an active activation key',
-            [
-                'actorId' => $actorId,
-                'lpaId' => $lpaId
-            ]
-        );
+        $createdDate = DateTime::createFromFormat('Y-m-d', $response->getData()['Created']);
 
-        return true;
+        if ((int) $createdDate->diff(new DateTime(), true)->format('%a') <= 14) {
+            $this->logger->notice(
+                'Activation key created within the last 14 days already exists for actor {actorId} on LPA {lpaId}',
+                [
+                    'actorId' => $actorId,
+                    'lpaId' => $lpaId
+                ]
+            );
+        } else {
+            $this->logger->notice(
+                'Activation key request denied for actor {actorId} on LPA {lpaId}' .
+                'as they have an active activation key',
+                [
+                    'actorId' => $actorId,
+                    'lpaId' => $lpaId
+                ]
+            );
+        }
+        return $createdDate;
     }
 
     /**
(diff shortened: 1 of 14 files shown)


service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
```diff
@@ -156,25 +157,37 @@ public function compareAndLookupActiveActorInLpa(array $lpa, array $userDataToMa
...
-                'actorId' => $actorId,
-                'lpaId' => $lpaId
-            ]
-        );
+        $createdDate = DateTime::createFromFormat('Y-m-d', $response->getData()['Created']);
 
-        return true;
+        if ((int) $createdDate->diff(new DateTime(), true)->format('%a') <= 14) {
+            $this->logger->notice(
+                'Activation key created within the last 14 days already exists for actor {actorId} on LPA {lpaId}',
```
cooperaj, line 174:
```suggestion
                'Activation key recently created for actor {actorId} on LPA {lpaId}',
```

service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
```diff
@@ -156,25 +157,37 @@ public function compareAndLookupActiveActorInLpa(array $lpa, array $userDataToMa
...
-            'Activation key request denied for actor {actorId} on LPA {lpaId} as they have an active activation key',
-            [
-                'actorId' => $actorId,
-                'lpaId' => $lpaId
-            ]
-        );
+        $createdDate = DateTime::createFromFormat('Y-m-d', $response->getData()['Created']);
 
-        return true;
+        if ((int) $createdDate->diff(new DateTime(), true)->format('%a') <= 14) {
```
cooperaj, line 172: Mmm. I'm wondering if we need the distinction between 'you have a key' and 'you have a recent key' at this point. What would happen if we pass the created date back with both types of response. Would it be something we could map on the front end. The result is the same effectively (no progress) but we just change what message we show.

cooperaj: Specifically we'd be able to just say "is the date within 14 days" in the Handler and show one view or another. Alternatively you could shift it right and return a different response type to the handler from the service.

service-api/app/test/AppTest/Service/Lpa/OlderLpaServiceTest.php
```diff
@@ -543,16 +545,17 @@ public function older_lpa_lookup_throws_an_exception_if_actor_has_activation_key
...
-        $this->expectExceptionCode(StatusCodeInterface::STATUS_BAD_REQUEST);
-        $this->expectExceptionMessage('LPA not eligible as an activation key already exists');
-
-        $service->checkLPAMatchAndGetActorDetails($dataToMatch);
+        try {
+            $service->checkLPAMatchAndGetActorDetails($dataToMatch);
+        } catch (BadRequestException $ex) {
+            $this->assertEquals(StatusCodeInterface::STATUS_BAD_REQUEST, $ex->getCode());
+            $this->assertEquals(['activation_key_created' => $createdDate], $ex->getAdditionalData());
+        }
```
cooperaj, line 558: May need to return from the catch and add a forced test fail at the end of the method in case the exception isn't thrown (though you may get a 'no assertions made' error

service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
GemTay: Yeah thats totally doable and I agree, probably better! I was just sticking to the format of the API giving different responses for different cases but this case it seems kinda dumb. We could pass the created date back and just do the date comparison in the switch case in the CheckYourAnswersHandler

service-front/app/features/context/UI/LpaContext.php
```diff
@@ -1490,19 +1492,27 @@ public function iProvideTheDetailsFromAValidPaperDocument()
                     )
                 );
         } else {
+            if ((int) $this->codeCreatedDate->diff(new DateTime(), true)->format('%a') <= 14) {
```
cooperaj, line 1495: My concert for this check is that `$this->codeCreatedDate` may not exist depending on step call order.  The preceding check specifically checks for a null which would always work.

service-front/app/src/Actor/src/Handler/CheckYourAnswersHandler.php
```diff
@@ -132,6 +133,19 @@ public function handlePost(ServerRequestInterface $request): ResponseInterface
                         ['user'  => $this->user]
                     ));
                 case OlderLpaApiResponse::HAS_ACTIVATION_KEY:
+                    $createdDate = DateTime::createFromFormat('Y-m-d', $result->getData()['activation_key_created']);
```
cooperaj, line 136: Can we move the body of this case to a new private method? Will help reduce the complexity of the code.

cooperaj (changes requested):
One change and it's ready to go 👍

cooperaj (dismissed):
LGTM 👍

cooperaj (approved):

PR #734, opened 2021-02-25
Updated the version puppeteer

charlesmmarshall:
# Purpose

Attempt to resolve circle ci build failures on the pdf service

## Approach

Upgrade the base docker image and align the package.json versions to match

During testing of the image upgrade I was able to create a failure by running multiple builds in quick succession, I add the `--no-cache` option to the build step and have since seen no further errors - may not be the root, but the time difference seemed minimal so the stability trade off seemed worth while

## Checklist


johnnolan (approved):

service-pdf/docker/app/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM buildkite/puppeteer:5.2.1 AS production
```
cooperaj, line 1: is there any reason we don't run an 'apt upgrade'  in here?

.circleci/config.yml
```diff
@@ -767,7 +767,7 @@ orbs:
           - run:
               name: Build
               command: |
-                docker build --target production --file service-pdf/docker/app/Dockerfile --tag pdf:latest .
```
cooperaj, line 770: None of our build tasks specify --no-cache. Is this now needed?

charlesmmarshall: Apologies, i didnt update the description on this pr since that change, will do that shortly, but normally no we wouldnt

however, with this build error it seemed to be occurring regularly on the 2nd attempt onwards, even when updating the image, but adding the no-cache flag i have not seen a failure afterwards and very little change in the build time

service-pdf/docker/app/Dockerfile
charlesmmarshall: Running a full upgrade / update in run is generally considered bad practice and sometimes may not work - its actually very common thing and called out in best practice guide - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

the other gotcha with it is you could fall unlucky and get a point version difference between steps due to a release happening at the distro level

We do upgrade specific packages when we know they are required though

cooperaj:
PDF generation  looks fine.

PR #740, opened 2021-02-26
Use local chrome installed in image

charlesmmarshall:
# Purpose

Reading more into the errors in the build stage it appears to be happening around the download of local version of chromium in the npm install with timeout / connection issues

The base image we're using comes with a stable version of chrome installed, so this change sets the env flags to skip the download (which is where the fails look to be coming from) and set it to use local version

Ran the npm tests on local and they pass without issue

Fixes UML-1309 (hopefully)

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -767,7 +767,7 @@ orbs:
           - run:
               name: Build
               command: |
-                docker build --no-cache --target production --file service-pdf/docker/app/Dockerfile --tag pdf:latest .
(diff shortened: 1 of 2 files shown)


johnnolan (dismissed):

cooperaj (approved):

cooperaj:
It's amusing that the upstream image specifically comments that they only include OS version of chrome cause it's the easiest way to install the dependencies needed for the puppeteer version.

PR #821, opened 2021-03-30
Uml 1317 new add lpa api handler

GemTay:
# Purpose

Moves the logic from the check LPA handler into the API to refactor the add an LPA journey

Fixes UML-1317

## Approach

Created a new handler `AddLpaValidationHandler` which calls a new class `AddLpa` to perform all the current logic in the frontend handler. If any validation fails inside the AddLpa class, it throws an exception with any additional data required in the twig template. I added integration and acceptance tests for the new handler, as well as unit tests for the new AddLpa and LpaAlreadyAdded classes.

I will create a separate ticket for removing the codes dynamodb config/factory/etc to keep this pr small

## Checklist


service-api/app/src/App/src/Handler/AddLpaValidationHandler.php
```diff
@@ -0,0 +1,49 @@
...
+            !isset($data['dob'])
+        ) {
+            throw new BadRequestException("'actor-code', 'uid' and 'dob' are required fields");
+        }
+
+        $response = $this->addLpa->validateAddLpaData($data, $userId);
+
+        return new JsonResponse($response, 200);
+    }
+}
```
GemTay, line 49: The second half of the journey (ActorCodeConfirmHandler) doesn't have much functionality or logic, so I have not created a new handler for it as it would probably look the same...

service-api/app/src/App/src/Service/Lpa/AddLpa.php
```diff
@@ -0,0 +1,59 @@
...
+     * @param string $userId
+     *
+     * @return array
+     */
+    public function validateAddLpaData(array $data, string $userId): array
+    {
+        if (null !== $lpaAddedData = ($this->lpaAlreadyAdded)($userId, $data['uid'])) {
+            throw new BadRequestException('LPA already added', $lpaAddedData);
+        }
+
```
GemTay, line 46: I haven't included logging for these outcomes as I planned to setup logging with new event codes in the front, when this is all hooked up. Although if its necessary to have logging in both sides, let me know and i'll add it in :)

cooperaj: I think that we should probably have it, since they're two completely separate applications. Though I would say that the event codes are only necessary in one place.

cooperaj (commented):
Super nice work. One thing about the logging...

cooperaj (approved):

PR #895, opened 2021-05-13
UML-1095-Delete-LPA

GemTay:
# Purpose

Allows users to remove LPA's from their account, and cancels any active codes they have in the process

Fixes UML-1095

## Approach

Created a confirm and a remove handler in the front, with routes that are feature flagged until the next ticket is completed. Created a RemoveLpa class which is called by the handler and handles the API responses. Created a flash message to appear on the dashboard telling the user which LPA has been removed. Added logic into the API to cancel any active codes created by the user on the LPA. Created a full set of unit and BDD tests.

Subject to change after discussion regarding how we perform the deletion in the database

## Checklist


service-api/app/src/App/src/Handler/LpasResourceHandler.php
```diff
@@ -6,35 +6,32 @@
...
  * @package App\Handler
  * @codeCoverageIgnore
  */
 class LpasResourceHandler implements RequestHandlerInterface
 {
-    /**
-     * @var LpaService
-     */
-    private $lpaService;
+    /** @var LpaService */
```
cooperaj, line 25: With the introduction of the type it's no longer necessary to use a comment to state the type (unless additional information is provided in the comment)

service-api/app/src/App/src/Service/Lpa/RemoveLpa.php
```diff
@@ -55,17 +63,28 @@ public function __invoke(string $userId, string $token): ?array
...
+                }
             }
         }
 
-        return $this->userLpaActorMapRepository->delete($token);
+        // get the LPA to display the donor name and lpa type in the flash message
+        // we don't use getByUserLpaActorToken as it returns null if actor is inactive
+        $lpaRemovedData = $this->lpaService->getByUid($userActorLpa['SiriusUid'])->getData();
+
+        $this->userLpaActorMapRepository->delete($token);
```
cooperaj, line 85: Do we verify success of this or does it throw on error?

service-api/app/src/App/src/Service/Lpa/RemoveLpa.php
```diff
@@ -7,41 +7,49 @@
...
+     * Removes actor association from all viewer codes created by a user
+     * and removes an LPA from a users account
      *
      * @param string $userId The user account ID that must correlate to the $token
-     * @param string $token UserLpaActorToken that map an LPA to a user account
-     * @return ?array A structure that contains processed LPA data and metadata
+     * @param string $token  UserLpaActorToken that map an LPA to a user account
+     *
+     * @return array A structure that contains processed LPA data and metadata
+     * @throws Exception
```
cooperaj, line 41:
```suggestion
     * @throws NotFoundException
```
Or maybe it can throw others. It's probably good to be explicit

service-front/app/src/Actor/templates/actor/lpa-blank-dashboard.html.twig
```diff
@@ -11,6 +11,9 @@
         {{ include('@partials/welsh-switch.html.twig') }}
 
         <main class="govuk-main-wrapper" id="main-content" role="main">
+
+            {{ include('@actor/partials/flash-message.html.twig', {'flash': flash, 'flash_key':'Actor\\Handler\\RemoveLpaHandler::REMOVE_LPA_FLASH_MSG'}) }}
```
cooperaj, line 15: Should the flash_key be quoted like that?

service-front/app/src/Actor/templates/actor/lpa-dashboard.html.twig
```diff
@@ -14,6 +14,8 @@
 
         {{ include('@actor/partials/flash-message.html.twig', {'flash': flash, 'flash_key':'Actor\\Handler\\CheckLpaHandler::ADD_LPA_FLASH_MSG'}) }}
 
+        {{ include('@actor/partials/flash-message.html.twig', {'flash': flash, 'flash_key':'Actor\\Handler\\RemoveLpaHandler::REMOVE_LPA_FLASH_MSG'}) }}
+
```
cooperaj, line 18: Mmm. May need to look at how these messages are working. It shouldn't be the case that you have a different include for every message you might want to show. It should just be the one, that shows all messages.

service-front/app/src/Actor/templates/actor/partials/check-code-details.html.twig
```diff
@@ -18,10 +18,15 @@
...
-                <dt class="govuk-summary-list__key">{% trans %}Created by{% endtrans %}</dt>
-                <dd class="govuk-summary-list__value">{{ code.CreatedBy }}</dd>
-            </div>
+                <div class="govuk-summary-list__row">
+                    <dt class="govuk-summary-list__key">{% trans %}Created by{% endtrans %}</dt>
+                    {% if code.CreatedBy is defined %}
+                        <dd class="govuk-summary-list__value">{{ code.CreatedBy }}</dd>
+                    {% else %}
+                        {# this should only be the case where the user who created the code deleted their lpa, so we can't figure out who the code was created by #}
+                        <dd class="govuk-summary-list__value"></dd>
```
cooperaj, line 27: Might be worth adding a TODO or otherwise link to the talked about work on keeping a Sirius UID in place so we can populate this regardless of the deletion state.

service-front/app/src/Actor/templates/actor/partials/lpa-details.html.twig
```diff
@@ -35,6 +35,10 @@
             {% else %}
                 <span class="moj-badge moj-badge--grey">{% trans %}No organisations have access{% endtrans %}</span>
             {% endif %}
+            {% if feature_enabled('delete_lpa_feature') %}
+                <br><br>
```
cooperaj, line 39: This smells like a CSS hack :P

service-front/app/test/CommonTest/Service/Lpa/AddLpaApiResponseTest.php
```diff
@@ -9,36 +9,22 @@
 
 class AddLpaApiResponseTest extends TestCase
 {
-    /**
-     * @test
-     * @covers ::__construct
-     * @covers ::validateResponseType
-     */
+    /** @test */
```
cooperaj, line 12: It's good practice to use the @covers annotation as it helps direct our coverage reports correctly. Any reason these were removed?

I appreciate we don't use it everywhere but would prefer not to remove any we have in place.

cooperaj (commented):
Looks bangin

service-front/app/src/Actor/templates/actor/partials/lpa-details.html.twig
GemTay: Haha I forgot to come back to this, it was a quick css hack because I couldn't be bother to style it yet - i'll do it properly now :D

service-front/app/test/CommonTest/Service/Lpa/AddLpaApiResponseTest.php
GemTay: When I ran the front tests with coverage, this error was appearing `Trying to @cover or @use not existing method "::__construct".` - I've now added in at the top of the test file ` * @coversDefaultClass \Common\Service\Lpa\AddLpaApiResponse` which seems to have fixed it, so i've added the covers tags back in

service-api/app/src/App/src/Service/Lpa/RemoveLpa.php
GemTay: The operation returns all of the old values for the row of data it has just deleted, but I dont verify against these values, I guess I should? I've noticed that for a lot of the existing DB operations we don't verify success or handle errors, maybe because we don't expect errors to ever occur due to all the checks we do before performing the final DB operation?

service-front/app/src/Actor/templates/actor/lpa-dashboard.html.twig
GemTay: True, i'll pass both parameters in, however the first set of array keys are then going to be overridden by the 2nd lot since the keys are the same, so i'll try to figure out a solution that doesn't over complicate the flash message twig file

GemTay:
> Looks good, though I think we may need a ticket to tidy up the flash message partial.

Created :) 
https://opgtransform.atlassian.net/secure/RapidBoard.jspa?rapidView=142&projectKey=UML&view=planning&selectedIssue=UML-1483&issueLimit=100

PR #899, opened 2021-05-14
Fix flash message css issue

GemTay:
# Purpose

Fixes styling issue of the flash messages when viewing on devices with smaller screens. The issue was discovered as part of the Delete LPA ticket

## Approach

Applied styling changes to ensure the flash message text starts on the same line as the svg icon. The changes also prevents the flash message and welsh language switch from overlapping when viewed on devices with smaller screens

## Checklist


cooperaj (approved):

GemTay:
image
image
The flash message now displays correctly on the same line on all screen sizes

SeemaMenon (approved):

PR #917, opened 2021-05-25
Back-button-translation-switch-css-fix

GemTay:
# Purpose

The recently fixed flash message styling PR has unintentionally changed how the back button and welsh switch are styled together. The back button is now being displayed on the level below the welsh switch, this PR fixes that so that they are inline again

## Approach

Added display: inline to the welsh switch css element

## Checklist


GemTay:
image
The issue

GemTay:
image

The issue is fixed with this change

SeemaMenon (approved):

PR #920, opened 2021-05-28
Uml 1364 close more info by default

lukejolliffe:
# Purpose

To ensure "What can I do with this service" section is always closed by default  

Fixes UML-1364

## Checklist


lukejolliffe:
This needs to be changed so that blank-dashboard view also closes the details tag by default

Test should also be added for this case

service-front/app/features/actor-dashboard.feature
```diff
@@ -40,15 +40,9 @@ Feature: The user is able to see correct information on their dashboard
...
     When I am on the dashboard page
-    Then I can see that the What I can do link is open
-
-  @ui
-  Scenario: As a user who has created access codes What Can I do with my LPAs is closed
-    Given I have added an LPA to my account
-    When I can see that my LPA has 2 active codes with expiry dates 2041-04-01T23:59:59+01:00 2041-04-01T23:59:59+01:00
     Then I can see that the What I can do link is closed
 
   @ui
```
SeemaMenon, line 48: @lukejolliffe  -  May be worth keeping these scenarios and check  the  link stays closed, irrespective of having LPA's [with or without access codes] in account or not.

GemTay: Since were removing a feature, i dont think we need tests to check that something isnt happening, it just seems over kill to me

SeemaMenon (approved):

service-front/app/features/actor-dashboard.feature
SeemaMenon: ok

PR #923, opened 2021-06-01
Update node versions for M1 compatibility

lukejolliffe:
building binaries do not work with sass 5.0 so updated to 6.0.

node 14 updated to 15

python, make and g++ all needed to be installed to make the sass binary

# Purpose

Currently there are no node-sass binaries for arm macs. To get around this docker will build the binaries itself but the tools used to build the binaries need to be added to the container.

## Checklist

* [ x] I have performed a self-review of my own code

### DIFF
--- a/service-front/web/Dockerfile
+++ b/service-front/web/Dockerfile
@@ -1,7 +1,9 @@
-FROM node:14-alpine
+FROM node:15-alpine
 
 RUN apk add --no-cache \
       chromium \
+      python \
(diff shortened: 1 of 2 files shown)


cooperaj:
Due to the platform differences I'm going to have to pull this to test it out. Will get it done and report back.

cooperaj:
Am closing this as the work was done as a part of the general update PR

PR #962, opened 2021-06-23
UML-1524 - Address TFSec findings

andrewpearce-digital:
# Purpose
Address TFSec findings

Fixes UML-1524

## Approach

- Update descriptions for security groups and rules
- mark exceptions for ingress and egress rules that are currently required
- use name prefixes for security groups over (explicitly) naming to allow create-before-destroy lifecycles
- enable container insights on ECS clusters
- drop invalid headers on load balancers
- define kms key for sns topics

## Learning

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group

## Checklist


terraform/environment/dynamodb.tf
```diff
@@ -2,6 +2,9 @@ resource "aws_dynamodb_table" "actor_codes_table" {
   name         = "${local.environment}-ActorCodes"
   billing_mode = "PAY_PER_REQUEST"
   hash_key     = "ActorCode"
+  # server_side_encryption {
```
johnnolan, line 5: Any reason for these being commented? Just for my education

johnnolan (approved):
Apart from the comment all looks good with my eyes

terraform/environment/dynamodb.tf
andrewpearce-digital: I need to make some keys so this will hang around until UML-1599

PR #965, opened 2021-07-01
UML-1507 : Changes that dashboard doesn't break when donor date of birth is  missing

SeemaMenon:
# Purpose

https://opgtransform.atlassian.net/browse/UML-1507

Fixes UML-1507

## Approach

## Learning

## Checklist

### DIFF
--- a/service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
+++ b/service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
@@ -67,18 +67,6 @@ class OlderLpaService
 
     public function checkDataMatch(array $actor, array $userDataToMatch): ?array
     {
-        //check if actor has a valid dob
-        if (!isset($actor['dob'])) {
-            $this->logger->notice(
-                'The actor {id} for the LPA {uId} does not have a DOB in sirius',
-                [
-                    'id' => $actor['uId'],
-                    'uId' => $userDataToMatch['reference_number'],
-                ]
-            );
-            return null;
-        }
-
         // Check if the actor has more than one address
         if (count($actor['addresses']) > 1) {
             $this->logger->notice(
@@ -224,12 +212,6 @@ class OlderLpaService
         }
 
         if (!($this->validateLpaRequirements)($lpaMatchResponse->getData())) {
-            $this->logger->info(
-                'The LPA {uId} entered by user is not eligible due to registration date',
-                [
-                    'uId' => $dataToMatch['reference_number'],
-                ]
-            );
             throw new BadRequestException('LPA not eligible due to registration date');
         }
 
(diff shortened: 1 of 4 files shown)


service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
```diff
@@ -67,6 +67,18 @@ public function cleanseUserData(array $data): ?array
 
     public function checkDataMatch(array $actor, array $userDataToMatch): ?array
     {
+        //check if actor has a valid dob
```
GemTay, line 70: if the donors DOB on the LPA is blank, it wont match what they have entered anyway, so this code isnt needed

service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
```diff
@@ -212,6 +224,12 @@ public function checkLPAMatchAndGetActorDetails(array $dataToMatch): array
         }
 
         if (!($this->validateLpaRequirements)($lpaMatchResponse->getData())) {
+            $this->logger->info(
```
GemTay, line 227: we already log this in the validateLpaRequirements class

GemTay (changes requested):
When users create a code and give it to an organisation, the LPA summary on the viewer side will either break or the donors DOB will be blank. I'm not sure how organisations will feel about this and whether it means the LPA can still be used. I think we need to go back to Sarah and see how we'd like to deal with this

service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
SeemaMenon: you are right .... removed.

service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
SeemaMenon: Yeah, get you there.

I am so tempted to not do a data match check if the relevant details on the LPA are not available.

SeemaMenon:
> When users create a code and give it to an organisation, the LPA summary on the viewer side will either break or the donors DOB will be blank. I'm not sure how organisations will feel about this and whether it means the LPA can still be used. I think we need to go back to Sarah and see how we'd like to deal with this

The viewer side  [and the actor side] would not break, but would show a blank entry for the dob.
This has been discussed [with Jude in Sarah's absence] and I think for now, business is OK with a blank entry  for DB in view summary[Than have a breaking dashboard on the actor side].Is mentioned in the ticket I think. 

What Jude mentioned to me was the new ticket 1588 would be the new feature to make users aware and be proactive.
I also think it's a data cleansing  activity which is in discussion.

GemTay (approved):

PR #985, opened 2021-07-02
Uml 1116 admin search by email

cooperaj:
# Purpose

Add email lookup to the admin panel

Fixes UML-1116

## Approach

Cribbed stuff from various MoJ repos.

## Checklist

* [ ] ~I have added welsh translation tags and updated translation files~
* [ ] ~I have run an accessibility tool on any pages I have made changes to and fixed any issues found~


GemTay (approved):
Nice job, looks really good! I've not seen the template naming convention of eg /default.partial.gohtml instead of /partials/default.gohtml before but I can see how it works! Theres a bit of test coverage missing but i'll still approve to get it merged in before the sprint ends as it can be done later anyway :)

service-admin/internal/server/data/account.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if *result.Count > 0 {
+		results := []ActorUser{}
+
+		err = dynamodbattribute.UnmarshalListOfMaps(result.Items, &results)
+		if err != nil {
+			log.Error().Err(err).Msg("unable to convert dynamo result into ActorUser")
+		}
+
+		// we'll only ever want the one result
```
lukejolliffe, line 57: Were we planning on allowing a "fuzzy search" for email? Incase someone has entered their email incorrectly? I remember something being discussed, maybe we should write a ticket?

lukejolliffe (dismissed):

service-admin/internal/server/data/account.go
cooperaj: It's not possible with our current stack choices (to my knowledge) so until we've built out a bunch of elastic search stuff it'll be a ticket we can't play.

lukejolliffe (approved):

PR #992, opened 2021-07-15
Uml 1574 split older form into separate pages

lukejolliffe:
# Purpose

Existing form to capture user details in Older Journey has been split into individual forms per input. This change is purely to aid ease of use

Fixes UML-1574
## Approach

Split forms, handlers, templates and routes. It seemed sensible to move each of these splits into their own containing folder. 

When changing the tests for this PR I noticed that some of the LPA context could be pulled out into a separate context relating to requesting an activation key. I've made this change and made sure I have correctly merged the changes made to LPA context.

With this new layout it had a knock on effect to the change links on the check answers page. Each change link has been changed to link to the correct split page. 

After discussions with the team the continue button now takes the user back to the check answers page based on the state of the session. This behaviour made the back button functionality quite unpredictable, It has now been changed so that it uses the JS history object to navigate back in the browser history. 

Using the session to navigate to the check answers page meant that when users would fail at adding an LPA and click the "Try Again" button they would have all their details entered already. I have now set up a flag to make this button clean the session of the existing form inputs.  
(If this is not acceptable I will instead suggest on try again we route the user to the check answers page) 

## Learning

I struggled trying to add the back button functionality in PHP.  In the end this ended up being added in JS with just a simple history.back() on the link.

## Checklist

* [ ] ~~I have added welsh translation tags and updated translation files~~
(Not currently needed, given there are no content changes this ticket can go live without being under a feature flag) 


cooperaj:
"It has now been changed so that it uses the JS history object to navigate back in the browser history."

I've not started looking through yet but we do have referrer capturing code in place for just this reason. It avoids using JS which may be problematic for some users.

service-front/app/features/context/UI/RequestActivationKeyContext.php
```diff
@@ -0,0 +1,533 @@
...
+ */
+class RequestActivationKeyContext implements Context
+{
+    use ActorContext;
+    use BaseUiContextTrait;
+
+    /**
+     * @Then /^I am taken back to the reference number page where I can see my answer and change it$/
+     */
+    public function iAmTakenBackToTheReferenceNumberPageWhereICanSeeMyAnswerAndChangeIt()
```
cooperaj, line 37: You should be able to use PHPStorm to reformat the code so that method names are alphabetical. For ground up context rewrites like this thats always useful - even if they tend to get worse from that point it's always good to start in a well organised fashion.

service-front/app/languages/messages.pot
```diff
@@ -3830,7 +3830,7 @@ msgid "These details must match the information we have about you on our records
 msgstr ""
 
 #: src/Actor/templates/actor/request-activation-key.html.twig:56
-msgid "This must be your current postcode and match the one we have on our records. The activation key will be posted to this address. <a href=\"/lpa/change-details\">Let us know if you have moved.</a>"
+msgid "This must be your current postcode and match the one we have on our records. The activation key will be posted to this address. <a href=\"/lpa/change-details\" class=\"govuk-link\"> Let us know if you have moved.</a>"
```
cooperaj, line 3833: It will be best to replace this hard coded url with a %link% tag and pass that in from the template (using a route helper to pull it by keyname). This allows route changes that are tied to the routes.php file and not a translation file.

service-front/app/src/Actor/src/Form/RequestActivationKey/RequestNames.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        return [
+            'first_names' => [
+                'filters'  => [
+                    ['name' => StringTrim::class],
+                ],
+                'validators' => [
+                    [
+                        'name'                   => NotEmpty::class,
+                        'break_chain_on_failure' => true,
```
cooperaj, line 58: I'm not certain if this is of any use seeing as we only have 1 rule in this chain (I think, I may be misremembering validation chains)

service-front/app/src/Actor/src/Handler/RequestActivationKey/DateOfBirthHandler.php
```diff
@@ -89,12 +86,26 @@ public function handlePost(ServerRequestInterface $request): ResponseInterface
...
+    private function routeFromAnswersInSession(): RedirectResponse
+    {
+        if ($this->hasFutureAnswersInSession()) {
+            return $this->redirectToRoute('lpa.check-answers');
+        } else {
+            return $this->redirectToRoute('lpa.postcode');
+        }
+    }
+
+    private function hasFutureAnswersInSession(): bool
```
cooperaj, line 107: Like it!

service-front/app/src/Actor/src/Handler/RequestActivationKey/PostcodeHandler.php
```diff
@@ -0,0 +1,91 @@
...
+
+        switch ($request->getMethod()) {
+            case 'POST':
+                return $this->handlePost($request);
+            default:
+                return $this->handleGet($request);
+        }
+    }
+
+    public function handleGet(ServerRequestInterface $request): ResponseInterface
```
cooperaj, line 63: It could be worth (just an idea) stopping users from accessing this route (and previous) if they don't have previous answers in the session.

service-front/app/src/Actor/src/Handler/RequestActivationKey/ReferenceNumber.php
```diff
@@ -0,0 +1,115 @@
...
+            case 'POST':
+                return $this->handlePost($request);
+            default:
+                return $this->handleGet($request);
+        }
+    }
+
+    public function handleGet(ServerRequestInterface $request): ResponseInterface
+    {
+        if ($request->getQueryParams()['clearSession']) {
```
cooperaj, line 65: I wouldn't expect this to be a user controlled thing. A handler that deals with the "start again" button should clear the session before directing here.

cooperaj: Though, to be clear this works well enough, it just wouldn't be my goto.

service-front/app/src/Actor/templates/actor/cannot-find-lpa.html.twig
```diff
@@ -33,7 +33,7 @@
 
                     {{ include('@partials/contact-details/contact-details.html.twig') }}
 
-                    <a href="{{ path('lpa.add-by-paper') }}"  draggable="false" class="govuk-button">{% trans %}Try again{% endtrans %}</a>
+                    <a href="{{ path('lpa.add-by-paper',{}, {'clearSession': 'true'})}}" draggable="false" class="govuk-button">{% trans %}Try again{% endtrans %}</a>
```
cooperaj, line 36: Maybe rename it to something like 'startAgain'.

Just to avoid using more technical language in a url.

service-front/app/src/Actor/templates/actor/check-your-answers.html.twig
```diff
@@ -12,7 +12,7 @@
         {{ include('@partials/welsh-switch.html.twig') }}
 
         <div role="navigation" aria-labelledby="back-link-navigation">
-            <a href="{{ path('lpa.add-by-paper') }}" class="govuk-back-link" id="back-link-navigation">{% trans %}Back{% endtrans %}</a>
+            <a href="#" onclick="history.back()" class="govuk-back-link" id="back-link-navigation">{% trans %}Back{% endtrans %}</a>
```
cooperaj, line 15: There is a middleware in place (and twig functions I believe) that make using the browser supplied referrer a better option.

service-front/app/src/Common/src/Handler/ContactUsPageHandler.php
```diff
@@ -21,7 +21,7 @@ class ContactUsPageHandler extends AbstractHandler
     /**
      * @var UrlValidityCheckService
```
cooperaj, line 22: No need for a docblock if the code is more explicit.

service-front/app/test/ActorTest/Form/RequestActivationKey/RequestDateOfBirthTest.php
```diff
@@ -0,0 +1,42 @@
...
+    public function getFormName(): string
+    {
+        return 'request_activation_key';
+    }
+
+    public function getFormElements(): array
+    {
+        return [
+            '__csrf'                => Csrf::class,
+            'dob'                   => Date::class
```
cooperaj, line 33:
```suggestion
            '__csrf' => Csrf::class,
            'dob'    => Date::class
```

service-front/app/test/ActorTest/Form/RequestActivationKey/RequestPostcodeTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function getFormName(): string
+    {
+        return 'request_activation_key';
+    }
+
+    public function getFormElements(): array
+    {
+        return [
+            '__csrf'                => Csrf::class,
```
cooperaj, line 33: As above

service-front/app/config/routes.php
```diff
@@ -179,15 +179,30 @@
...
             Mezzio\Authentication\AuthenticationMiddleware::class,
-            Actor\Handler\RequestActivationKeyInfoHandler::class
+            Actor\Handler\RequestActivationKey\RequestActivationKeyInfoHandler::class
         ], ['GET', 'POST'], 'lpa.add-by-paper-information');
 
-        $app->route('/lpa/add-by-paper', [
+        $app->route('/lpa/request-code/lpa-reference-number', [
             Mezzio\Authentication\AuthenticationMiddleware::class,
-            Actor\Handler\RequestActivationKeyHandler::class
+            Actor\Handler\RequestActivationKey\ReferenceNumber::class
```
SeemaMenon, line 187: Might as well name the class as ReferenceNumberHandler  to keep the naming consistent

service-front/app/src/Actor/templates/actor/check-your-answers.html.twig
lukejolliffe: The issue I came across with the referrer is that it doesn't hold a history so if back is pressed multiple times the user will just get caught in a loop between two pages. 

This means the user would not be able to navigate back from postcode -> dob -> your name -> ref number instead it would go :  postcode -> dob.

Maybe we should just make back navigate to check answers if the forms have already been filled?

service-front/app/languages/messages.pot
lukejolliffe: This is on a govuk_form_element hint. Is there a way to use the %link% tag on these?

cooperaj: Yeah, should be possible. The twig construct might be complex but it should work. I'll see if I can figure out how it might work.

service-front/app/src/Actor/templates/actor/check-your-answers.html.twig
cooperaj: Ah, we may need a more bespoke solution for this then. This needs UX

service-front/app/src/Actor/src/Form/RequestActivationKey/RequestDateOfBirth.php
```diff
@@ -0,0 +1,56 @@
...
+use Laminas\Validator\{Digits, NotEmpty, StringLength};
+use Mezzio\Csrf\CsrfGuardInterface;
+
+/**
+ * Class RequestActivationKey
+ * @package Actor\Form\RequestActivationKey
+ */
+class RequestDateOfBirth extends AbstractForm implements InputFilterProviderInterface
+{
+    public const FORM_NAME = 'request_activation_key';
```
GemTay, line 23: Perhaps rename these forms to include the piece of information that is being captured eg 'request_activation_key_dob' and same for the others :)

service-front/app/languages/messages.pot
lukejolliffe: Ah it's not to bad
https://symfony.com/doc/current/reference/twig_reference.html#trans

cooperaj (commented):
Couple of changes and it's good to go.

GemTay (approved):
Looks good

PR #994, opened 2021-07-16
UML-1604: Add missing requirements.txt for lambda function

johnnolan:
# Purpose

Adds missing requirements.txt for lambda function which causes request to error as it is no longer available as default in AWS Lambda images

Fixes UML-1604

## Approach

Add a requirements.txt file with the dependency

## Learning

Use requirements

## Checklist

### DIFF
--- a/terraform/account/lambda/docker-compose-ship-to-opg-metrics.yml
+++ b/terraform/account/lambda/docker-compose-ship-to-opg-metrics.yml
@@ -5,7 +5,7 @@ services:
     image: ship-to-opg-metrics
     build:
       context: ./ship-to-opg-metrics
-      dockerfile: Test.Dockerfile
+      dockerfile: test.Dockerfile
     ports:
       - 9000:8080
     volumes:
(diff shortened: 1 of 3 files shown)


lukejolliffe (approved):
Not essential but noticed the clsf-to-sqs folder has a Test.Dockerfile so I looked online to see which one was correct but saw people use this

Dockerfile.<context>

Which means they'll all group together in the IDE. Feel free to ignore this but may be worth renaming them both to Dockerfile.test 

Otherwise this all looks good to me?

johnnolan:
> Not essential but noticed the clsf-to-sqs folder has a Test.Dockerfile so I looked online to see which one was correct but saw people use this
> 
> Dockerfile.<context>
> 
> Which means they'll all group together in the IDE. Feel free to ignore this but may be worth renaming them both to Dockerfile.test
> 
> Otherwise this all looks good to me?

Love this idea, I will do another PR to do it, thanks for the research!

PR #998, opened 2021-07-21
Uml 1568: As a user I want to be asked to confirm the LPA data is correct before requesting a key

SeemaMenon:
# Purpose

https://opgtransform.atlassian.net/browse/UML-1568

Fixes UML-1568

## Approach

## Learning

## Checklist


service-api/app/features/context/Acceptance/LpaContext.php
```diff
@@ -2023,23 +2024,25 @@ public function aLetterIsRequestedContainingAOneTimeUseCode()
...
-                'dob'               => $this->userDob,
-                'postcode'          => $this->userPostCode,
-                'force_activation_key' => false
+                'reference_number'          => $this->lpaUid,
+                'first_names'               => $this->userFirstnames,
+                'last_name'                 => $this->userSurname,
+                'dob'                       => $this->userDob,
+                'postcode'                  => $this->userPostCode,
+                'force_activation_key'      => false,
+                'request_activation_key'    => false
```
GemTay, line 2033: Im confused as to what this new `request_activation_key` attribute is for, the purpose of this journey is to request an activation key?

service-api/app/src/App/src/Handler/LpasActionsHandler.php
```diff
@@ -73,9 +74,12 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             }
         }
 
-        //If all criteria pass, request letter with activation key
-        $this->olderLpaService->requestAccessByLetter($lpaMatchResponse['lpa-id'], $lpaMatchResponse['actor-id']);
+        if ($requestData['request_activation_key']) {
```
GemTay, line 77: Instead of this, it would be better to split this second part of the journey (user clicking to confirm the lpa is correct) into a new API handler and route. Much like how we have the `AddLpaValidationHandler` and `AddLpaConfirmationHandler` for the add an LPA journey

service-front/app/src/Actor/templates/actor/check-right-lpa.html.twig
```diff
@@ -0,0 +1,80 @@
...
+
+                    <p class="govuk-body">{% trans %}Using the information you gave us, we found these details.{% endtrans %}</p>
+
+                    <dl class="govuk-summary-list govuk-summary-list">
+                        <div class="govuk-summary-list__row">
+                            <dt class="govuk-summary-list__key">
+                                {% trans %}Your name{% endtrans %}
+                            </dt>
+                            <dd class="govuk-summary-list__value">
+                                {{ user.first_names ~ ' ' ~  user.last_name }}
```
GemTay, line 32: Mm, i feel like this is kind of cheating, we have identified the actor on the LPA, and returned that data back from the API so we can use it here to display. Just displaying what they typed in (even though it must have matched) feels wrong. Its also worth noting that we don't compare the middle name they entered, so if they entered a middlename that didnt match the LPA, then they could see the confirm page with either a different middlename to the one they entered, or no middlename as they didnt include it on the LPA in the first place. These could be some questions for Sarah

GemTay: Also, theres an existing function called `actor_name` that you can use for formatting

service-front/app/src/Actor/templates/actor/partials/check-right-lpa-button-options.html.twig
```diff
@@ -0,0 +1,19 @@
+<div class="govuk-grid-row">
```
GemTay, line 1: This doesn't need to have been separated from the other template as it is not a reusable component, it will only be used in that one template, so best to just copy it into back in

service-front/app/src/Common/src/Service/Lpa/OlderLpaApiResponse.php
```diff
@@ -70,7 +72,8 @@ private function validateResponseType(string $response): bool
             self::DOES_NOT_MATCH,
             self::NOT_ELIGIBLE,
             self::HAS_ACTIVATION_KEY,
-            self::LPA_ALREADY_ADDED
+            self::LPA_ALREADY_ADDED,
+            self::ADD_LPA_FOUND
```
GemTay, line 76: This event code is for the add an LPA journey, so it shouldn't be used here. You'll need to create a new one such as `OLDER_LPA_FOUND` ?

service-front/app/src/Common/src/Service/Lpa/AddOlderLpa.php
```diff
@@ -99,9 +101,13 @@ public function __invoke(
                     throw $apiEx;
             }
         }
+        if (!empty($lpaData)) {
+            return new OlderLpaApiResponse(OlderLpaApiResponse::ADD_LPA_FOUND, $lpaData);
```
GemTay, line 105: i'd like to see a new DTO created for the response we expect back instead of just passing back an array

GemTay (changes requested):
My apologies but I don't think this is the best approach, i've got an idea in my head of how I would implement it so i'm happy to pair with you whenever you're free and we could go through it :)

service-api/app/src/App/src/Handler/LpasActionsHandler.php
SeemaMenon: @GemTay  - Makes sense the code design for the add lpa's and add older lpa's remain identical. I shall go through the former pattern.

service-front/app/src/Actor/templates/actor/check-right-lpa.html.twig
SeemaMenon: Here I was assuming, 'your name' needs to what the user provided.... I didn't think about it too much for the same reason you mentioned, that we don't check for an LPA match based on the user name provided.

May be you are right. I Shall check this with Sarah and extract the full name if asked to. Even so how do compare the names still remains a question given that we only ask first name and last name.

service-front/app/src/Actor/templates/actor/partials/check-right-lpa-button-options.html.twig
SeemaMenon: :) OK

service-front/app/src/Common/src/Service/Lpa/OlderLpaApiResponse.php
SeemaMenon: Aah I was meant to rename this. Yes rightly spotted.

SeemaMenon:
> My apologies but I don't think this is the best approach, i've got an idea in my head of how it could be improved it so i'm happy to pair with you whenever you're free and we could go through it :)

@Gemma - Happy to speak anytime today :)

service-api/app/features/actor-add-an-older-lpa.feature
```diff
@@ -13,7 +13,9 @@ Feature: Add an older LPA
     Given I am on the add an older LPA page
     When I provide the details from a valid paper LPA document
     And I confirm that those details are correct
-    Then a letter is requested containing a one time use code
+    Then I am shown details of the found LPA
+    And I confirm that those details are correct
+    Then a letter is requested containing activation key
```
cooperaj, line 18:
```suggestion
    Then I am shown the details of an LPA
    And I confirm that those details are correct
    Then a letter is requested containing a one time use code
```

service-api/app/features/actor-add-an-older-lpa.feature
```diff
@@ -63,7 +65,7 @@ Feature: Add an older LPA
     Given I am on the add an older LPA page
     And I lost the letter received having the activation key
     When I request for a new activation key again
-    Then a letter is requested containing a one time use code
+    Then a letter is requested containing activation key
```
cooperaj, line 68: Why this change?

service-front/app/features/actor-add-an-older-lpa.feature
```diff
@@ -14,7 +14,8 @@ Feature: Add an older LPA
     Given I am on the add an older LPA page
     When I provide the details from a valid paper document
     And I confirm that those details are correct
-    Then a letter is requested containing a one time use code
+    Then I am shown details of the found LPA
```
cooperaj, line 17:
```suggestion
    Then I am shown the details of an LPA
```

cooperaj: I think this step is used a bunch so won't comment again.

service-front/app/features/actor-add-an-older-lpa.feature
```diff
@@ -61,3 +62,16 @@ Feature: Add an older LPA
...
+    Given I am on the add an older LPA page
+    When I provide the details from a valid paper document
+    And I confirm that those details are correct
+    Then I am shown details of the found LPA
+
+  @ui
+  Scenario: The user is taken back to start of activation request if the found LPA is incorrect
+    Given I am on the check LPA details page
+    When  I realise this is not the correct LPA
+    Then I am taken back to the start of requesting an activation key again
```
cooperaj, line 77:
```suggestion
    Then I am taken back to the start of the "request an activation key" process
```

service-front/app/features/context/UI/LpaContext.php
```diff
@@ -2534,7 +2542,11 @@ public function theFullLPAIsDisplayedWithTheCorrect($message)
     public function theRevokedLPADetailsAreNotDisplayed()
     {
         $this->ui->assertPageAddress('/lpa/dashboard');
-        $this->ui->assertPageNotContainsText($this->lpa->donor->firstname . '' . $this->lpa->donor->middlenames . ' ' . $this->lpa->donor->surname);
+        $this->ui->assertPageNotContainsText(
+            $this->lpa->donor->firstname . '' .
```
cooperaj, line 2546:
```suggestion
            $this->lpa->donor->firstname . ' ' .
```

service-front/app/src/Actor/templates/actor/check-right-lpa.html.twig
cooperaj: This page is the "check we've found the right LPA" page so should show details from the LPA.

SeemaMenon:
Gemma has pushed a new PR for this ticket after refactoring.

PR #1002, opened 2021-07-22
UML-1556-OOL-ask-for-actor-role-when-no-match

GemTay:
# Purpose

As a user 
If my LPA has been found but my details do not match
Then I am asked if I am a donor or attorney 
So that we can verify the user is authorised by that donor (they will know donor details)

This ticket is about directing the user through additional data capture that would allow a caseworker to properly cleanse an LPA record for the requesting user (both LPA data and personal user data).

Fixes UML-1556

## Approach

Created a new handler, form, route, and template for the actor role page, and hooked it upto the no match case in the `CheckYourAnswersHandler`. The ActorRoleHandler currently does nothing if either options are selected, and that work will be done separately

The new journey is under the `allow_older_lpas` feature flag

## Checklist


service-front/app/src/Actor/src/Form/ActorRole.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return [
+            'actor_role_radio' => [
+                'required'   => true,
+                'validators' => [
+                    [
+                        'name'                   => NotEmpty::class,
+                        'options'                => [
+                            'messages'           => [
+                                NotEmpty::IS_EMPTY => 'Select whether you are the donor or an attorney on the LPA',
```
cooperaj, line 52: Has this been added to the translation file?

GemTay: Uchenna is going to get the translations for this and everything else on the OOL journey to add them in at the end, so that all tickets aren't blocked by welsh. As its under the feature flag anyway we don't need it right away

lukejolliffe:
I'm wondering if we should add all the things that need to be translated to the pot file? It may make it easier to keep track of all the translations that are needed.

service-front/app/src/Actor/src/Form/ActorRole.php
lukejolliffe: Just need this bit in the pot file as well.

lukejolliffe (changes requested):

service-front/app/src/Actor/src/Form/ActorRole.php
GemTay: Whoops, missed that - added in now

lukejolliffe (dismissed):

lukejolliffe (approved):

service-front/app/src/Actor/src/Form/ActorRole.php
cooperaj: ^ yeah, thats what I meant :)

cooperaj (approved):

PR #1023, opened 2021-08-02
Fix forms without labels and accessibility issues

GemTay:
# Purpose

500 errors were appearing on the older lpa form pages as they did not contain labels. The issue only appeared on my machine as it is maybe running in a super strict mode?

## Approach

Set an aria_labelledby tag on the form fields which is set to the page heading id. This fixes the accessibility issues on the forms. Also made labels optional by wrapping them in an if block and checking if they are defined. 

## Checklist

### DIFF
--- a/service-front/app/src/Actor/templates/actor/request-activation-key/date-of-birth.html.twig
+++ b/service-front/app/src/Actor/templates/actor/request-activation-key/date-of-birth.html.twig
@@ -33,8 +33,7 @@
                             'label_class': 'govuk-!-font-weight-bold'
                         },
                         'aria_labelledby': 'dob_heading'
-                    })
-                    }}
+                    }) }}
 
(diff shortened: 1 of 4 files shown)


service-front/app/src/Actor/templates/actor/request-activation-key/date-of-birth.html.twig
```diff
@@ -28,7 +28,13 @@
...
                     {{ govuk_form_fieldset(form.get('dob'), {
                         'hint': 'For example, 31 03 1980'| trans,
-                        'attr': {'class': 'xyz', 'label_class': 'govuk-!-font-weight-bold'} }) }}
+                        'attr': {
+                            'class': 'xyz',
+                            'label_class': 'govuk-!-font-weight-bold'
+                        },
+                        'aria_labelledby': 'dob_heading'
+                    })
+                    }}
```
cooperaj, line 37:
```suggestion
                    }) }}
```
Just for consistency with other examples

cooperaj (approved):

PR #1024, opened 2021-08-04
UML-1610 Save LPA with TTL

lukejolliffe:
# Purpose

This PR will add a row into the UserLpaActorMap with a TTL value in a  new Column ActivateBy when a request for an older LPA is made.

Fixes UML-1610

## Approach

Added column in TF files, seeded data and to addition of the database when added via the older LPA route

In dynamo.py to get the ttl to be added programatically the json dump output now uses simplejson since the native json_dump would error on any value that wasn't a string.  

A new FeatureEnabled class and Factory needed to be made in the API layer so that we can switch functionality based on feature flags in the LpasActionHandler

## Learning

Where to add my breakpoints for Acceptance tests

## Checklist

* [ ] ~~I have added relevant logging with appropriate levels to my code~~
* [ ] ~~I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~~
* [ ] ~~I have added welsh translation tags and updated translation files~~
* [ ] ~~I have run an accessibility tool on any pages I have made changes to and fixed any issues found~~
* [ ] ~~The product team have tested these changes~~


andrewpearce-digital (commented):
Terraform and DyanmoDB script changes all look good

service-api/app/features/context/Acceptance/LpaContext.php
```diff
@@ -2019,6 +2027,22 @@ public function aLetterIsRequestedContainingAOneTimeUseCode()
...
+        $this->awsFixtures->append(
+            new Result(
+                [
+                    'Item' => $this->marshalAwsResultData(
+                        [
+                            'Id' => $this->userLpaActorToken,
+                            'UserId' => $this->base->userAccountId,
+                            'SiriusUid' => $this->lpaUid,
+                            'ActorId' => $this->actorId,
+                            'Added' => (new DateTime('2020-01-01'))->format('Y-m-d\TH:i:s.u\Z')
```
GemTay, line 2039: I dont think this could be an issue, but its always just best practice to avoid hardcoding datetime's, as we have had tests break in the past because of them. Just do a `(new DateTime('+1 week'))->format('Y-m-d\TH:i:s.u\Z')` instead

service-api/app/features/context/Acceptance/LpaContext.php
```diff
@@ -2444,6 +2468,22 @@ public function iRequestForANewActivationKeyAgain()
...
```
GemTay, line 2480: When users are requesting a new activation key again (they already have one), we should have a test for both when we're writing the record to out DB for the first time (this test step), and also when we're just updating the TTL of the record that already exists

service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
```diff
@@ -111,9 +111,9 @@ public function confirmDetails(string $code, string $uid, string $dob, string $a
             try {
                 $this->userLpaActorMapRepository->create(
                     $id,
-                    $actorId,
                     $details['lpa']['uId'],
-                    $details['actor']['details']['id']
+                    $actorId,
+                    (string)$details['actor']['details']['id']
```
GemTay, line 116: are `$actorId` and `(string)$details['actor']['details']['id']` not the same ?

service-api/app/features/context/Acceptance/LpaContext.php
lukejolliffe: There is a separate ticket for this UML-1643

service-api/app/features/context/Acceptance/LpaContext.php
lukejolliffe, line 2039:
```suggestion
                            'Added' => (new DateTime())->format('Y-m-d\TH:i:s.u\Z')
```

service-api/app/features/context/Acceptance/LpaContext.php
lukejolliffe, line 2480:
```suggestion
                            'Added' => (new DateTime())->format('Y-m-d\TH:i:s.u\Z')
```

service-api/app/features/context/Integration/LpaContext.php
```diff
@@ -80,6 +80,40 @@ public function aLetterIsRequestedContainingAOneTimeUseCode()
...
+
+        // UserLpaActorMap::getUsersLpas
+        $this->awsFixtures->append(
+            new Result(
+                [
+                    'Items' => [
+                        $this->marshalAwsResultData(
+                            [
+                                'SiriusUid' => $this->lpaUid,
+                                'Added' => (new DateTime('2020-01-01'))->format('Y-m-d\TH:i:s.u\Z'),
```
lukejolliffe, line 98:
```suggestion
                                'Added' => (new DateTime())->format('Y-m-d\TH:i:s.u\Z'),
```

service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
lukejolliffe: the actorId here is actually the userId, not sure why but I'll clean it up!

service-api/app/test/AppTest/DataAccess/DynamoDb/UserLpaActorMapTest.php
```diff
@@ -12,11 +12,13 @@
 use Prophecy\Argument;
 use DateTime;
 
+use function Symfony\Component\Translation\t;
```
cooperaj, line 15: Whats this used for?

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
```diff
@@ -0,0 +1,116 @@
...
+use App\Handler\LpasActionsHandler;
+use App\Service\Features\FeatureEnabled;
+use App\Service\Lpa\OlderLpaService;
+use PHPUnit\Framework\TestCase;
+use Prophecy\Argument;
+use Psr\Http\Message\ServerRequestInterface;
+
+class LpasActionsHandlerTest extends TestCase
+{
+    public function test_when_sirius_is_down_remove_stored_LPA_request()
```
cooperaj, line 15: As a preference and to keep things consistent we use a @test docblock rather than a prefix.
```suggestion
    /**
     * @test
     */
    public function when_sirius_is_down_remove_stored_LPA_request()
```

cooperaj: Or, the single line variant to keep it tidy
```suggestion
    /** @test */
    public function test_when_sirius_is_down_remove_stored_LPA_request()
```

service-api/app/test/AppTest/Service/Lpa/OlderLpaServiceTest.php
```diff
@@ -798,4 +808,58 @@ public function older_lpa_get_by_uid_response(): Lpa
...
+    {
+        $createCalls = 0;
+        $this->userLpaActorMap->create(
+            Argument::that(
+                function (string $id) {
+                    $this->assertRegExp('|^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$|', $id);
+                    return true;
+                }
+            ),
+            Argument::exact('test-lpaId'),
```
cooperaj, line 848: Argument::exact is equal to just specifying the scalar value i.e. 'test-lpaId'

service-api/seeding/dynamodb.py
```diff
@@ -1,7 +1,8 @@
 import os
 import boto3
-import json
+import simplejson as json
```
cooperaj, line 3: What does this do?

service-api/app/test/AppTest/Service/Lpa/OlderLpaServiceTest.php
```diff
@@ -15,11 +17,14 @@
 use App\Service\Lpa\LpaService;
 use App\Service\Lpa\OlderLpaService;
 use App\Service\Lpa\ValidateOlderLpaRequirements;
+use Common\Entity\User;
 use DateTime;
 use Exception;
 use Fig\Http\Message\StatusCodeInterface;
+use PhpParser\Node\Arg;
```
cooperaj, line 24: This doesn't look used?

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $handler = new LpasActionsHandler($olderLpaServiceProphecy->reveal(), $featureEnabledProphecy);
+
+        $requestProphecy = $this->prophesize(ServerRequestInterface::class);
+        $requestProphecy->getParsedBody()
+            ->willReturn($requestData);
+        $requestProphecy->getAttribute('actor-id')->willReturn('123');
+
+        $olderLpaServiceProphecy->checkLPAMatchAndGetActorDetails('123', $requestData)->willReturn($matchedLPAData);
+
+        $featureEnabledProphecy->method('__invoke')->willReturn(false);
```
cooperaj, line 96:
```suggestion
        $featureEnabledProphecy->__invoke()->willReturn(false);
```
Works and is used elsewhere in the codebase.

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
```diff
@@ -0,0 +1,116 @@
...
+
+        $olderLpaServiceProphecy->storeLPARequest('number', '123', '123')->shouldNotBeCalled();
+
+        $olderLpaServiceProphecy->hasActivationCode('number', '123')->willReturn(null);
```
cooperaj, line 100: This shouldn't be necessary, though as you mentioned something doesn't seem right.

service-api/app/src/App/src/DataAccess/DynamoDb/UserLpaActorMap.php
```diff
@@ -56,22 +51,29 @@ public function get(string $lpaActorToken): ?array
 
     /**
      * @inheritDoc
+     * @throws Exception
      */
-    public function create(string $lpaActorToken, string $userId, string $siriusUid, int $actorId)
+    public function create(string $uuid, string $lpaId, string $userId, string $actorId, string $expiryInterval = null)
```
cooperaj, line 56: Why rename the parameters? In PHP8 named parameters are available and this kind of change could break code quite easily. Did you change the doc referred to by @inheritDoc?

service-api/app/src/App/src/Handler/LpasActionsHandler.php
```diff
@@ -73,8 +85,25 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+                $lpaMatchResponse['actor-id']
+            );
+        }
+
+        try {
+            //If all criteria pass, request letter with activation key
+            $this->olderLpaService->requestAccessByLetter($lpaMatchResponse['lpa-id'], $lpaMatchResponse['actor-id']);
+        } catch (ApiException $apiException) {
+            if ($requestId) {
+                $this->olderLpaService->removeLpaRequest($requestId);
```
cooperaj, line 103: This fails silently if the record is not there? i.e. feature flag toggled off.

cooperaj (changes requested):
Added a few things

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
cooperaj: Oh, sorry, bad suggestion. You'll need to pass in the expected parameter to 'invoke' as well. i.e. 'save-lpa-requests' or whatever it is.

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+        if (!$exceptionRaised) {
+            self::fail('Exception should be raised when there is a bad response from sirius');
+        }
+    }
+
+    public function test_no_records_stored_or_removed_when_feature_flag_is_false()
+    {
+        $olderLpaServiceProphecy = $this->prophesize(OlderLpaService::class);
+        $featureEnabledProphecy =  $this->getMockBuilder(FeatureEnabled::class)
```
cooperaj, line 72: Wait. I didn't see this, why use Mock builder and not prophesize? That would explain the need to mock all functions. Only prophesize does that magic for you.

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+class LpasActionsHandlerTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function when_sirius_is_down_remove_stored_LPA_request()
+    {
+        $olderLpaServiceProphecy = $this->prophesize(OlderLpaService::class);
+        $featureEnabledProphecy =  $this->getMockBuilder(FeatureEnabled::class)
```
cooperaj, line 21: This should be a prophecy. We only use mocks in some very specific circumstances (ones that prophecy is not capable of mocking, such as final or abstract classes)

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
```diff
@@ -0,0 +1,119 @@
...
+        $handler = new LpasActionsHandler($olderLpaServiceProphecy->reveal(), $featureEnabledProphecy);
+
+        $requestProphecy = $this->prophesize(ServerRequestInterface::class);
+        $requestProphecy->getParsedBody()
+            ->willReturn($requestData);
+        $requestProphecy->getAttribute('actor-id')->willReturn('123');
+
+        $olderLpaServiceProphecy->checkLPAMatchAndGetActorDetails('123', $requestData)->willReturn($matchedLPAData);
+
+        $featureEnabledProphecy->method('__invoke')->willReturn(true);
```
cooperaj, line 48:
```suggestion
        $featureEnabledProphecy->__invoke('save_older_lpa_requests')->willReturn(true);
```

service-api/app/test/AppTest/Handler/LpasActionsHandlerTest.php
lukejolliffe: I didn't think I saw anything in the prophesize documentation. I'm not even sure if I saw if it didn't work I just assumed it wouldn't maybe? I'll attempt to change this then :)

service-api/seeding/dynamodb.py
lukejolliffe: simplejson can actually print Doubles that are being used by dynamoDb. Using the standard json library the print out errors

service-api/app/test/AppTest/Service/Lpa/OlderLpaServiceTest.php
```diff
@@ -68,6 +72,7 @@ public function setUp()
         $this->getAttorneyStatusProphecy = $this->prophesize(GetAttorneyStatus::class);
         $this->validateOlderLpaRequirements = $this->prophesize(ValidateOlderLpaRequirements::class);
         $this->userLpaActorMap = $this->prophesize(UserLpaActorMap::class);
+        $this->featureEnabled = $this->prophesize(FeatureEnabled::class);
```
cooperaj, line 72: Generally we do a bit of a Java thing here and suffix our prophecies as "Prophecy". It looks like two sneaked in in a previous PR but it's probably better to be consistent across the project.

service-api/app/test/AppTest/DataAccess/DynamoDb/UserLpaActorMapTest.php
lukejolliffe: Who knows :D

service-api/app/src/App/src/DataAccess/DynamoDb/UserLpaActorMap.php
lukejolliffe: Yeah this change is messy and shouldn't be kept. I did something dodgy (I think I completely rewrote this line for some reason) when adding the new parameter. When I was reading the code siriusUid is a little ambiguous in this case, A siriusUid implies it could also be an actor which it couldn't

service-api/app/src/App/src/DataAccess/DynamoDb/UserLpaActorMap.php
```diff
@@ -56,22 +51,34 @@ public function get(string $lpaActorToken): ?array
...
-        // The current DateTime, including microseconds
-        $now = (new DateTime())->format('Y-m-d\TH:i:s.u\Z');
+    public function create(
+        string $lpaActorToken,
+        string $userId,
+        string $siriusUid,
+        string $actorId,
+        string $expiryInterval = null
+    ) {
+        $added = new \DateTime();
```
cooperaj, line 63: This is fine as is but you might want to look at the DateTimeImmutable class for instances where changing after the fact is not needed. It helps nail things down for the future.

service-api/app/src/App/src/DataAccess/DynamoDb/UserLpaActorMap.php
```diff
@@ -56,22 +51,34 @@ public function get(string $lpaActorToken): ?array
...
+        $array = [
+            'Id'        => ['S' => $lpaActorToken],
+            'UserId'    => ['S' => $userId],
+            'SiriusUid' => ['S' => $siriusUid],
+            'ActorId'   => ['N' => $actorId],
+            'Added'     => ['S' => $added->format('Y-m-d\TH:i:s.u\Z') ]
+        ];
+
+        //Add ActivateBy field to array if expiry interval is present
+        if ($expiryInterval != null) {
```
cooperaj, line 73:
```suggestion
        if ($expiryInterval !== null) {
```
Always be super explicit (granted, that in the case of null there isn't much that can go wrong)

service-api/app/src/App/src/DataAccess/Repository/UserLpaActorMapInterface.php
```diff
@@ -20,15 +18,23 @@ interface UserLpaActorMapInterface
      * @param string $lpaActorToken
      * @param string $userId
      * @param string $siriusUid
-     * @param int $actorId
+     * @param string $actorId
```
cooperaj, line 21: With the reversions to the parameter name changes, and given the confusion, it could be worth adding some parameter description text. i.e.

```suggestion
     * @param string $actorId The Sirius formated UID that is associated with an actor in the system
```

service-api/app/src/App/src/Handler/LpasActionsHandler.php
```diff
@@ -20,6 +21,9 @@ class LpasActionsHandler implements RequestHandlerInterface
 {
     private OlderLpaService $olderLpaService;
 
+    /**
```
cooperaj, line 24: A docblock like this isn't necessary unless it adds additional information not already found in the method signature i.e. description text, usages etc.

service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
```diff
@@ -86,11 +86,11 @@ public function validateDetails(string $code, string $uid, string $dob): ?array
      * @param string $code
      * @param string $uid
      * @param string $dob
-     * @param string $actorId
+     * @param string $userId
      * @return array|null
```
cooperaj, line 90: This isn't even correct. The method signature says nullable string. 

As a general rule, if you're in a file making changes, updating the docblocks to be accurate is good code scouting.

service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
```diff
@@ -289,6 +301,18 @@ public function checkLPAMatchAndGetActorDetails(string $userId, array $dataToMat
         return $this->lookupActorInLpa($lpaMatch->getData(), $dataToMatch);
     }
 
+    public function removeLpaRequest(string $requestId)
```
cooperaj, line 304: This method is not actually verifying that the LPA we have stored is a request v.s. an actual LPA record. That being the case it's technically identical to an other public method in the LpaService that allows deletion of LPA records. If you want to keep this here for simplicities sake then it should probably be private (or implement the additional checks to ensure only requests are covered.

service-api/app/test/AppTest/Service/ActorCodes/ActorCodeServiceTest.php
```diff
@@ -155,7 +155,7 @@ function (string $id) {
             ),
             Argument::exact('test-user'),
             Argument::exact('test-uid'),
-            Argument::exact(1)
+            Argument::exact("1")
```
cooperaj, line 158: If this is a string then it should be single quoted to stop the interpolation overhead. That said this line could also be 
```suggestion
            '1'
```

service-api/app/test/AppTest/Service/Lpa/OlderLpaServiceTest.php
```diff
@@ -85,8 +98,39 @@ public function request_access_code_letter(): void
...
 
+        $this->featureEnabledProphecy->__invoke('save_older_lpa_requests')->willReturn(true);
+
+        $this->userLpaActorMapProphecy->create(
+            Argument::type('string'),
+            Argument::type('string'),
+            Argument::type('string'),
+            Argument::type('string'),
+            Argument::type('string')
+        )->shouldBeCalled();
```
cooperaj, line 109: We should avoid usage of inexact matching unless we're doing processing in the SUT that produces unpredictable data i.e. password hashing. Since we're in full control of the test input here (bar the guid) we should be matching on exactly that stuff.

```suggestion
        $this->userLpaActorMapProphecy->create(
            Argument::type('string'),
            $this->userId,
            $this->lpaUid,
            $this->actorId,
            'P1Y'
        )->shouldBeCalled();
```

cooperaj: Same applies to other instances of this in this test file

service-api/app/test/AppTest/Service/Lpa/OlderLpaServiceTest.php
```diff
@@ -798,4 +881,58 @@ public function older_lpa_get_by_uid_response(): Lpa
...
+            ),
+            'test-user',
+            'test-lpaId',
+            '1',
+            'P1Y'
+        )->shouldBeCalled();
+
+        $service = $this->getOlderLpaService();
+
+        $service->storeLPARequest('test-lpaId', 'test-user', '1');
```
cooperaj, line 905: Why these strings and the not class properties as used above?

cooperaj (approved):
Nice work!

PR #1027, opened 2021-08-09
UML-1606 Actor role journeys

GemTay:
# Purpose

As a user 
If my LPA has been found but my details do not match
Then I am asked if I am a donor or attorney 
So that we can verify the user is authorised by that donor (they will know donor details)

This ticket is about directing the user through additional data capture that would allow a caseworker to properly cleanse an LPA record for the requesting user (both LPA data and personal user data).

Fixes UML-1606

## Approach

Created a new handler, route, form and template for the donor details page. Hooked it upto the actor role page when the user selects the attorney option. The data gathered on the donor details page is stored in the session to be used on the confirm answers page. The donor details are unset from the session in the ActorRoleHandler Donor if statement, so that the data will no be presented on the final 'Check answers and consent if needed' page, when that ticket is played. New behat ui tests have beem written for the 2 journeys (actor is donor OR actor is attorney), and unit tests were created for the DonorDetails form

## Learning

Is there a better way of using the feature flag in the behat tests? I have had to wrap lots of steps in an if statement that checks the feature flag. I'm looking into whether there is an easier way of doing this in behat.

## Checklist


service-front/app/features/context/UI/RequestActivationKeyContext.php
```diff
@@ -449,6 +516,21 @@ public function iProvideTheDetailsFromAValidPaperLPAWhichIHaveAlreadyAddedToMyAc
             );
     }
 
+    /**
+     * @When /^I provide the donor's details$/
+     */
+    public function iProvideTheDonorSDetails()
```
lukejolliffe, line 522:
```suggestion
    public function iProvideTheDonorsDetails()
```

service-front/app/src/Actor/src/Handler/RequestActivationKey/DonorDetailsHandler.php
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * @param ServerRequestInterface $request
+     * @return ResponseInterface
+     */
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $user = $this->getUser($request);
+        $form = new DonorDetails($this->getCsrfGuard($request));
+        $session = $this->getSession($request, 'session');
+
```
lukejolliffe, line 51: Should we be populating the form here based on the session? I think the plan is to have a check answers page that can go back to change the answers given if a mistake has been made.

lukejolliffe (approved):

PR #1040, opened 2021-08-16
Uml 1578 check active activation key exists

SeemaMenon:
# Purpose

https://opgtransform.atlassian.net/browse/UML-1578

Fixes UML-1578

## Approach

## Learning

## Checklist


service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -125,6 +125,107 @@ public function getByUserLpaActorToken(string $token, string $userId): ?array
         return [];
     }
 
+    /**
```
cooperaj, line 128: There is a huge amount of code duplication here. Will need refactoring down. This also steps quite a lot on the code in uml-1611 via the `LpaAlreadyAdded` class. Would recommend holding off till that goes in as we're already doing that process in a sub-optimal way.

service-api/app/test/AppTest/Service/Lpa/LpaAlreadyAddedTest.php
```diff
@@ -40,7 +40,7 @@ private function getLpaAlreadyAddedService(): LpaAlreadyAdded
     public function test_returns_null_if_lpa_not_already_added()
     {
         $this->lpaServiceProphecy
-            ->getAllForUser('12345')
+            ->getAllAddedLpasForUser('12345')
```
cooperaj, line 43: What happens to this test if the feature flag is disabled?

service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -125,6 +125,107 @@ public function getByUserLpaActorToken(string $token, string $userId): ?array
...
+        // Return all the LPA details based on the Sirius Ids.
+        $lpas = $this->lpaRepository->lookup($lpaUids);
+
+        $result = [];
+
+        // Map the results...
+        foreach ($lpaActorMaps as $item) {
+            //Filter LPAs that have activation key requests made already.
+            //The column ActivateBy will only exist on LPAs that have not been activated
+            if ($keyCheck) {
```
cooperaj, line 203: If keycheck is false this method does nothing?

service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -125,6 +125,107 @@ public function getByUserLpaActorToken(string $token, string $userId): ?array
...
+        // Return all the LPA details based on the Sirius Ids.
+        $lpas = $this->lpaRepository->lookup($lpaUids);
+
+        $result = [];
+
+        // Map the results...
+        foreach ($lpaActorMaps as $item) {
+            // Filter already added LPA.
+            //The column ActivateBy will not exist on LPAs that have been activated or added
+            if (is_null($item['ActivateBy'])) {
```
cooperaj, line 153: We load *all* lpas in a users account from the Sirius API and then ignore some of them?

service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -125,6 +125,107 @@ public function getByUserLpaActorToken(string $token, string $userId): ?array
...
+            }
+        }
+        return $result;
+    }
+
+    /**
+     * Return all activation key request made LPAs for the given user_id
+     *
+     * @param string $userId User account ID to fetch LPAs for
+     * @param bool $keyCheck flag set to true if the request is for active Activation Key check
```
cooperaj, line 181: Altering a methods behaviour based on a boolean parameter is generally considered bad practice. Given it doesn't appear to do much (other than make this method return nothing if it is false) it's probably not necessary.

service-api/app/src/App/src/Service/Lpa/LpaAlreadyAdded.php
```diff
@@ -5,25 +5,31 @@
 namespace App\Service\Lpa;
 
 use Psr\Log\LoggerInterface;
+use App\Service\Features\FeatureEnabled;
```
cooperaj, line 8: This part of the project has been done in #1041. Lets get that in first and rework around it.

cooperaj (changes requested):
Some comments to look at

cooperaj:
I've worked in a large amount of refactoring of the OlderLpaService class as it was just too big. It still retains the duplicated code and odd doings that needs fixing but the refactoring should make it clearer as to where and how that should work. 

Adding feature flag checks to handlers is not the appropriate way to do what we need in this case and we should instead do those kind of things in whichever service classes are correct.

Whats become clear to me as I've done the refactoring work is that this ticket is relatively easy to actually implement. We already have an _LpaAlreadyAdded_ service class that does the work - though it currently checks and alter behaviour based on _ActivateBy_ column checks theres no reason that can be further feature flagged to simply return a **Not Activated** flag in it's data that the calling service can act appropriately on. 

The PR is currently failing a single acceptance test because the code to extend what a valid date for an LPA registration date is has not yet been added. Also a lot of integration tests but that because they all point to service methods that no longer exist.

PR #1041, opened 2021-08-16
Uml 1611 - TTL field existence

cooperaj:
# Purpose

We need to change our check for the existence of the Lpa within a users account so that if the _ActivateBy_ column exists then the check passes (as if the user does not have the Lpa in their account.

Fixes UML-1611

## Approach

Add feature flag usage and refactor code to check the db *without* pulling every LPA the user has from Sirius

## Checklist

* [ ] ~I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~
* [ ] ~I have added welsh translation tags and updated translation files~
* [ ] ~I have run an accessibility tool on any pages I have made changes to and fixed any issues found~
* [ ] ~The product team have tested these changes~


GemTay (approved):
Nice and clean!

PR #1045, opened 2021-08-27
Uml 1697 filter not activated lpa from dashboard

lukejolliffe:
# Purpose

Change implementation of getUserLpas so only activated LPAs are recieved. This will then stop LPAs from being visible on the dashboard if they are not activated

Fixes UML-1697

## Approach

Changing this getUserLPAs method means that we are changing functionality as little as possible. Everywhere that gets LPAs are most likely using this method and will now mirror the existing functionality. 

In the future we should potentially add another method to get the non activated LPA requests to display on the dashboard. 

## Checklist

* ~~[ ] I have added relevant logging with appropriate levels to my code~~
* ~~[ ] I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~~
* ~~[ ] I have added welsh translation tags and updated translation files~~
* ~~[ ] I have run an accessibility tool on any pages I have made changes to and fixed any issues found~~
* ~~[ ] The product team have tested these changes~~

### DIFF
--- a/service-api/app/src/App/src/Service/Lpa/LpaAlreadyAdded.php
+++ b/service-api/app/src/App/src/Service/Lpa/LpaAlreadyAdded.php
@@ -47,25 +47,27 @@ class LpaAlreadyAdded
     {
         $lpasAdded = $this->lpaService->getAllActivatedLpasForUser($userId);
 
-        foreach ($lpasAdded as $userLpaActorToken => $lpaData) {
-            if ($lpaData['lpa']['uId'] === $lpaUid) {
-                $this->logger->info(
-                    'Account with Id {id} has attempted to add LPA {uId} which already exists in their account',
-                    [
-                        'id' => $userId,
-                        'uId' => $lpaUid
-                    ]
-                );
-                return [
-                    'donor' => [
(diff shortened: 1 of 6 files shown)


service-api/app/src/App/src/Service/Lpa/LpaAlreadyAdded.php
```diff
@@ -45,80 +45,31 @@ public function __construct(
...
-
-            foreach($savedLpaRecords as $record) {
-                if ($record['SiriusUid'] === $lpaUid && !array_key_exists('ActivateBy', $record)) {
-                    $lpa = $this->lpaService->getByUserLpaActorToken($record['Id'], $userId);
-
-                    // a count of a null array will be 0
-                    if (count($lpa) === 0) {
-                        return null;
-                    }
+        $lpasAdded = $this->lpaService->getAllActivatedLpasForUser($userId);
```
GemTay, line 48: The original solution is better suited for this LpaAlreadyAdded class, otherwise we're fetching ALL of a users LPA's from Sirius, to then just discard most, if not all of them if there isnt a match.

GemTay (changes requested):
Not sure this works everywhere

PR #1046, opened 2021-08-31
Uml 1613 strip ttl

lukejolliffe:
# Purpose

When going through the add LPA with code journey, If the LPA already exists with a TTL in the users account. The TTL should be removed. 

Fixes UML-1613

## Approach

Add check for LPA in account within ActorCodeService. Check will determine if ttl is removed or entry created in database.

Assumptions :
* LPAs already added without an expiry will not be able to get past this point because they would have failed validation
* Only 1 Lpa with TTL of the same ID will be added to the users account at a time. Code at the moment would just pick the first LPA to remove TTL from not multiple

## Learning

## Checklist


service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
```diff
@@ -46,6 +46,50 @@ public function __construct(
...
+    /**
+     * Removes TTL from entry that is already inside the database
+     * @param array $lpaIds the LPA sirius IDs paired to their UID in the database
+     * @param string $lpaId the LPA id to remove the TTL from
+     *
+     * @return string|null returns the database ID of the LPA that has had it's TTL removed
+     */
+    private function removeTTLFromRequest(array $lpaIds, string $lpaId): ?string
+    {
+        $this->userLpaActorMapRepository->removeActivateBy($lpaIds[$lpaId]);
```
GemTay, line 88: why are you passing an array of id's and also a single id? You just need pass in the lpaActorToken of the record you want the ttl stripped from

service-api/app/src/App/src/Service/ActorCodes/ActorCodeService.php
```diff
@@ -132,4 +163,16 @@ public function confirmDetails(string $code, string $uid, string $dob, string $u
...
         return $id;
     }
+
+    /**
+     * Gets all users LPAs Ids in account
+     * @param string $userId  Identification number of the user
+     *
+     * @return array an array with LPAUid as keys to database UUID values
+     */
+    private function lpasInAccount(string $userId): array
```
GemTay, line 173: as this function is only being used here in this one service, there isnt much value in separating it out into its own function

SeemaMenon (approved):

SeemaMenon (approved):

PR #1064, opened 2021-09-01
Filter not activated LPAs 2.0

lukejolliffe:
# Purpose

Change implementation of getUserLpas so only activated LPAs are recieved. This will then stop LPAs from being visible on the dashboard if they are not activated

Fixes UML-1697

## Approach

Changing this getUserLPAs method means that we are changing functionality as little as possible. Everywhere that gets LPAs are most likely using this method and will now mirror the existing functionality. 

## Checklist

* ~~[ ] I have added relevant logging with appropriate levels to my code~~
* ~~[ ] I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~~
* ~~[ ] I have added welsh translation tags and updated translation files~~
* ~~[ ] I have run an accessibility tool on any pages I have made changes to and fixed any issues found~~
* ~~[ ] The product team have tested these changes~~


cooperaj (commented):
Code looks fine, I have a concern with implementation though in that altering a methods return based on boolean parameters is generally seen as bad practice as it can be avoided by restructuring the code itself.

The getAllForUser method is called in 2 locations so, although I appreciate the simplicity of this change, wanting to avoid wider scale changes is not a particularly relevant concern.

Happy to discuss.

lukejolliffe:
> Code looks fine, I have a concern with implementation though in that altering a methods return based on boolean parameters is generally seen as bad practice as it can be avoided by restructuring the code itself.
> 
> The getAllForUser method is called in 2 locations so, although I appreciate the simplicity of this change, wanting to avoid wider scale changes is not a particularly relevant concern.
> 
> Happy to discuss.

That's reasonable, I'll change it into two different methods

lukejolliffe:
This PR is helped by https://github.com/ministryofjustice/opg-data-lpa/pull/51 which will add a requested LPA to the users LPAs which will then either be shown on the dashboard or not depending on the state of this ticket

SeemaMenon (approved):

PR #1074, opened 2021-09-07
UML-1711 Before scenario feature toggle

GemTay:
# Purpose

Enables scenarios to be tagged with the required feature flag name and value. This way we can test all user journeys

Fixes UML-1711

## Approach

Created a FeatureFlagContext.php file for each context (UI, Front Integration, API Integration & Acceptance). The FeatureFlagContext contains a before step which validates and sets the feature flag and value defined in the tag eg `@ff:allow_older_lpas:false`. I tagged the appropriate scenarios and tweaked some test steps.

## Learning

Don't set properties in the `prepareContext` integration function such as:
`$this->lpaService = $this->container->get(LpaService::class);` , as the container values within the defined class will not contain any changes made to the container such as in the before step to a feature flag value.

With the factory method, once the factory has been instantiated, whenever we call the FeatureEnabled class to check a value, it will always refer to that same factory instance to check against. This is a problem, as when we change values defined in the container in our before step, we now need a way to force the factory to start using a new definition. This piece of code achieves that:
`$this->base->container->set(
    FeatureEnabled::class,
    new FactoryDefinitionHelper($this->base->container->get(FeatureEnabledFactory::class))
);`

There is an issue on the Front UI & API Acceptance tests where by the feature flag @BeforeStep is being called before the Base method's @BeforeStep `gatherContexts()`, and therefore all of the trait properties within the FeatureFlagContext.php are not set such as $container. A solution for the time being has been implemented to call `gatherContexts($scope)` inside the FeatureFlagContext.php @BeforeStep. This is fine for now until we figure out a way to define the order in which before steps are called if there is more than one

## Checklist


service-front/app/features/context/Integration/FeatureContext.php
```diff
@@ -0,0 +1,78 @@
...
+        $lpaContextContainer->set('config', $config);
+        $lpaContextContainer->set(
+            FeatureEnabled::class,
+            new FactoryDefinitionHelper($lpaContextContainer->get(FeatureEnabledFactory::class))
+        );
+    }
+
+    protected function prepareContext(): void
+    {
+        // TODO: Implement prepareContext() method.
```
lukejolliffe: Would be best practice to just add a single line comment to these empty methods explaining why they are not used

service-api/app/features/context/Integration/FeatureContext.php
```diff
@@ -0,0 +1,49 @@
...
+                    FeatureEnabled::class,
+                    new FactoryDefinitionHelper($this->container->get(FeatureEnabledFactory::class))
+                );
+            }
+        }
+    }
+
+    protected function prepareContext(): void
+    {
+
```
lukejolliffe: Would be best practice to just add a single line comment to these empty methods explaining why they are not used

service-front/app/features/context/Integration/FeatureContext.php
GemTay: Yeah thats true, i've just updated them to say "Not needed for this context"

lukejolliffe (approved):

lukejolliffe (approved):

PR #1082, opened 2021-09-09
UML-1578 LPA already requested

GemTay:
# Purpose

As a user I want to be told when I am trying to add an LPA that I have already requested a key for.

Fixes UML-1578

## Approach

LpaAlreadyAdded class now returns a `notActivated = true` in the data which results in a bad request being thrown and caught in the front to handle this scenario. Fully covered by tests

This PR is based off 1711 Before Scenario

## Checklist


lukejolliffe (approved):

lukejolliffe (changes requested):
The AlreadyAddedCheck will need to be changed in ValidateAddLpaData otherwise I think it will stop users from adding LPAs by code?

cooperaj (dismissed):

lukejolliffe (approved):

PR #1097, opened 2021-09-22
Uml 1643 update existing activation key (refactor pt1)

cooperaj:
# Purpose

As a part of the work on UML-1643 it's been necessary to refactor some of the logic around saving records into the UserActorLpaMap table. This simplifies the code that then needs to change records in this table.

Additionally some fixes have gone in to ensure that development mode files do not affect behat test runs.

## Approach

Move the loop of code  that retried UUID's down into the datamapper layer and remove that from the 2 calling services.

## Checklist

* ~I have added relevant logging with appropriate levels to my code~
* ~I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~
* ~I have added welsh translation tags and updated translation files~
* ~I have run an accessibility tool on any pages I have made changes to and fixed any issues found~
* ~The product team have tested these changes~

### DIFF
--- a/service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
+++ b/service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
@@ -5,15 +5,12 @@ declare(strict_types=1);
 namespace App\Service\Lpa;
 
 use App\DataAccess\ApiGateway\ActorCodes;
-use App\DataAccess\Repository\KeyCollisionException;
 use App\DataAccess\Repository\LpasInterface;
 use App\DataAccess\Repository\UserLpaActorMapInterface;
 use App\Exception\ApiException;
 use App\Service\Features\FeatureEnabled;
 use DateTime;
-use Exception;
 use Psr\Log\LoggerInterface;
-use Ramsey\Uuid\Uuid;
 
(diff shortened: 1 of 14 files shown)


service-api/app/src/App/src/Service/Lpa/OlderLpaService.php
```diff
@@ -88,13 +85,9 @@ private function removeLpa(string $requestId)
...
     {
-        $requestId = null;
+        $recordId = null;
         if (($this->featureEnabled)('save_older_lpa_requests')) {
-            $requestId = $this->storeLPARequest(
-                $uid,
-                $userId,
-                $actorUid
-            );
+            $recordId = $this->userLpaActorMap->create($userId, $uid, $actorUid, 'P1Y');
```
GemTay, line 90: This doesn't update the TTL expiry of an existing record if forced, is that going to be done as a part 2 of this ticket?

cooperaj: yep. Or rather, some work to allow that will be pt 2

I'm trying to keep these small... er

GemTay (approved):
LGTM :)

PR #1146, opened 2021-10-06
Uml 1643 force activation key (pt4)

cooperaj:
# Purpose

Final PR to implement the refreshing of the ActivateBy column if the LPA record already exists within the database.

Fixes UML-1643

## Approach

If the LPA record is found in the database whilst doing validation checks, and that record has an ActivateBy column, then instead of returning that we already have it we resolve the supplied actor data as normal and attach the record ID to the response. This allows the calling confirmation code to attach that to the request letter call and therefore update the ActivateBy column with a new +1 year date. 

## Checklist

### DIFF
--- a/service-api/app/src/App/src/Service/Lpa/AddOlderLpa.php
+++ b/service-api/app/src/App/src/Service/Lpa/AddOlderLpa.php
@@ -60,7 +60,7 @@ class AddOlderLpa
     {
         // Check if it's been added to the users account already
         $lpaAddedData = ($this->lpaAlreadyAdded)($userId, (string) $matchData['reference_number']);
-        if (null !== $lpaAddedData) {
+        if ($lpaAddedData !== null) {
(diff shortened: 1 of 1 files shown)


service-api/app/src/App/src/Service/Lpa/AddOlderLpa.php
```diff
@@ -59,7 +59,8 @@ public function __construct(
     public function validateRequest(string $userId, array $matchData): array
     {
         // Check if it's been added to the users account already
-        if (null !== $lpaAddedData = ($this->lpaAlreadyAdded)($userId, (string) $matchData['reference_number'])) {
+        $lpaAddedData = ($this->lpaAlreadyAdded)($userId, (string) $matchData['reference_number']);
+        if (null !== $lpaAddedData) {
```
lukejolliffe, line 63:
```suggestion
        if ($lpaAddedData !== null) {
```

SeemaMenon (approved):

PR #1251, opened 2021-11-19
update heading level to a variable on contact templates

lukejolliffe:
# Purpose

WCAG standards dictate that heading levels should never be skipped. Using this contact template could cause this to happen because the headings were dictated by the template and not based on the page they are put into.

Fixes UML-1764

## Approach

Parameter of heading_level is given to the template so that the template can be forced to use a different heading level if necessary

## Checklist


service-front/app/src/Actor/templates/actor/check-lpa.html.twig
```diff
@@ -80,7 +80,7 @@
                             <p>
                                 {% trans %}If the information is wrong, or if a name is misspelt, please call us:{% endtrans %}
                             </p>
-                            {{ include('@partials/contact-details/telephone.html.twig') }}
+                            {{ include('@partials/contact-details/telephone.html.twig', {heading_level:2}) }}
```
cooperaj, line 83:
```suggestion
                            {{ include('@partials/contact-details/telephone.html.twig', { heading_level: 2 }) }}
```

For a bit more readablity can we put some array like spacing in there?

cooperaj (approved):

PR #1252, opened 2021-11-19
UML-1875 - Notify team in Slack when there is a path to live failure

andrewpearce-digital:
# Purpose

Let the team know in Slack when the path to live fails

Fixes UML-1875

## Approach

- add post-step to each job on path to live
- call slack/status with fail_only condition to notify team channel of path to live failures

## Learning

- https://circleci.com/docs/2.0/configuration-reference/#pre-steps-and-post-steps-requires-version-21
- https://circleci.com/developer/orbs/orb/circleci/slack
- https://support.circleci.com/hc/en-us/articles/360047082992-Send-slack-notification-at-end-of-workflow

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -279,6 +279,7 @@ workflows:
               - slack/status:
                 fail_only: false
                 failure_message: 'TESTING - Path to Live Failure'
+                success_message: 'TESTING - Path to Live Failure'
                 webhook: '${PROD_RELEASE_SLACK_WEBHOOK}'
         - use-my-lpa/lint_terraform:
             name: lint_terraform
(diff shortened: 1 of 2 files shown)


andrewpearce-digital:
A message sent with the team channel's webhook
image

charlesmmarshall (approved):

johnnolan (approved):

johnnolan:
Just wondering, as they are all the same (from what I can tell) can they be put into any sort of little function/reference to reduce duplication? Is that even a thing? Otherwise looks good!

andrewpearce-digital:
> 

I have in the past, but I got syntax errors this time. You can use &reference and *reference.

PR #1272, opened 2021-12-03
Uml-1749 : Restrict the numbers of people able to submit their LPA to the cleansing team

SeemaMenon:
# Purpose
https://opgtransform.atlassian.net/browse/UML-1749

Fixes UML-1749

## Approach

- Introduced new flag 'streamline_cleansing_lpas'
- Check added in Api service layer to filter lpa's that are registered after 2019 and where user details do not match when flag turned on
- Tests added

## Learning

## Checklist

* [X] I have added relevant logging with appropriate levels to my code


service-api/app/features/actor-add-an-older-lpa.feature
```diff
@@ -202,3 +202,17 @@ Feature: Add an older LPA
     Then I am told my activation key is being sent
     And I should expect it within 2 weeks time
     And I will receive an email confirming this information
+
+  @acceptance @integration @pact @ff:allow_older_lpas:true @ff:streamline_cleansing_lpas:true
+  Scenario: The user cannot add an older LPA to their account when request for cleansing streamlining flag tuned on
+    Given I am on the add an older LPA page
+    And I provide details of LPA registered after 1st September 2019 which do not match a valid paper document
+    When I confirm details of the found LPA are correct and flag is turned ON
```
cooperaj, line 210: This can be written as 

`When I confirm the details of the found LPA are correct and flag is turned :flag`

The test method would then look like 

```
    /**
     * @When I confirm the details of the found LPA are correct and flag is turned :flag
     */
    public function iConfirmDetailsOfTheFoundLPAAreCorrectAndFlagIsTurned($flag)
```

Notice the regex language isn't used. Alternatively you should probably put quotes around the flag value to better indicate to the reader of the feature file that it is a changable value.

service-api/app/features/context/Integration/LpaContext.php
```diff
@@ -1641,6 +1642,7 @@ public function iProvideDetailsOfAnLPAThatDoesNotExist()
 
     /**
      * @When /^I provide details "([^"]*)" "([^"]*)" "([^"]*)" "([^"]*)" that do not match the paper document$/
+     * @When /^I provide details of LPA registered on 1st September 2019 which do not match a valid paper document$/
```
cooperaj, line 1645: I'm not sure how this works given none of the functions parameters will be set correctly?

service-api/app/src/App/src/Service/Lpa/AddOlderLpa.php
```diff
@@ -92,14 +99,23 @@ public function validateRequest(string $userId, array $matchData): array
 
         // Find actor in LPA
         $resolvedActor = ($this->findActorInLpa)($lpaData, $matchData);
+
+        // Streamline lpa's sent to cleansing team that are registered after 2019 and when user details do not match
```
cooperaj, line 103:
```suggestion
        // We may want to turn off the ability for a user to have their case pushed to the cleansing 
        // team if they fail to match and have a "newer" older lpa. In which case they'll be told we 
        // can't find their LPA.
```

Just a bit more verbose so that we'll know what this is doing in the future.

service-api/app/src/App/src/Service/Lpa/AddOlderLpa.php
```diff
@@ -34,14 +37,18 @@ public function __construct(
         LpaAlreadyAdded $lpaAlreadyAdded,
         OlderLpaService $olderLpaService,
         ValidateOlderLpaRequirements $validateOlderLpaRequirements,
-        LoggerInterface $logger
+        LoggerInterface $logger,
+        FeatureEnabled $featureEnabled,
+        RestrictSendingLpaForCleansing $restrictSendingLpaForCleansing
```
cooperaj, line 42:
```suggestion
        RestrictSendingLpaForCleansing $restrictSendingLpaForCleansing,
        LoggerInterface $logger,
        FeatureEnabled $featureEnabled
```

Seems like a silly thing but I've been trying to keep 'utility' classes at the end of the constructor and 'doing things' classes at the start.

service-api/app/test/AppTest/Service/Lpa/RestrictSendingLpaForCleansingTest.php
```diff
@@ -0,0 +1,52 @@
...
+        return new RestrictSendingLpaForCleansing(
+            $this->loggerProphecy->reveal(),
+        );
+    }
+
+    /**
+     * @test
+     * @throws Exception
+     */
+    public function throws_not_found_exception_when_lpa_status_registered_and_actorMatch_is_null()
```
cooperaj, line 39:
```suggestion
    public function throws_not_found_exception_when_lpa_status_registered_and_actorMatch_is_null(): void
```

service-api/app/test/AppTest/Service/Lpa/RestrictSendingLpaForCleansingTest.php
```diff
@@ -0,0 +1,52 @@
...
+    public function restrictSendingLpaForCleansing(): RestrictSendingLpaForCleansing
+    {
+        return new RestrictSendingLpaForCleansing(
+            $this->loggerProphecy->reveal(),
+        );
+    }
+
+    /**
+     * @test
+     * @throws Exception
```
cooperaj, line 37:
```suggestion
```
Since you're using the expectsException method this class will not throw in the classical sense. So this isn't needed.

service-front/app/features/bootstrap/behat.config.php
```diff
@@ -29,10 +29,11 @@
...
     'feature_flags' => [
-        'use_older_lpa_journey' => true,
-        'delete_lpa_feature'    => true,
-        'allow_older_lpas'      => true,
-        'save_older_lpa_requests' => true,
+        'use_older_lpa_journey'             => true,
+        'delete_lpa_feature'                => true,
+        'allow_older_lpas'                  => true,
+        'save_older_lpa_requests'           => true,
+        'streamline_cleansing_lpas'         => true,
```
cooperaj, line 36:
```suggestion
        'use_older_lpa_journey'     => true,
        'delete_lpa_feature'        => true,
        'allow_older_lpas'          => true,
        'save_older_lpa_requests'   => true,
        'streamline_cleansing_lpas' => true,
```
Should only add as many extra spaces as are needed for the longest line to make sense.

service-api/app/features/context/Integration/LpaContext.php
SeemaMenon: This step was not needed here...removed now.

docker-compose.yml
```diff
@@ -123,6 +123,7 @@ services:
       DELETE_LPA_FEATURE: "true"
       ALLOW_OLDER_LPAS: "true"
       ALLOW_MERIS_LPAS: "false"
+      DONT_SEND_LPAS_REGISTERED_BEFORE_SEP_2019_TO_CLEANSING_TEAM_ENABLED: "true"
```
cooperaj, line 126: mmm, this seems rather long. Could probably drop "_ENABLED" from the end as that's superfluous but even then.

cooperaj: it's also incorrect according to the logic? The logic says restrict sending for lpa non-matches *after* 2019

SeemaMenon: Its long but clear is what we thought, I get confused with these flag names too....and me and @andrewpearce-digital reached a consensus then to this long name.

I shall change it to 'DONT_SEND_LPAS_REGISTERED_AFTER_SEP_2019_TO_CLEANSING_TEAM'

cooperaj (approved):

PR #1337, opened 2022-01-14
UML-1931 : Allow Meris numbers to be entered as reference numbers

SeemaMenon:
# Purpose
https://opgtransform.atlassian.net/browse/UML-1931

Fixes UML-1931

## Approach

Added a checksum validator behind the flag 'allow_meris_lpas'
The validator does some basic checks and also implements luhn algorithm

## Learning

## Checklist


service-front/app/src/Common/src/Validator/ChecksumValidator.php
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return bool
+     */
+    public function isValid($reference_number)
+    {
+        $isValid = true;
+        $ref_number_checksum = '';
+
+        //check flag 'allow_meris_lpas' status here
+        if (!(strlen($reference_number) == 7 or strlen($reference_number) == 12)) {
```
cooperaj, line 37: `===` where possible please.

service-front/app/src/Common/src/Validator/ChecksumValidator.php
```diff
@@ -0,0 +1,63 @@
...
+            $this->error(self::MUST_BE_LENGTH);
+            $isValid = false;
+        } elseif (strlen($reference_number) == 12 and !preg_match('/^7/', (string)$reference_number)) {
```
cooperaj, line 40: Alternatively use the PHP string functions (https://www.php.net/manual/en/function.stripos.php) instead of regex as it would be slightly faster and maybe more readable?

service-front/app/src/Actor/src/Form/RequestActivationKey/RequestReferenceNumber.php
```diff
@@ -42,42 +47,53 @@ public function __construct(CsrfGuardInterface $csrfGuard)
...
+                    StringLength::TOO_LONG => 'The LPA reference number you entered is too long',
+                    StringLength::TOO_SHORT => 'The LPA reference number you entered is too short'
+                ],
+            ],
+        ];
+        $checkSum = [
+            'name'    => ChecksumValidator::class
+        ];
+
+        if (($this->featureEnabled)('allow_meris_lpas')) {
```
cooperaj, line 84: Maybe something that can be done is to move this feature check to the validator? This means we would be able to use the checksum on the "turned off" page?

service-front/app/src/Common/src/Validator/ChecksumValidator.php
SeemaMenon: My bad...Sorry

service-front/app/src/Actor/src/Form/RequestActivationKey/RequestReferenceNumber.php
SeemaMenon: We could do that and have it as a separate validator and use it irrespective of the flag value, but an appropriate error message with translation is not ready yet to be implemented.  The current invalid error message shown is a made up one and Sarah is OK with it because it is behind the flag atm.
[This ticket 1931 mentioned having the check behind the flag]

I have checked this suggestion with Sarah and requested for an appropriate error message to be provided along with others [which will be another ticket she mentioned] So may be separate out the checksum validator as part of that ticket?

service-front/app/src/Common/src/Validator/ChecksumValidator.php
SeemaMenon: ok...could do.

service-front/app/src/Actor/src/Form/RequestActivationKey/RequestReferenceNumber.php
SeemaMenon: Ignore above @cooperaj , I will need to remove the luhn check for Meris numbers and shall move the feature check to the validator which makes sense now.

SeemaMenon:
This PR is replaced with a new one. This ticket basically got split and luhn check for Sirius reference numbers will be a new ticket.

PR #1338, opened 2022-01-14
UML-2015 Ensure we check both ID types when showing CreatedBy

cooperaj:
# Purpose

The created by field is broken for LPAs added to accounts after the end of December. This adds additional checking logic so that we check the right place.

Fixes UML-2015

## Approach

Check both the Uid and Id for a case actor when doing a comparison

## Checklist

### DIFF
--- a/service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
+++ b/service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
@@ -103,7 +103,7 @@ class CheckAccessCodesHandler extends AbstractHandler implements UserAware, Csrf
                 $shareCodes[$key]['form'] = $form;
             }
 
-            $this->logger->info(
+            $this->logger->debug(
                 'Resolved actor id to {type}:{actor_id}',
                 [
                     'actor_id' => $code['ActorId'],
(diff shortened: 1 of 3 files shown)


service-front/app/src/Actor/src/Handler/CheckAccessCodesHandler.php
```diff
@@ -100,13 +93,19 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                 $shareCodes[$key]['form'] = $form;
             }
 
-            if ($lpaData->lpa->getDonor()->getId() == $code['ActorId']) {
+            if (
+                $lpaData->lpa->getDonor()->getId() === intval($code['ActorId'])
+                || $lpaData->lpa->getDonor()->getUId() === $code['ActorId']
+            ) {
```
SeemaMenon, line 99: Think the Id and Uid comparison with actorId doesn’t work as expected and hence by passes the if condition.
May be a == because they are not of same type?

cooperaj: We should always use strict comparisons. If they're not the same type the issue is how we make them so.

SeemaMenon: Agree on that

andrewpearce-digital (approved):

SeemaMenon (approved):

PR #1351, opened 2022-01-24
UML-1752 Conditional routing

lukejolliffe:
# Purpose

Allow tests to test the conditional routing under feature flags. Currently the routes are static based on the flag variable as set when the routes file is loaded.

This static routing caused issues when changing the feature flag values during our tests, to change the routing we would have to somehow change the route file before the test is set up.

Fixes UML-1752

## Approach

To enable the routing to change while the platform is running there is a ConditonalRoutingMiddleware this acts like a lazy loading middleware however it is given two potential handlers which are decided by the given flag value in the container. 

## Checklist

* ~~I have added relevant logging with appropriate levels to my code~~
* ~~New event_codes have been documented on the [wiki page]~~(https://opgtransform.atlassian.net/wiki/spaces/LSML2/pages/3277881441/Understanding+the+event+logs)
* ~~I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~~
* ~~I have added welsh translation tags and updated translation files~~
* ~~I have run an accessibility tool on any pages I have made changes to and fixed any issues found~~


service-front/app/src/Common/src/Middleware/Routing/ConditionalRoutingMiddleware.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
```
cooperaj, line 2: Missing declare strict.

service-front/app/src/Common/src/Middleware/Routing/ConditionalRoutingMiddleware.php
```diff
@@ -0,0 +1,46 @@
...
+use Psr\Http\Server\RequestHandlerInterface;
+
+class ConditionalRoutingMiddleware implements MiddlewareInterface
+{
+    private ContainerInterface $middlewareContainer;
+    private string $featureFlagName;
+    private string $trueRoute;
+    private string $falseRoute;
+
+    public function __construct(
```
cooperaj, line 20: Would likely need a docblock so we can explain why the container is being added here. Especially seeing as we avoid that literally everywhere possible otherwise.

service-front/app/test/CommonTest/Middleware/Routing/ConditionalRoutingMiddlewareTest.php
```diff
@@ -0,0 +1,70 @@
...
+            'FalseRoute'
+        );
+
+        $requestInterfaceProphecy = $this->prophesize(ServerRequestInterface::class);
+        $requestHandlerInterfaceProphecy = $this->prophesize(RequestHandlerInterface::class);
+
+        $trueRouteProphecy->handle(Argument::cetera())->shouldBeCalled();
+
+        $sut->process($requestInterfaceProphecy->reveal(), $requestHandlerInterfaceProphecy->reveal());
+    }
```
cooperaj, line 69: Needs test for when flag is undefined

service-front/app/src/Common/src/Middleware/Routing/ConditionalRoutingMiddleware.php
```diff
@@ -0,0 +1,53 @@
...
+class ConditionalRoutingMiddleware implements MiddlewareInterface
+{
+    private ContainerInterface $middlewareContainer;
+    private string $featureFlagName;
+    private string $trueRoute;
+    private string $falseRoute;
+
+    /**
+     * @param ContainerInterface $container The container needed to get the state of the feature flag when the route
+     *                                      is needed.
```
cooperaj, line 21:
```suggestion
     * @param ContainerInterface $container It is necessary to pass the container so we can resolve the feature flag
     *                                      at runtime. Passing just the FeatureEnabled component results in having to 
     *                                      re-initialise a number of container items to facilitate testing - increasing complexity.
```

service-front/app/test/CommonTest/Middleware/Routing/ConditionalRoutingMiddlewareTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+    public function setUp(): void
+    {
+        $this->containerProphecy = $this->prophesize(ContainerInterface::class);
+        $this->requestInterfaceProphecy = $this->prophesize(ServerRequestInterface::class);
+        $this->requestHandlerInterfaceProphecy = $this->prophesize(RequestHandlerInterface::class);
+    }
+
+    /** @test */
+    public function test_when_feature_flag_is_on_true_route_is_called()
```
cooperaj, line 28: Worth adding `void` to all test methods at this point.

cooperaj (changes requested):

cooperaj (approved):

PR #1356, opened 2022-01-27
Use correct npm invocation

cooperaj:
# Purpose

We're using the wrong installation command and it could result in dependencies not exactly matching what we're expecting. The correct command to use is `ci`. This PR replaces instances of `install` so that we're doing the right thing.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -674,7 +674,7 @@ orbs:
               name: Install dependencies
               command: |
                 if [ ! -d service-front/web/node_modules ]; then
-                  cd service-front/web/ && npm install
+                  cd service-front/web/ && npm ci
                 fi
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -714,20 +714,20 @@ orbs:
           - checkout
           - restore_cache:
               name: Restore node modules cache
-              key: node_modules-{{ checksum "service-front/web/package-lock.json" }}
+              key: node_modules-{{ checksum "service-front/web/package-lock.json" }}-{{ arch }}
```
cooperaj, line 717: Doing this means we match other places in the file for cache storage names.

andrewpearce-digital (approved):

PR #1412, opened 2022-03-08
UML-1547 Associate activated LPA records with the Activation Code used

lukejolliffe:
# Purpose

Associates the LPA number with the Activation code used to add it to the user's account. This allows us to track history of usage for Activation codes 

Fixes UML-1547

## Approach

Just adds a activation code parameter to "activateRecord" and "create" methods in ActorCodeService

## Checklist


service-api/app/src/App/src/DataAccess/DynamoDb/UserLpaActorMap.php
```diff
@@ -60,7 +60,8 @@ public function create(
         string $siriusUid,
         ?string $actorId,
         ?DateInterval $expiryInterval = null,
-        ?DateInterval $dueByInterval = null
+        ?DateInterval $dueByInterval = null,
+        ?string $code = null
```
cooperaj, line 64: Mmm. I would argue that the order on this methods parameters is bad. Is it more likely that we pass in a $code than a dueBy/Expiry? or even and actorId? If so then the $code should be before those parameters. To save the majority of our calls to this method having multiple nulls in the call.

lukejolliffe: It's called twice in Older using the date interval and only once when in the actor code service being added. 

Actor ID is the most passed in since it can be used by both. 

I'd like to make a mental note to go back to this when we bring API to 8.1 so we can just use named variables instead and make everything look nicer.

cooperaj (approved):

PR #1424, opened 2022-03-16
Adjust cookies to use Max-age instead of expires.

lukejolliffe:
# Purpose

Using Max-Age means that we don't have to use dates to calculate when the cookie will expire. That is all done by the browser. 

Contributes Towards UML-1873

## Learning

Max-age does similar to expires however you only need to give it expired seconds rather than dates. 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#:~:text=Max%2DAge%3D%3Cnumber%3E%20Optional

## Checklist


johnnolan (dismissed):

service-front/web/src/javascript/cookieHelper.js
```diff
@@ -18,27 +18,16 @@ const getCookie = name => {
...
-    if (window.location.protocol === 'https:') {
-        cookieString = `${cookieString}; Secure`;
-    }
-
-    return cookieString;
-}
-
-const setCookie = (name, value, options) => {
-    document.cookie = createCookie(name, value, options);
+    var maxAge = options.days * 24 * 60 * 60;
```
johnnolan, line 27: Can be a const

service-front/web/src/javascript/cookieHelper.js
```diff
@@ -18,27 +18,16 @@ const getCookie = name => {
...
+    var cookie =`${name}=${encodeURIComponent(value)}` + "; path=/" + "; max-age=" + maxAge + "; Secure";
```
johnnolan, line 28: Can be const

johnnolan (approved):

PR #1429, opened 2022-03-21
Uml 1873 cookie banner

lukejolliffe:
# Purpose

Update the cookie banner to the latest? GDS standard

Fixes UML-1873

## Approach

Adjust cookies so that only one Cookie used, show banner if that cookie is not set, don't show if cookie is set and has a value. 

## Checklist


service-front/web/src/javascript/cookieConsent.js
```diff
@@ -1,45 +1,75 @@
...
+    )[0];
+    cookieBanner.toggleAttribute('hidden', !show);
+    if (show) {
+      this._setCookieBannerClickEvents();
     }
+  }
 
-    _enableAllCookies(event) {
-        approveAllCookieTypes();
-        setCookie('seen_cookie_message', 'true');
```
cooperaj, line 37: Wheres the seen_cookie_message gone?

lukejolliffe: I removed it since it wasn't really needed. We can figure out if they have seen the cookie message by seeing if the cookie_policy is set.

cooperaj: I don't think thats strictly true. It's possible to have seen the banner and ignored it, maybe gone to the cookies page. In which case it has been seen and the defaults should be used.

cooperaj: Which IIRC is why it was put in.

PR #1466, opened 2022-04-06
Shade colour rather than defining our own

lukejolliffe:
# Purpose

Red should be shaded so that if gov colours ever change our system also updates

## Approach

govuk-shade 

## Learning

govuk-shade exists. There should be a discussion at some point if shade should be the 25% suggested by GDS

## Checklist

### DIFF
--- a/service-front/web/src/scss/ulpa-colours.scss
+++ b/service-front/web/src/scss/ulpa-colours.scss
@@ -1,1 +1,0 @@
-$higher-contrast-red: #aa2a16;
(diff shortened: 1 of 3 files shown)


service-front/web/src/scss/ulpa-colours.scss
```diff
@@ -1 +0,0 @@
-$higher-contrast-red: #aa2a16;
```
johnnolan, line 1: This file will be included somewhere that will need to be removed

lukejolliffe: 😑 forgot that

johnnolan (approved):

PR #1664, opened 2022-07-07
Uml 2266 server coverage

lukejolliffe:
# Purpose

Increase the coverage of the server.go file

Fixes UML-2266

## Approach

Make app a struct so that we can inject variables into code rather than constructing them in the file. 

Refactored middleware handlers by adding methods that return the handler function with extra parameters already in. This means that we can just call these with a handler which makes the chaining a lot neater. 

Write header is moved to the end of the error handler after all failures could occur. Writing the header in this way can only happen once so we need the last error that occurs to be reflected. 

## Checklist


andrewpearce-digital (dismissed):

service-admin/internal/server/server.go
```diff
@@ -41,38 +48,54 @@ func (w *errorInterceptResponseWriter) Write(p []byte) (int, error) {
...
+	a.r.PathPrefix("/").Handler(handlers.StaticHandler(os.DirFS("web/static")))
+
+	JSONHandler := NewJSONHandler(log.Logger)
+	templateHandler := NewTemplateHandler(LoadTemplates(os.DirFS("web/templates")))
+	errorHandler := NewErrorHandler(a.tw)
+
+	return JSONHandler(templateHandler(errorHandler(a.r)))
+}
+
+func NewAuthorisationHandler(token *auth.Token) func(http.Handler) http.Handler {
```
cooperaj, line 72: These arn't handlers. They're middlewares.

service-admin/internal/server/server.go
```diff
@@ -41,38 +48,54 @@ func (w *errorInterceptResponseWriter) Write(p []byte) (int, error) {
...
+	return func(h http.Handler) http.Handler {
```
cooperaj, line 73: Not quite what what wrapping with an additional handerfunc has achieved?

cooperaj (changes requested):
Will need a quick chat through this I think.

service-admin/internal/server/server_test.go
```diff
@@ -0,0 +1,251 @@
...
+			}}
+
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			defer func() { _ = recover() }()
+
+			got := server.WithErrorHandling(tt.args.next, tw)
+			assert.HTTPStatusCode(t, got.ServeHTTP, "GET", "/", nil, tt.expectedStatus)
+
```
lukejolliffe, line 205: t.Errorf("expected panic")

PR #1685, opened 2022-07-18
Uml-2267 signing test coverage

cooperaj:
# Purpose

Improving golang test coverage

Fixes UML-2267

## Checklist

* ~I have added relevant logging with appropriate levels to my code~
* ~New event_codes have been documented on the [wiki page](https://opgtransform.atlassian.net/wiki/spaces/LSML2/pages/3277881441/Understanding+the+event+logs)~
* ~I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~
* ~I have added welsh translation tags and updated translation files~
* ~I have run an accessibility tool on any pages I have made changes to and fixed any issues found~
* ~I have notified the Interaction Designer of any content changes so that appropriate screenshots/flow diagram changes can be made~
* ~The product team have tested these changes~

### DIFF
--- a/service-admin/internal/server/auth/signing.go
+++ b/service-admin/internal/server/auth/signing.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"fmt"
+	"github.com/rs/zerolog/log"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"sync"
@@ -14,11 +16,23 @@ import (
 	"github.com/sethvargo/go-retry"
 )
 
+type HTTPClient interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
 type SigningKey struct {
 	PublicKeyURL string
 }
 
-var publicKeyCache = &sync.Map{}
+var (
+	PublicKeyCache *sync.Map
+	Client         HTTPClient
+)
+
+func init() {
+	PublicKeyCache = &sync.Map{}
+	Client = &http.Client{}
+}
 
 func (k *SigningKey) Fetch(ctx context.Context, keyID string) (*ecdsa.PublicKey, error) {
 	if k.PublicKeyURL == "" {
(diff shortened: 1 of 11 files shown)


service-admin/internal/server/auth/signing_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"sync"
+	"testing"
+
+	"github.com/golang-jwt/jwt/v4"
+	. "github.com/ministryofjustice/opg-use-an-lpa/service-admin/internal/server/auth"
+	"github.com/stretchr/testify/assert"
+)
+
+type mockHTTPClient struct {
+	retryCount int
```
lukejolliffe, line 19: I think there is a variable missing or confused here. The retryCount here is being set to 2 because we have 2 expected retries. 

This is then given to the request func so that then starts with the count of 2, it will never hit the state of 0 and 1.

service-admin/internal/server/auth/signing_test.go
```diff
@@ -0,0 +1,144 @@
...
+				ctx:   context.Background(),
+				keyID: "test",
+			},
+			requestFunc: func(r *http.Request, count int) (*http.Response, error) {
+				switch count {
+				case 0:
+					// success
+					return &http.Response{
+						StatusCode: 200,
+						Body:       io.NopCloser(bytes.NewReader([]byte(goodKeyPem))),
```
lukejolliffe, line 98: This isn't a failed fetch, 404 is a vaild response so the code executes correctly and doesn't retry

service-admin/internal/server/auth/signing.go
```diff
@@ -37,6 +52,7 @@ func (k *SigningKey) Fetch(ctx context.Context, keyID string) (*ecdsa.PublicKey,
 	if err := retry.Do(ctx, r, func(ctx context.Context) error {
```
lukejolliffe, line 52: When I was testing with two retries in the test the 3 second timeout was not enough? Should the code be refactored so this isn't constant and can be configured?

lukejolliffe (changes requested):

service-admin/internal/server/auth/signing_test.go
cooperaj: See line 25 ;)

service-admin/internal/server/auth/signing_test.go
cooperaj: Due to the decrementing counter this is the final success condition i.e. it has already retried twice to get here.

lukejolliffe: If it starts at 2 though it should go to the default case and pass the test without failing.

PR #1687, opened 2022-07-19
Add coverage to dynamodb.go

lukejolliffe:
# Purpose

add coverage 

Fixes UML-2269

## Approach

prefix is now added to the DynamoConnection struct. This way it is not global and can be accessed in parallel by multiple tests. 

## Checklist

### DIFF
--- a/service-admin/internal/server/data/dynamo.go
+++ b/service-admin/internal/server/data/dynamo.go
@@ -8,9 +8,18 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-var prefix string
+type DynamoDBClient interface {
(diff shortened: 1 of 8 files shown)


service-admin/internal/server/data/dynamo.go
```diff
@@ -35,6 +35,6 @@ func NewDynamoConnection(region string, endpoint string, tablePrefix string) *dy
 	return svc
 }
 
-func prefixedTableName(name string) string {
+func PrefixedTableName(name string) string {
```
cooperaj, line 38: This does not need exporting as it's set in the constructor.

SeemaMenon (approved):

PR #1697, opened 2022-07-25
Fixes to twig template and css tweaks to improve accessibility.

cooperaj:
# Purpose

Improve screenreader accessibility of the language switcher

Fixes UML-2314

## Checklist


service-front/web/src/scss/_translation-switch.scss
```diff
@@ -1,14 +1,19 @@
...
     display: inline;
-  }
 
-  &__partition {
-    display: inline;
-    margin-right: 10px;
-    margin-left: 10px;
+    &:after {
+      padding: 0 5px;
+      content: " | ";
```
lukejolliffe, line 10: I get this being read out in voiceover, Even though it's not perfect GDS have used a border for their vertical line 

https://www.access.service.gov.uk/login/signin/creds

lukejolliffe (changes requested):

service-front/app/src/Common/templates/partials/welsh-switch.html.twig
```diff
@@ -1,28 +1,30 @@
 {% set route_name = get_route_name() %}
 
-<div class="trans-switch" role="navigation" aria-labelledby="language-switch">
-    <ul class="trans-switch__container" id="language-switch" aria-label="english-welsh-language-switch">
-        <li class="govuk-list trans-switch__list govuk-body-s">
+<div class="trans-switch" role="navigation" aria-labelledby="language-switch-en language-switch-cy">
```
lukejolliffe, line 3: Does this div give us anything? I think one of the main criticisms of the language toggle that the screen reader user had was that there were too many levels/ things that get read out that make the whole thing confusing? 

Maybe I got the wrong end of the stick, but I would thing that it would be better if it read just simply

Change Language 
    Current Language is English
    Newid iaith i'r Gymraeg

Currently it reads :

Current Language is English | Newid iaith i'r Gymraeg
   Current Language is English
    Newid iaith i'r Gymraeg

lukejolliffe (changes requested):

service-front/app/src/Common/templates/partials/welsh-switch.html.twig
cooperaj: The issue is that we need to read out the language in the appropriate language. The only way to inform the browser of said language is using lang="" attributes and theres no way to attach those to an aria-label so my hope is that pulling that text from appropriate elements makes it work.

service-front/web/src/scss/_translation-switch.scss
cooperaj: Yeah, but it also looks worse :p

cooperaj: Ok. So the pipe that's being read out isn't even the one thats added with the CSS. I've no idea where it's getting it as it's supposed to be the concatenation of the labels (aria-labelledby). Will continue to poke.

service-front/app/src/Common/templates/partials/welsh-switch.html.twig
cooperaj: Div dropped and just use english for the label. I'm not sure we can do much else.

lukejolliffe (approved):

lukejolliffe (changes requested):
Seems like axe dev now has an issue with it, hopefully its a simple fix 

EDIT: If you remove the role="navigation" from the ul then it fixes it :)

lukejolliffe (approved):

PR #1739, opened 2022-08-08
Uml 2250 get activation key from codes repo

lukejolliffe:
# Purpose

Query the codes table to check for existence of activation key

Fixes UML-2250

## Approach

activation_key.go contains two implementations of the activation key service. One to query the local codes table for dev environments and one to query the actual codes api with a aws v4 signed post request to the api. 

Both implementations give back a Array of ActivationKeys that the first key will be given to the template to render the information from.  

## Learning

Aws v4 signer now needs the Sha256 hash of the body as a hex string. Without this we were having issues with a signing error. 

Recieved is actually spelt Received

Go templating struggles to get values out of a nested anonymous interface, hence the decision to go for defined structs for both the response and activation key

There are multiple sites that will take json and turn it into a go struct for you. Found this out AFTER writing one myself :) 

## Checklist


service-admin/.vscode/launch.json
```diff
@@ -16,7 +16,8 @@
                 "ADMIN_JWT_SIGNING_KEY_URL": "http://localhost:5000",
                 "AWS_ACCESS_KEY_ID": "-",
                 "AWS_SECRET_ACCESS_KEY": "-",
-                "ADMIN_LOGOUT_URL": "https://login-admin-lastingpowerofattorney.auth.eu-west-1.amazoncognito.com/logout"
+                "ADMIN_LOGOUT_URL": "https://login-admin-lastingpowerofattorney.auth.eu-west-1.amazoncognito.com/logout",
+                //"LPA_CODES_API_ENDPOINT": "https://dev.lpa-codes.api.opg.service.justice.gov.uk"
```
SeemaMenon, line 20: commented code in here

lukejolliffe, line 20:
```suggestion
```

SeemaMenon (approved):

PR #1798, opened 2022-08-24
fix searching for email address

lukejolliffe:
# Purpose

Fix searching for email addresses 

Fixes UML-2648

## Approach

Go back to stripping unnecessary characters for each function rather than only calling once at beginning

## Checklist

### DIFF
--- a/service-admin/internal/server/data/activation_key.go
+++ b/service-admin/internal/server/data/activation_key.go
@@ -73,7 +73,7 @@ func (aks *OnlineActivationKeyService) GetActivationKeyFromCodes(ctx context.Con
 	hasher.Write(jsonStr)
 	shaHash := hex.EncodeToString(hasher.Sum(nil))
 
-	credentials, err := aks.config.Credentials.Retrieve(context.TODO())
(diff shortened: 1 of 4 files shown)


service-admin/internal/server/data/activation_key.go
```diff
@@ -73,7 +73,7 @@ func (aks *OnlineActivationKeyService) GetActivationKeyFromCodes(ctx context.Con
 	hasher.Write(jsonStr)
 	shaHash := hex.EncodeToString(hasher.Sum(nil))
 
-	credentials, err := aks.config.Credentials.Retrieve(context.Background())
+	credentials, err := aks.config.Credentials.Retrieve(context.TODO())
```
johnnolan, line 76: TODO? Is that a thing or a typo?

lukejolliffe: It is a thing and makes no difference but I'll change it back anyway!
```suggestion
	credentials, err := aks.config.Credentials.Retrieve(context.Background())
```

johnnolan (approved):

PR #1880, opened 2022-09-12
Add hidden string to authenticate method

lukejolliffe:
# Purpose

Use hidden string to avoid accidental printing of plaintext credentials

Fixes UML-2655 

## Checklist


lukejolliffe:
I've fixed the PHPCs stuff in my other PR

PR #1936, opened 2022-09-29
Uml 2009 make delete clearer

lukejolliffe:
# Purpose

Add donors name and LPA type into remove LPA warning page to make sure user is deleting the correct LPA
Fixes UML-2009

I've added in a function to LPA extensions to convert from hw, pf -> health and welfare, property and finance

## Checklist


service-front/app/languages/cy/LC_MESSAGES/messages.po
```diff
@@ -7511,3 +7501,12 @@ msgstr ""
 "<strong>Pwysig: </strong>Mae gan yr LPA hon gyfarwyddiadau neu ddymuniadau."
 "<br>\n"
 "            Efallai y bydd angen i chi ddangos LPA papur i sefydliadau."
+
+# Current translation is "Are you sure you want to remove this LPA?"
+#: src/Actor/templates/actor/confirm-remove-lpa.html.twig:15
+msgid "Are you sure you want to remove %firstname% %surname%'s LPA?"
+msgstr "A ydych yn siŵr eich bod am ddileu’r LPA hon?"
```
cooperaj, line 7508: The translated string does not contain either of the two parameters?

service-front/app/src/Actor/templates/actor/confirm-remove-lpa.html.twig
```diff
@@ -12,7 +12,14 @@
...
                 <div class="govuk-grid-column-two-thirds">
-                    <h1 class="govuk-heading-xl">{% trans %}Are you sure you want to remove this LPA?{% endtrans %}</h1>
+                    <h1 class="govuk-heading-xl">{% trans with
+                            {
+                                '%firstname%': lpa.getDonor.firstname,
+                                '%surname%': lpa.getDonor.surname,
+                            }
+                        %}Are you sure you want to remove %firstname% %surname%'s LPA?{% endtrans %}</h1>
+
+                    <p class="govuk-body"><strong>{% trans %}LPA type:{% endtrans %}</strong> {{ lpa_type_full_name(lpa.getCaseSubtype())|trans }}</p>
```
cooperaj, line 22: This trans call will not extract anything useful anywhere. It's likely the case that 'Health and welfare' and 'Property and finance' are already extracted but this is sill going to be a missing reference. 

I'd much rather see this done as a single statement too i.e. the lpa type being a parameter of the 'LPA type:' string.

service-front/app/src/Common/src/View/Twig/LpaExtension.php
```diff
@@ -34,15 +33,26 @@ public function getFunctions(): array
...
             new TwigFunction('donor_name_with_dob_removed', [$this, 'donorNameWithDobRemoved']),
             new TwigFunction('is_donor_signature_date_too_old', [$this, 'isDonorSignatureDateOld']),
+            new TwigFunction('lpa_type_full_name', [$this, 'lpaTypeFullName'])
         ];
     }
 
+    public function lpaTypeFullName(string $typeShorthand): string
+    {
+        if ($typeShorthand === 'hw') {
+            return 'Health and welfare';
```
cooperaj, line 43: Potentially the best thing to do here would be to use a proper translation function call, like happens with some of the flash message contents in handlers. That way the extract path is more clear.

service-front/app/src/Common/src/View/Twig/LpaExtension.php
```diff
@@ -175,7 +187,8 @@ public function isCodeCancelled(?array $code): ?bool
      */
     public function hasCodeExpired(?string $expiryDate): ?bool
     {
-        if (!empty($expiryDate && $date = new DateTime($expiryDate))) {
+        if ($expiryDate) {
```
cooperaj, line 190: Would prefer to see this as a more strongly typed check.

service-front/app/src/Common/src/View/Twig/LpaExtension.php
```diff
@@ -215,23 +229,16 @@ public function isDonorSignatureDateOld(Lpa $lpa): bool
      * Creates an international date formatter that is capable of doing locale based dates.
      *
      * @param string $locale
-     * @param string|null $pattern Optional pattern to format the date as
      * @return IntlDateFormatter
      */
-    private function getDateFormatter(string $locale, ?string $pattern): IntlDateFormatter
```
cooperaj, line 221: Why the change in signature?

cooperaj (changes requested):
few things to clear up.

service-front/app/languages/cy/LC_MESSAGES/messages.po
lukejolliffe: That's deliberate since the welsh isn't ready it's getting written into another ticket.

service-front/app/src/Common/src/View/Twig/LpaExtension.php
lukejolliffe:
```suggestion
        if ($expiryDate !== null) {
```

service-front/app/src/Common/src/View/Twig/LpaExtension.php
lukejolliffe: It was never bering used with the pattern. We were just calling it with a static null

service-front/app/src/Common/src/View/Twig/LpaExtension.php
lukejolliffe: @cooperaj I've turned this back to a loosy goosy check because our tests expect an empty string to result in null.

SeemaMenon (approved):

cooperaj (approved):
Suggestion to use null safe assignment rather than a ternary that returns null.

PR #1982, opened 2022-10-12
Uml-2559 - all non scheduled, non path to live based GH actions

jamesrwarren:
# Purpose

Convert non scheduled, non path to live workflows to github actions

Fixes UML-2559

## Approach

This consists of 4 workflows basically. Dependabot only workflows, admin only, terraform only and standard branch.

There are some annoying concessions that I had to make to get this to work.

1) Paths just don't cut it. It's impossible to have one workflow for for path A, B , C etc and have it run different workflow for just path A. You could do one workflow for path B and one for path A but you can't say that if files have only changed on one path then do one workflow but if it's changed on that path and others then do another workflow. Relying on some sort of labelling or commit messages seemed wrong also so I went for the option of having a single workflow with skipping logic in it dependent on the path taken. I use the changed-files action to check if only one path has changed and then use that as a future filter during the workflow.
This had a further issue....

2) if you skip a job and it's needed for next job to run then it doesn't run. This almost scuppered the plan except we can put in some fairly ugly logic that says always run the workflows but check if the 'need' workflows are success or skipped. This way it still stops on failure but carries on if skipped. It would be nice if this was default behaviour really.

3) dependabot workflows were different enough that I really wanted separate workflow. This is easy to do on push as we can just specify the branch. However I also wanted the main PR workflow to be on pull_request but this only has the filter of what you're merging into (main). There's no further way of filtering at the top level (like based on user or anything else) so if we made it on pull_request then we couldn't filter between the two workflows... So another concession is having the workflow run on push. If anyone has bright ideas on better solution then I'm interested.

## Learning

Quite a lot around the logic of github actions. In some ways it has some nice features for reusing code but in other ways it has some quite frustrating gaps. 

## Checklist


lukejolliffe (approved):

lukejolliffe (changes requested):
Investigate build time

lukejolliffe (dismissed):

lukejolliffe (approved):

PR #1987, opened 2022-10-17
Uml 2008 esbuild

lukejolliffe:
# Purpose

Improve build times and streamline front web builds
Fixes UML-2008

## Approach

Node has been updated to 18 so that we can use the cp function in the build. 

I've made it so that when we build we run babel on all the JS in the src folder to turn es6 to es5 and polyfill. Then esbuild will be able to bundle the JS and polyfills. 

Esbuild still needs to target es2015 and ie11 because even though it currently isn't capable of turning ES6 code to ES5 these options ensure that it doesn't write es6 code. 

Esbuild was also struggling to find the paths for assets, I've fixed this by using the scss variables for both moj pattern library ($moj-images-path) and gds ($govuk-assets-path), i've set this path to be relative paths to assets folders rather than defaulting to using the absolute path of /assests/ like it was before. 

To create the PDF css file I've used an import of gds in the file to bring in the files that we need then added base64 loaders for the font and svg assets. This now happens on npm build since it's really fast I thought it was pretty redundant having a second command and build.js file for it. 

Once I had set the Babel stuff up correctly to allow use with IE11 the tests started breaking again, this was because it was trying to pull in the babel config that we are using for the build. Due to this I have separated the babel config to two env options. These are used by supplying BABEL_ENV=<env_name> before the babel command. 

 

## Learning

## Checklist


service-front/web/build.js
```diff
@@ -0,0 +1,76 @@
...
+      outdir: config.out_dir,
+      minify: false,
+      plugins: [],
+      target: ['es2015', 'ie11'],
+      platform: 'browser',
+      metafile: true,
+      supported: {
+        arrow: false,
+      },
+      treeShaking: false,
```
lukejolliffe, line 46:
```suggestion
```

SeemaMenon (approved):

PR #2070, opened 2022-11-11
UML-2561 create path to live in GH actions

jamesrwarren:
# Purpose

Final part of GH switchover. Move the path to live to github actions

Fixes UML-2561

## Approach

Tried out the basic code with ephemeral branch replacing dev/pre/prod etc and it's all working (see actions workflows for example) so this *should* work out the box.

Couple of things though when releasing this one:
- Need to do small follow up afterwards to get rid of circle config and need a small PR in org infra to remove the circle vars and webhook.
- We should release this when there isn't any urgent fixes going out just in case there's an issue as I've checked as far as I can but obviously haven't tested with the real environments yet!

## Learning

NA

## Checklist


lukejolliffe (approved):
Prism has moved on 4 patches since this last build we were using but it's still broken so I won't comment on the removal in docker-compose.dependencies.yml

PR #2117, opened 2022-12-06
UML-2739 allow dashboard to load on 404 from sirius

lukejolliffe:
# Purpose

404 cause the dashboard to end in 500 error when loading LPAs we would like this to not cause the user to see an error but instead a lack of LPAs for the ones that have failed in their account

Fixes UML-2739

## Approach

instead of failing on array key when looping through returned arrays, check for null and continue the loop if it fails.

## Checklist


service-api/app/src/App/src/Service/Lpa/LpaService.php
```diff
@@ -278,7 +278,12 @@ private function lookupAndFormatLpas($lpaActorMaps): array
 
         // Map the results...
         foreach ($lpaActorMaps as $item) {
-            $lpa = $lpas[$item['SiriusUid']];
+            $lpa = $lpas[$item['SiriusUid']]?? null;
+
+            if ($lpa == null) {
```
cooperaj, line 283:
```suggestion
            if ($lpa === null) {
```

PR #2127, opened 2022-12-13
Uml-2423 scan vulnerabilities

jamesrwarren:
# Purpose

We do trivy scanning but the ECR scanning wasn't brought over from circleci. This enables it on a per PR basis.

Fixes UML-2423

## Approach

Decided to slightly modify the script so that it will fail the build if it finds anything too problematic. 

I put it in the workflow alongside the behat tests as it needs a bit of time for the scan to happen on the aws side before we can get the results so during the apply this gives us that time.

Also checks scan results for our pdf service but I decided against killing the build if there was an issue with this. I would like a nice way of alerting us to the issue that isn't spamming slack.. will think about that as an improvement.

Fixed the issues found by trivy and ecr on the lambda image.

## Learning

NA

## Checklist


lambda-functions/upload-statistics/Dockerfile
```diff
@@ -43,6 +43,8 @@ FROM python-alpine
 ARG FUNCTION_DIR
 # Set working directory to function root directory
 WORKDIR ${FUNCTION_DIR}
+# Uninstall installation packages
+RUN pip uninstall -y wheel setuptools
```
jamesrwarren, line 47: shouldn't need these at all now that packages are installed

cooperaj (approved):

PR #2156, opened 2023-01-05
add lock timeout to plan command. 

lukejolliffe:
# Purpose

If TF lock is in conflict currently this fails the PR straight away. Adding a timeout will allow the PR to wait up to 5 minutes before failing. If the lock becomes available again in this time the TF plan should go ahead.

## Learning

https://developer.hashicorp.com/terraform/cli/commands/plan#lock-timeout-duration:~:text=%2Dlock%2Dtimeout%3DDURATION)

## Checklist

### DIFF
--- a/.github/workflows/_run-terraform.yml
+++ b/.github/workflows/_run-terraform.yml
@@ -118,7 +118,7 @@ jobs:
           TF_VAR_admin_container_version: ${{ steps.version-output.outputs.admin-tag }}
         run: |
           terraform workspace show
-          terraform plan -input=false -parallelism=30
+          terraform plan -input=false -parallelism=30 -lock-timeout=5m
         working-directory: terraform/${{ inputs.terraform_path }}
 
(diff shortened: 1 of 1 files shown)


SeemaMenon (approved):

johnnolan (commented):
LGTM but unsure of any infra issues with the lock timeout. Will defer to @jamesrwarren on that.

Also what happens if 1 locks but 2 are waiting, will they both take turns or try at the same time causing the same error again?

lukejolliffe:
> LGTM but unsure of any infra issues with the lock timeout. Will defer to @jamesrwarren on that.
> 
> Also what happens if 1 locks but 2 are waiting, will they both take turns or try at the same time causing the same error again?

I'm assuming one of them would take the lock and the other would keep waiting. (probably randomly based on whichever retries getting the lock first)  Then if it ends up taking > 5 mins it'll time out.

PR #2364, opened 2023-10-11
UML-3115 Move ECS Terraform to region module

ntse:
# Purpose

Move all ECS-related Terraform resources to the region module in as part of the work to allow for multi-region deployments.

Fixes UML-3115

## Approach

Move all of the ECS-related code and use temporary variables to pass in identifiers of resources created outside of the region module (e.g. loadbalancers). Create a refactor.tf file to update Terraform state.

## Learning

## Checklist


terraform/environment/region/terraform.tf
```diff
@@ -0,0 +1,18 @@
+terraform {
+  required_version = ">= 1.5.6"
+
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+      configuration_aliases = [
+        aws.region,
```
andrewpearce-digital, line 8: the eu-west-1 instance of the module passes in providers for region and management, you should specify them here. I think it might otherwise use the wrong provider when no aliased provider is given.

I'm not sure how it might impact the plan/apply but i find it be explicit about the provider being used. When a provider is not properly given, it may use the user it finds credentials for which may plan fine but fails to apply due to permissions or the wrong account.

terraform/environment/terraform.tf
```diff
@@ -47,6 +47,18 @@ provider "aws" {
   }
 }
 
+provider "aws" {
+  region = "eu-west-1"
+  alias = "eu_west_1"
```
andrewpearce-digital, line 52: I think if you add the alias to the pre-existing eu-west-1 provider above, you could avoid "moved/new provider" issues but you will have to specify the provider in every resource/module that was using it by default

andrewpearce-digital (changes requested):
I think there might be an issue with the way providers are defined and passed in. Specifying a region provider, and passing in an unspecified management provider might result in the management provider being used wherever you have not told a resource what provider to use.

Also could use a fmt

terraform/environment/terraform.tf
andrewpearce-digital: The strategy to move all resources fromt he default provider addresses this comment

andrewpearce-digital (approved):

PR #2390, opened 2023-10-30
UML-3121 make redirect uri dynamic

MishNajam:
# Purpose

Make redirect_uri use the correct domain depending on environments

Fixes UML-3121

## Approach

Uses the Mezzio serverUrlHelper and urlHelper classes in service-front and passes this down to service-api.

## Learning

The `$redirectUrl` being passed from the frontend to the api needs to match the `redirect_url` set in the docker dependencies file. 
Selecting Welsh modifies the URL by adding '/cy/'. We have removed this from the redirectUrl to make this match. 
The ui_locale value is stored in the AuthSession to allow returning from one login to a welsh page if selected. 

## Checklist

- I have added relevant logging with appropriate levels to my code
- New event_codes have been documented on the [wiki page](https://opgtransform.atlassian.net/wiki/spaces/LSML2/pages/3277881441/Understanding+the+event+logs)
- I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant
- I have added welsh translation tags and updated translation files
- I have run an accessibility tool on any pages I have made changes to and fixed any issues found
- I have notified the Interaction Designer of any content changes so that appropriate screenshots/flow diagram changes can be made
- The product team have tested these changes


service-front/app/features/one-login.feature
```diff
@@ -6,3 +6,10 @@
       Given I am on the temporary one login page
       When I click the one login button
       Then I am redirected to the redirect page
+
+    @ui @actor @ff:allow_gov_one_login:true @welsh
+    Scenario: I initiate authentication via one login in Welsh
+      Given I am on the temporary one login page
+      And I request to view the content in welsh
+      When I click the one login button in welsh
```
cooperaj, line 14: This should ideally be `When I click the login button`. The assertion on the language would then be in the Then.

The usefulness of the context files is that a step can create state that a later step can access. In this case you'd want to set the locale in the context when you request to view the content in welsh.

cooperaj (approved):

PR #2399, opened 2023-11-06
Endpoint to handle callback from OneLogin

cooperaj:
# Purpose

This endpoint will handle end user requests upon return from One Login. At this stage it will return a dummy /  “hello world” response. This enables testing of the end to end control. 

Fixes UML-3102

## Learning

Going to need a rework of the smoke tests or implementation of alternative ui/acceptance testing for the end to end part.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,12 @@ run:
 	$(COMPOSE) run $(filter-out $@,$(MAKECMDGOALS))
 .PHONY: run
 
+pull:
+	@echo "Logging into ECR..."
+	$(ECR_LOGIN)
+	$(COMPOSE) pull
+.PHONY: pull
+
 # Starts the application and seeds initial data.
 up_all: | up_dependencies up_mock up_services up_functions seed
 .PHONY: up_all
--- a/service-front/app/config/routes.php
+++ b/service-front/app/config/routes.php
@@ -62,7 +62,6 @@ $viewerRoutes = function (Application $app, MiddlewareFactory $factory, Containe
 };
 
 $actorRoutes = function (Application $app, MiddlewareFactory $factory, ContainerInterface $container): void {
-    $USE_OLDER_LPA_JOURNEY = 'use_older_lpa_journey';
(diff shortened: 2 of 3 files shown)


service-front/app/config/routes.php
```diff
@@ -100,7 +99,7 @@
     $app->get('/session-expired', Actor\Handler\ActorSessionExpiredHandler::class, 'session-expired');
     $app->get('/session-check', Actor\Handler\ActorSessionCheckHandler::class, 'session-check');
     $app->get('/session-refresh', Common\Handler\SessionRefreshHandler::class, 'session-refresh');
-    $app->get('/home/login', Actor\Handler\LoginPageHandler::class, 'auth-redirect');
+    $app->get('/home/login', Actor\Handler\OneLoginCallbackHandler::class, 'auth-redirect');
```
MishNajam, line 102: Would this need to be behind a feature flag now?

Makefile
```diff
@@ -36,6 +36,12 @@ run:
 	$(COMPOSE) run $(filter-out $@,$(MAKECMDGOALS))
 .PHONY: run
 
+pull:
```
MishNajam, line 39: For my own understanding: why was this added 🤔

service-front/app/config/routes.php
cooperaj: Potentially. But since it's a route that *only* works when accessed as a part of a legitimate flow and does nothing otherwise (or will do, curently it works in that you'll get the hello world) it's not strictly needed. Can add if you'd prefer.

Makefile
cooperaj: make build only builds things but since we're now using images hosted on AWS more (read, the one login mock) we need a way to also update those if needed.

service-front/app/config/routes.php
cooperaj: I've added a wrapper for the correct feature flag now that 404's if it's not turned on. I've also added some basic sanity checking so it can't be accessed incorrectly

MishNajam (approved):

PR #2426, opened 2023-11-14
Uml 2912 make Use work properly via https, locally. (WIP)

nickdavis2001:
# Purpose

_Currently Use does not work via https "out of the box" locally, which makes development awkward e:g many smoke tests fail locally. It would be great to get all smoke tests passing locally_

## Approach

_Add ssl proxy containers in front of viewer-web and actor-web in the same way as is currently done for Make_

## Learning

_Learned that in fact the current certificates setup for Make doesn't actually work, but cypress bypasses this so that it appears to work!_

## Checklist


local-ssl/certs/ca.key
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
mattmachell, line 1: Even though these aren't real certs and are only for local dev, committing can trip a bunch of secret scanners and processes (dependabot/trivy etc), so better to have a certificate generating script instead (DigiDeps has this if you need an example).

PR #2503, opened 2024-01-16
UML-3249 Upgrade packages in Lambda function

ntse:
# Purpose

Resolve security alerts for stats upload Lambda.

Fixes UML-3249

## Approach

Upgrade the system packages of the stats_upload Lambda using Debian's `apt` package manager.

## Learning

## Checklist

### DIFF
--- a/lambda-functions/upload-statistics/Dockerfile
+++ b/lambda-functions/upload-statistics/Dockerfile
@@ -3,7 +3,7 @@ ARG FUNCTION_DIR="/function"
 
 FROM python:3.11-slim
 
-RUN apt -y update && apt -y install gcc python3-dev
+RUN apt --assume-yes update && apt --assume-yes upgrade && apt --assume-yes install gcc python3-dev 
 
 ARG FUNCTION_DIR
 
--- a/terraform/environment/region/actor_load_balancer.tf
+++ b/terraform/environment/region/actor_load_balancer.tf
@@ -111,7 +111,7 @@ resource "aws_lb_listener_rule" "redirect_use_root_to_gov" {
 
 # rewrite to live service url
 resource "aws_lb_listener_rule" "rewrite_use_to_live_service_url" {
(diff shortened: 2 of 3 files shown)


lambda-functions/upload-statistics/Dockerfile
```diff
@@ -3,7 +3,7 @@ ARG FUNCTION_DIR="/function"
 
 FROM python:3.11-slim
 
-RUN apt -y update && apt -y install gcc python3-dev
+RUN apt --assume-yes update && apt --assume-yes upgrade && apt --assume-yes install gcc python3-dev
```
cooperaj, line 6: didn't think we were supposed to update images like this if we could help it?

ntse: Previously, this was a best practice suggested by Docker but has been removed from their best practice guidelines:

The current version is here: https://docs.docker.com/develop/develop-images/instructions/#apt-get but you can see in an older mirror that the justifications for it didn't make much sense in any case (which I suspect is why it's now removed)

https://docker-docs.uclv.cu/develop/develop-images/dockerfile_best-practices/#run

It was previously recommended to avoid it because you can't run apt-get install in an unprivileged container but we switch between root and the application user in the Dockerfile so this isn't a problem. It also suggested contacting the upstream maintainers for updating packaging which isn't something that works well in reality.

The only issue that I can see here (which is an issue with all of our alpine containers) is that the latest versions of packages in the apt repos could change between a Pull Request completing and passing its integration tests and it being rebuilt in the path to live which would mean we have two slightly different containers. I've raised UML-3252 to fix that

cooperaj: I agree 100%. All my personal images do an upgrade as the first step as those reasons always seemed spurious at best.

cooperaj (approved):

PR #2524, opened 2024-02-07
Redirect /login to /home

Lbagg1:
# Purpose
The login form needs to be redirected

/login needs redirecting to /home

AC

redirect/login to /home

Using one login feature flag

Tests need to pass

Fixes UML-3267

## Approach

## Learning

## Checklist


service-front/app/config/routes.php
```diff
@@ -95,13 +96,25 @@
...
     // User auth
-    $app->route('/login', Actor\Handler\LoginPageHandler::class, ['GET', 'POST'], 'login');
+    $app->route('/login', [
+        new ConditionalRoutingMiddleware(
+            $container,
+            $factory,
+            $ALLOW_GOV_ONE_LOGIN,
+            function ($request, $handler) use ($container) {
+                return new \Laminas\Diactoros\Response\RedirectResponse('/home');
+            },
```
cooperaj, line 106:
```suggestion
            function ($request, $handler) {
                return new \Laminas\Diactoros\Response\RedirectResponse('/home');
            },
```

Forgot to deal with this yesterday. We no longer need the container in our anonymous function so this can be removed.

service-front/app/src/Common/src/Middleware/Routing/ConditionalRoutingMiddleware.php
```diff
@@ -13,25 +14,22 @@
 
 class ConditionalRoutingMiddleware implements MiddlewareInterface
 {
-    private ContainerInterface $middlewareContainer;
 
     /**
-     * @param ContainerInterface $container It is necessary to pass the container so we can resolve the feature flag
```
cooperaj, line 19: Should probably keep this comment against the ContainerInterface. Not sure how it came to disappear

cooperaj (changes requested):

service-front/app/config/routes.php
cooperaj: We don't technically use the request or handler either so maybe try to remove those too and see what happens?

In something like Go this would break as the function then would not match the expected interface but I think PHP will just roll with it.

cooperaj (approved):

PR #2611, opened 2024-05-20
Uml 3404 smoke working against mock

cooperaj:
# Purpose

Allows the successful completion of smoke tests when the login is provided by the One Login mock we have running in the environment.

Fixes UML-3404

## Approach

Implementation of feature flag in smoke tests that allows us to test differing pathways.

## Learning

Because we cache the information provided by the mock it will only function at a local development machine level if the stack has been brought fully down and back up, and then not subsequently accessed before running the test suite.

## Checklist


docker-compose.yml
```diff
@@ -81,6 +71,8 @@ services:
       COOKIE_EXPIRES: 1440 # cookie expiry for complete logout - initial value to be 24 hours.
     depends_on:
       - redis
+      - api-web
+      - service-pdf
```
gregtyler, line 75: It looks like you've removed `service-pdf`, so how can this depend on it?

tests/features/actor-login.feature
```diff
@@ -8,57 +8,8 @@ Feature: A user of the system is able to login
...
   Scenario: A user can login
     Given I access the login form
     When I enter correct credentials
     Then I am signed in
-
-  @acceptance
-    Scenario: A user cannot login with an incorrect password
-    Given I access the login form
-    When I enter incorrect login password
-    Then I am told my credentials are incorrect
```
gregtyler, line 21: Apologies if this is a silly question, but did you mean to remove these tests? I'd think we'd want to keep testing the old login flow until we remove it.

docker-compose.yml
cooperaj: In this case it's because the viewer app isn't actually dependent on the pdf service as it talks to the service-api and thats the one that talks to the pdf service.

cooperaj: Mmm. Good question. We never actually bring up the stack without also including the dependencies compose file (it's defined in there). I think what I can do is add an override to the dependencies file for the viewer-app that adds this.

tests/features/actor-login.feature
cooperaj: These tests were never actually run. The only tests run here are tagged `@smoke`. 

At this point it's just a throwback to when I wanted a single set of feature files rather than the three separate locations we have.

gregtyler: Ah fair enough 😄

gregtyler (approved):
LGTM but, as you say, could do with a range of views

allenannom (approved):

jay-whitwell (approved):

PR #2632, opened 2024-06-10
Uml 3439 enable opcache

cooperaj:
# Purpose

Enables opcache in PHP to improve application resource usage and speed. Also enable the DI container cache.

Fixes UML-3439

## Checklist

### DIFF
--- a/service-api/docker/app/app-php.development.ini
+++ b/service-api/docker/app/app-php.development.ini
@@ -2,3 +2,8 @@ log_errors = On
 expose_php = On
 display_errors = stderr
 error_reporting = E_ALL
+
+opcache.memory_consumption=128
+opcache.interned_strings_buffer=8
+opcache.max_accelerated_files=4000
+opcache.revalidate_freq=0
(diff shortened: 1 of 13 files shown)


service-api/docker/app/app-php.development.ini
```diff
@@ -2,3 +2,8 @@ log_errors = On
 expose_php = On
 display_errors = stderr
 error_reporting = E_ALL
+
+opcache.memory_consumption=128
+opcache.interned_strings_buffer=8
+opcache.max_accelerated_files=4000
```
mattmachell, line 8: Did you check the memory consumption vs the ECS task definition size?

cooperaj: Yep. The task is current set to 1GB so this should work ok. The default PHP-FPM size is the same 128MB so it's only using up "one" worker. 

That said, for some reason the PHP-FPM is set to have a static 25 threads spun out, which would never happen with the memory the task has. Looks like we're leaning on the AWS autoscaling even more than I thought

nickdavis2001 (approved):

allenannom (approved):

PR #2660, opened 2024-06-18
Ensure email and password changing endpoints return 410 gone

Lbagg1:
# Purpose

To ensure users cannot get to these pages manually

AC 

behind One Gov Logon feature flag

change-password endpoint

change-email endpoint

Fixes UML-3331

## Approach

## Learning

## Checklist


service-front/app/src/Common/src/Handler/Factory/GoneHandlerFactory.php
cooperaj, line 1: Because the GoneHandler's constructor does not use configuration values and just easily resolvable classes it shouldn't need a factory at all. Thats just all automagically done for you.

cooperaj (dismissed):
I really like this solution but I also quite like the way the middleware acted as a sort of catchall for old routes that might not be in the routes file anymore.

This works for what we need right now so I think that's ok but a potential future bit of work would be to introduce a GoneMiddleware that sits where the original did in the pipeline and lets us handle routes that eventually get removed completely from the routes file.

cooperaj (approved):
Very nice.

PR #2704, opened 2024-07-12
Fix up the healthcheck endpoint

cooperaj:
## Approach

Switch out to the PSR standards http client code introduced earlier for the healthcheck.

Additionally ensured that the healthcheck works locally against the configured mocks.

## Checklist


service-api/app/src/App/src/DataAccess/ApiGateway/RequestSigner.php
```diff
@@ -25,10 +25,12 @@ public function __construct(readonly private SignatureV4 $signer, array $headers
...
     public function sign(RequestInterface $request): RequestInterface
     {
+        $request = $this->signer->signRequest($request, $this->credentials);
+
         foreach ($this->headers as $header => $value) {
             $request = $request->withHeader($header, $value);
         }
 
-        return $this->signer->signRequest($request, $this->credentials);
+        return $request;
```
gregtyler, line 34: I think this may cause signing to break, since V4 signatures include the headers as part of the signature. If you add headers _after_ signing, they'll be included in the signature AWS generates server-side and the signatures won't match,

service-api/app/src/App/src/Handler/HealthcheckHandler.php
```diff
@@ -86,24 +121,16 @@ private function checkDynamoEndpoint(): array
...
-    private function checkCodesApiEndpoint(): array
-    {
-        $url = sprintf('%s/v1/healthcheck', $this->codesApiUrl);
-
-        return $this->apiCall(new Request('GET', $url));
-    }
-
     /**
-     * @param RequestInterface $request
+     * @param RequestInterface $uri
```
gregtyler, line 125: Is this right? The PHP typehints suggest the parameter is still called `$request`

service-api/app/src/App/src/DataAccess/ApiGateway/RequestSigner.php
cooperaj: Mmm. Darn. Hadn't considered that. 

The issue here is that the mock codes service expects a static Authorisation header value, hence overwriting it after the fact works.

If you set that value first then the aws signer overwrites it with it's own. 

I guess this class needs a different approach to handle that case. I think what I'll do is treat the 'Authorization' header as a special case. If you're setting that manually then you know you're going to be breaking the aws signature side of things - which is fine for us because that only happens when doing local dev.

service-api/app/src/App/src/Handler/HealthcheckHandler.php
cooperaj: Not sure what happened here. Will fix.

allenannom (approved):

PR #2724, opened 2024-07-24
UML-3510 Set System Message in Admin

Lbagg1:
# Purpose

current behaviour:

Clearing system message and selecting set system message in admin portal is not removing system message from service and seeing error “error updating system message”
Expected behaviour:

Given there is no text in both the English and Welsh message box for Use message
And there is no text in either the English or Welsh message box for View message
And I select set system message
And i will see a flash message “system message has been removed”
Then no message will be in data store

Fixes UML-3510

## Checklist


cooperaj (approved):
This all looks great! 

Get UAT signoff and it's good to go.

PR #2727, opened 2024-07-30
UML-3511 combine twig files and handler functionality.

Lbagg1:
# Purpose

Fixes UML-3511

Description

We currently have two twig files that compose the lpa dashboard; lpa-dashboard.twig.html and lpa-blank-dashboard.twig.html. The content on both is now very similar so they can be merged into a single template file to streamline and reduce redundancy. 

AC

a new template which includes conditional logic to display the correct content based on whether there are LPAs.

updated handler - LpaDashboardHandler

## Checklist


cooperaj (approved):
Looks great.

PR #2903, opened 2024-11-04
UML-3136: Update viewer journey templates to support combined LPAs

acsauk:
# Purpose

So, this was a bit of a slog, but from testing the viewer journey for the current LPA format, combined Sirius, and combined LPA-store LPAs, it looks like it's working. 

I went with new templates to support the combined LPA rather than adding logic to existing templates as this felt a bit cleaner and should hopefully have less cleanup in the future. 

I made an attempt to start the API work that would be required to support this once the mock API response is removed but the PR was already getting huge so I moved it to its [own branch](https://github.com/ministryofjustice/opg-use-an-lpa/tree/UML-3136-extract-api-changes). Feel free to disregard if the API work isn't picked up for a while as it will quickly be outdated. 

I ran out of time before I could add behat tests but it may be worth waiting to implement them when the data model has solidified and the API side has been implemented.

I made a few additions to the front end unit tests such as using some test helpers rather than relying on JSON files to get an LPA instrumented and ready to test. 

Hopefully, this all makes sense and I haven't made any rash assumptions on the format of things but shout up if you want to talk through any of the PR and we can jump on a call.

Fixes UML-3136


service-api/app/src/App/src/Enum/WhenTheLpaCanBeUsed.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App\Enum;
+
+enum WhenTheLpaCanBeUsed: string
+{
+    case WHEN_CAPACITY_LOST                              = 'when-capacity-lost';
```
cooperaj, line 7: Formatting seems fun ;)

service-front/app/src/Common/src/Entity/Address.php
cooperaj, line 1: In theory, though I've not read further, the Address type is only used by the original format lpas (addresses are now part of the Person) so shouldn't need editing?

service-front/app/src/Common/src/Entity/CombinedLpa.php
```diff
@@ -54,4 +60,94 @@ public function jsonSerialize(): mixed
 
         return $data;
     }
+
+    #[DoNotSerialize]
```
cooperaj, line 64: We don't need the DoNotSerialize tags it appears. They're only useful for serialisation using the library and we're not doing the serialisation at all.

service-front/app/src/Common/src/Entity/Person.php
```diff
@@ -28,4 +28,95 @@ public function __construct(
...
+    }
+
+    #[DoNotSerialize]
+    public function getCompanyName(): string|null
+    {
+        return $this->name;
+    }
+
+    #[DoNotSerialize]
+    public function getAddresses(): array
```
cooperaj, line 111: This would be the new functionality that uses Address then.

service-front/app/src/Viewer/templates/viewer/check-code-found-combined-lpa.html.twig
```diff
@@ -0,0 +1,83 @@
...
+                    {{ govuk_error_summary(form) }}
+
+                    {{ govuk_form_open(form) }}
+
+                    <h1 class="govuk-heading-xl">{% trans with {'%firstname%': lpa.donor.firstname, '%surname%': lpa.donor.surname} %}We’ve found %firstname% %surname%'s LPA{% endtrans %}</h1>
+                    <table id="confirmation-table" class="govuk-table">
+                        <tbody class="govuk-table__body">
+                        <tr class="govuk-table__row">
+                            <th class="govuk-table__header" scope="row">{% trans %}Type of LPA{% endtrans %}</th>
+                            <td class="govuk-table__cell">{{ lpa.caseSubtype.propertyAndAffairs ? 'Property and finance' | trans : 'Health and welfare' | trans  }}</td>
```
cooperaj, line 25: Does twig hand the transformation of "propertyAndAffairs" to "isPropertyAndAffairs"?

service-front/app/src/Viewer/templates/viewer/enter-code.html.twig
cooperaj, line 1: Is this not needed?

service-front/app/src/Viewer/templates/viewer/check-code-found-combined-lpa.html.twig
acsauk: Yeah, see https://symfony.com/doc/current/templates.html#template-variables (struggled to find this in the actual twig docs but I'm pretty sure it's a core twig function rather than symfony specific)

service-front/app/src/Common/src/Entity/CombinedLpa.php
acsauk: Ah of course, I was on autopilot from pairing with Allen. I'll remove from all the frontend entities.

service-api/app/src/App/src/Enum/WhenTheLpaCanBeUsed.php
acsauk: Haha I'm so used to autoformat  in Go - I'll tidy up before final review.

service-front/app/src/Viewer/templates/viewer/enter-code.html.twig
acsauk: Yup - turns out it was the first page 😂 not sure how I missed that, reinstated

PR #2984, opened 2024-12-05
UML-3750 added exception for null DOB

allenannom:
# Purpose

_Added an exception to be thrown when the DOB is null. Then implemented a try/catch with a warning log_

Fixes UML-3750

## Approach

appropriately outputs the known error and logs _

## Checklist

### DIFF
--- a/service-api/app/src/App/src/Service/Lpa/FindActorInLpa.php
+++ b/service-api/app/src/App/src/Service/Lpa/FindActorInLpa.php
@@ -8,6 +8,7 @@ use App\Service\Lpa\FindActorInLpa\ActorMatch;
 use App\Service\Lpa\FindActorInLpa\ActorMatchingInterface;
 use App\Service\Lpa\FindActorInLpa\FindActorInLpaInterface;
 use App\Service\Lpa\GetAttorneyStatus\GetAttorneyStatusInterface;
+use Exception;
 use Psr\Log\LoggerInterface;
 use App\Service\Lpa\GetAttorneyStatus\AttorneyStatus;
 
@@ -137,7 +138,22 @@ class FindActorInLpa
 
         $match = self::MATCH;
 
-        $match = $actor->getDob()->format('Y-m-d') !== $matchData['dob']
+        try {
+            $actorDob = $actor->getDob();
+        } catch (Exception $e) {
+            $this->logger->warning(
+                'Actor DOB is null',
+                [
+                    'actor_id'   => $actor->getUid(),
+                    'actor_data' => $actorData,
+                    'error'      => $e->getMessage(),
+                ]
+            );
+
+            return self::NO_MATCH__DOB;
+        }
+
+        $match = $actorDob->format('Y-m-d') !== $matchData['dob']
             ? $match | self::NO_MATCH__DOB
             : $match;
         $match = $actorData['first_names'] !== $matchData['first_names']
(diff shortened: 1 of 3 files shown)


SeemaMenon (approved):

service-api/app/src/App/src/Service/Lpa/SiriusPerson.php
```diff
@@ -80,8 +81,15 @@ public function getPostcode(): string
         return (string)$this->person['addresses'][0]['postcode'];
     }
 
+    /**
+     * @throws Exception
+     */
     public function getDob(): DateTimeInterface
     {
+        if (is_null($this->person['dob'])) {
+            throw new Exception('Actor DOB is not set');
```
cooperaj, line 90: I think a custom exception should be created/thrown for this. Something like `ActorDateOfBirthNotSetException`

service-api/app/src/App/src/Service/Lpa/FindActorInLpa.php
```diff
@@ -137,7 +138,22 @@ private function checkForActorMatch(ActorMatchingInterface $actor, array $matchD
 
         $match = self::MATCH;
 
-        $match = $actor->getDob()->format('Y-m-d') !== $matchData['dob']
+        try {
```
cooperaj, line 141: Rather than have the exception just wrap the problematic call my preference is to have them wrap as much code as is relevant for the operation.

In this case I'd toy with putting the entirety of the rest of the function in it or maybe lines 156 - 167, or for the smallest amount 156-158. Basically avoid having to create new variables to make the exception trapping happen (in this case ActorDob)

cooperaj (commented):
Seems to be missing the test case in the patch I sent that covers an LPA continuing to match on good actors even if a null dob actor is included.

service-api/app/src/App/src/Service/Lpa/FindActorInLpa.php
```diff
@@ -137,7 +138,22 @@ private function checkForActorMatch(ActorMatchingInterface $actor, array $matchD
...
+            $actorDob = $actor->getDob();
+        } catch (Exception $e) {
+            $this->logger->warning(
+                'Actor DOB is null',
+                [
+                    'actor_id'   => $actor->getUid(),
+                    'actor_data' => $actorData,
```
mattmachell, line 148: BY my reading this will log a bunch of PII, so I'd remove it unless that's specifically needed.

service-api/app/src/App/src/Service/Lpa/FindActorInLpa.php
```diff
@@ -126,18 +127,35 @@ private function checkForActorMatch(ActorMatchingInterface $actor, array $matchD
...
+            return self::NO_MATCH__DOB;
+        }
+
         $this->logger->debug(
             'Doing actor data comparison against actor with id {actor_id}',
             [
                 'actor_id'   => $actor->getUid(),
                 'to_match'   => $matchData,
-                'actor_data' => array_merge($actorData, ['dob' => $actor->getDob()]),
+                'actor_data' => array_merge($actorData, [
```
mattmachell, line 150: Please remove this as we don't want PII in logs.

mattmachell (changes requested):

service-api/app/src/App/src/Service/Lpa/FindActorInLpa.php
allenannom: debug is switched off in prod and this was existing logic that was refactored in the combined LPA tickets, specifically, "getDob()".  The updated test case covers the bug issue now, but not sure if this logic here will still be needed going forward though

cooperaj (approved):

mattmachell (dismissed):

cooperaj (approved):

PR #3114, opened 2025-02-04
UML-3736 Event receiver to create db entries

allenannom:
# Purpose

creating DB entries depending on the existence of a user

Fixes UML-3724

## Approach

searches user table to see if the user exists, if not, creates a new entry

## Learning

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service

https://vektra.github.io/mockery/latest/

https://go.dev/doc/tutorial/add-a-test

* [x] I have added relevant and appropriately leveled logging, **without PII**, to my code


lambda-functions/event-receiver/app/makeregister_event_handler.go
```diff
@@ -37,5 +39,62 @@ func (h *MakeRegisterEventHandler) EventHandler(ctx context.Context, record *eve
...
+func handleUsers(ctx context.Context, dynamoClient DynamodbClient, actor Actor) error {
+	// receive data, with data.ActorUID, using dynamo, try and get the row with that id from ActorUsers table and "identity" col
+	// new entry to Actor users of Id (v4 guid) and identity
+
+	var existingUser Actor
+
+	err := dynamoClient.OneByUID(ctx, actor.SubjectID, &existingUser)
+
+	if err != nil {
+		logger.ErrorContext(
```
cooperaj, line 70: If we log it here and log it in the calling function too will it be logged twice?

lambda-functions/event-receiver/internal/random/random.go
```diff
@@ -0,0 +1,31 @@
+package random
+
+import (
+	"crypto/rand"
+
+	"github.com/google/uuid"
+)
+
+func String(length int) string {
```
cooperaj, line 9: Didn't see where these were used. What do we need random strings/numbers for?

lambda-functions/event-receiver/app/makeregister_event_handler.go
gregtyler: If it's that you want to add a prefix to the error and pass it up, you can do something like:
```go
return fmt.Errorf("Failed to find existing user %s: %w", actor.ActorUID, err)
```

cooperaj (approved):

PR #3142, opened 2025-02-20
Uml 3737 add user lpa actor map

allenannom:
# Purpose

_creating DB entries to UserLpaActorMap table depending on the existence of an LPA_

* [x] I have added relevant and appropriately leveled logging, **without PII**, to my code


lambda-functions/event-receiver/app/makeregister_event_handler.go
```diff
@@ -42,7 +44,12 @@ func (h *MakeRegisterEventHandler) EventHandler(ctx context.Context, factory Fac
 	}
 
 	for _, actor := range data.Actors {
-		if err := handleUsers(ctx, factory.DynamoClient(), actor); err != nil {
+		if err := handleUsers(ctx, factory.DynamoClient(), actor, userId); err == nil {
+			err := handleLpas(ctx, factory.DynamoClient(), actor, userId, data.UID)
```
gregtyler, line 48: Probably worth calling `dynamodbClient := factory.DynamoClient()` once before the loop rather than calling it on each iteration.

gregtyler: And this might just be me, but I'd rather put the error-handling logic inside the `if` rather than the success logic. i.e. you'd do something like:
```go
if err := handleUsers(ctx, factory.DynamoClient(), actor, userId); err != nil {
  logger.ErrorContext(...)
  return err
}

if err := handleLpas(ctx, factory.DynamoClient(), actor, userId, data.UID); err != nil {
  return err
}
```

Otherwise it can be hard to follow if something is in error or not

lambda-functions/event-receiver/app/makeregister_event_handler.go
```diff
@@ -79,3 +86,32 @@ func handleUsers(ctx context.Context, dynamoClient DynamodbClient, actor Actor)
...
 	return nil
 }
+
+func handleLpas(ctx context.Context, dynamoClient DynamodbClient, actor Actor, userId string, LpaId string) error {
+	var existingLPA map[string]types.AttributeValue
+
+	err := dynamoClient.GetByLpaIDAndUserID(ctx, LpaId, userId, &existingLPA)
+	if err == nil && existingLPA != nil {
+		return nil
+	}
```
gregtyler, line 96: This logic took me a while to parse. Since I don't think you're doing anything with the existing value, I think it'd be easier if `GetByLpaIDAndUserID` returned a boolean of whether the LPA was found.
```suggestion
	exists, err := dynamoClient.GetByLpaIDAndUserID(ctx, LpaId, userId)
	if err == nil && exists {
		return nil
	}
```
At which point you could also rename it to something like `ExistsLpaIDAndUserID`

lambda-functions/event-receiver/app/makeregister_event_handler_test.go
```diff
@@ -168,10 +176,119 @@ func TestHandleCloudWatchEvent_FailedToPutActor(t *testing.T) {
...
+
+	actor := Actor{
+		ActorUID:  "9ac5cb7c-fc75-40c7-8e53-059f36dbbe3d",
+		SubjectID: "urn:fdc:gov.uk:2022:XXXX-XXXXXX",
+	}
+
+	err := handleLpas(ctx, mockDynamo, actor, userId, lpaId)
+
+	assert.NoError(t, err)
+	assert.Nil(t, err)
```
gregtyler, line 266: You don't need both of these

lambda-functions/event-receiver/internal/dynamo/client.go
```diff
@@ -88,3 +88,27 @@ func (c *Client) Put(ctx context.Context, v interface{}) error {
...
+			":userid": &types.AttributeValueMemberS{Value: userId},
+		},
+		KeyConditionExpression: aws.String("#LpaUid = :lpaUid AND #UserId = :userid"),
+		Limit:                  aws.Int32(1),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to query LPA mappings: %w", err)
+	}
+	if len(response.Items) == 0 {
+		return NotFoundError{}
```
gregtyler, line 110: Should this not return `nil`?

lambda-functions/event-receiver/app/makeregister_event_handler.go
gregtyler: Oh, also, I think `existingLpa` will never be `nil`. If it's not populated, it will be a zero-valued struct instead.

lambda-functions/event-receiver/app/makeregister_event_handler_test.go
```diff
@@ -78,7 +81,10 @@ func TestMakeRegisterEventHandler_Success(t *testing.T) {
 		user := args.Get(2).(*Actor)
 		*user = existingUser
 	}).Return(nil)
-	mockDynamo.On("Put", ctx, mock.Anything).Return(nil)
+	mockDynamo.On("Put", ctx, mock.Anything, mock.Anything).Return(nil)
```
gregtyler, line 84: You should be able check the values passed to the Put function, though you'll need to add `mockFactory.On("UuidString").Return("mock-uuid")`
```suggestion
	mockDynamo.On("Put", ctx, "ActorUsers", map[string]types.AttributeValue{
		"Id":       &types.AttributeValueMemberS{Value: "mock-uuid"},
		"Identity": &types.AttributeValueMemberS{Value: "urn:fdc:gov.uk:2022:XXXX-XXXXXX"},
	}).Return(nil)
```

nickdavis2001 (dismissed):

docker-compose.dependencies.yml
```diff
@@ -163,6 +163,9 @@ services:
     environment:
       - DEBUG=1
       - DOCKER_HOST=unix:///var/run/docker.sock
+      - AWS_ENDPOINT_DYNAMODB=http://dynamodb-local:8000
+      #- MAIN_DOCKER_NETWORK=default
```
cooperaj, line 167:
```suggestion
```

cooperaj (dismissed):

cooperaj (approved):

nickdavis2001 (approved):

PR #3197, opened 2025-03-17
Uml 3785 refactor add lpa to work with combined format

allenannom:
# Purpose
Ensure compatibility of variable types passed between functions 

### DIFF
--- a/service-api/app/src/App/src/Service/ActorCodes/Validation/CodesApiValidationStrategyInterface.php
+++ b/service-api/app/src/App/src/Service/ActorCodes/Validation/CodesApiValidationStrategyInterface.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Service\ActorCodes\Validation;
+
+use DateTimeImmutable;
+
+interface CodesApiValidationStrategyInterface
+{
+    public function getDob(): DateTimeImmutable;
+}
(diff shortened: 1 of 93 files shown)


service-api/app/src/App/src/Service/ActorCodes/Validation/CodesApiValidationStrategyInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Service\ActorCodes\Validation;
+
+use DateTimeImmutable;
+
+interface CodesApiValidationStrategyInterface
```
cooperaj, line 9: Recommend calling this CodesApiValidationInterface as any usage of it with alternative strategies won't be affected and it's more succinct.

service-front/app/src/Common/src/Entity/CaseActor.php
```diff
@@ -166,6 +166,6 @@ public function getSystemStatus(): bool
 
     public function setSystemStatus(bool $systemStatus): void
     {
-        $this->systemStatus = $systemStatus;
+        $this->systemStatus = $systemStatus ?? false;
```
cooperaj, line 169: The local system status should never be null since it's typed in the parameter?

cooperaj (dismissed):

cooperaj (approved):

PR #3229, opened 2025-04-07
Surfaced restrictionsAndConditions field in combined format lpa

allenannom:
# Purpose

_exposed restrictionsAndConditions field in the combined lpa formats with the goal of displaying it on the front end_


service-front/app/test/CommonTest/Entity/CanHydrateCombinedLpaTest.php
```diff
@@ -124,16 +126,26 @@ public function expectedSiriusLpa(): CombinedLpa
         ];
 
         return EntityTestHelper::makeCombinedLpa(
+            applicationHasGuidance:    false,
```
cooperaj, line 129: Lots of new things in this file that I don't recognise as properties related to this change?

allenannom: yeah i added them because the test was failing as it complained that it didnt have the required properties, but now its working without it there..

PR #3241, opened 2025-04-17
Update C4 diagram. Remove the structurizer folders 

mattmachell:
Update the C4 digram for Use to include recent additions.


docs/diagrams/use-an-lpa-c4-l2.svg
cooperaj, line 1: We're missing a link between view front end and pdf service. The use front end does not do PDFs.

Is there a case for having S3 hanging off the I&P service? (and Sirius).

mattmachell: Ah, good catch on the PDF service! Will update.

I'd kept the Sirius integrations more abstracted as the diagram is primarily about Use. i.e. Use only cares it has a service to call that gives back I&P. When I update the diagrams in the integrations I'll add that level of detail with S3.

cooperaj (approved):

PR #3275, opened 2025-05-16
paper verification code - LPA number reference

allenannom:
# Purpose

_added a new page to the workflow for paper verification_

## Approach

_New page
New handler
Content addition in both languages_


service-front/app/src/Viewer/src/Form/LpaCheck.php
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public function getInputFilterSpecification(): array
+    {
+        return [
+            'lpa_reference' => [
+                'required'   => true,
+                'filters'    => [
+                    ['name' => StringTrim::class],
+                ],
+                'validators' => [
```
cooperaj, line 49: If we're saying that code is an M followed by 12 numbers then I think these rules will need looking at. A user may enter:

- M123412341234
- M-1234-1234-1234
- M-123412341234

and these would all fail the string length and regex validators.

cooperaj (approved):

PR #3282, opened 2025-05-20
UML 3862 allow orgs to enter attorneys date of birth PV

allenannom:
# Purpose

_a new page to allow orgs to enter attorneys date of birth PV_

Fixes UML-####

## Approach

_new page with handler _


service-front/app/config/routes.php
```diff
@@ -59,6 +60,11 @@
         Common\Handler\InstructionsPreferencesBefore2016Handler::class,
         'lpa.instructions-preferences-before-2016'
     );
+
+    $app->route('/attorney-dob',
```
cooperaj, line 64: Not a blocker but I think we probably want to move all the PV code flow pages into a subfolder path. In this case something like '/paper-verification/attorney-dob'

PR #3297, opened 2025-06-11
Uml 3868 donor flow paper verification journey check answers page

allenannom:
# Purpose

_check answers page with option to edit each answer_


service-front/app/src/Viewer/src/Handler/PaperVerification/AttorneyDateOfBirthHandler.php
```diff
@@ -58,7 +59,13 @@ public function handlePost(ServerRequestInterface $request): ResponseInterface
         $this->form->setData($request->getParsedBody());
 
         if ($this->form->isValid()) {
-            $this->state($request)->dateOfBirth = $this->form->getData()['pv_date_of_birth'];
+            $postData = $this->form->getData();
+
+            $this->state($request)->dateOfBirth = (new DateTimeImmutable())->setDate(
```
cooperaj, line 64: I think (needs verification) that this would result in a datetime that has a time set to the time of creation - rather than 00:00:00

service-front/app/src/Viewer/src/Handler/PaperVerification/AttorneyDateOfBirthHandler.php
```diff
@@ -58,7 +59,13 @@ public function handlePost(ServerRequestInterface $request): ResponseInterface
...
+                (int) $postData['dob']['year'],
+                (int) $postData['dob']['month'],
+                (int) $postData['dob']['day']
+            );
             return $this->redirectToRoute($this->nextPage($this->state($request)));
```
cooperaj, line 69: I don't believe next page is hooked up right. Just making a note for the future.

service-front/app/src/Viewer/src/Handler/ProvideAttorneyDetailsForPVHandler.php
```diff
@@ -52,11 +52,11 @@ public function handlePost(ServerRequestInterface $request): ResponseInterface
         $this->form->setData($request->getParsedBody());
 
         if ($this->form->isValid()) {
-            $this->session->set('noAttorneys', $this->form->getData()['no_of_attorneys']);
-            $this->session->set('attorneysName', $this->form->getData()['attorneys_name']);
+            $this->session->set('noOfAttorneys', $this->form->getData()['no_of_attorneys']);
```
cooperaj, line 55: As we're deep in the workflow at this point none of this needs to go in the session. That was only neccessary for the first page since the code works for both sirius and modernise lpas.

tests/features/download-pdf.feature
```diff
@@ -4,7 +4,7 @@ Feature: Retain a copy of the shared LPA summary
   I can download a summary of a LPA
   So that I can retain a copy of the data as it was at the time I accessed it
 
-  @smoke @ff:paper_verification:from_env
+  @smoke @ff:paper_verification:false
```
cooperaj, line 7: These should work in all cases? 

If the summary is broken for the flag being on then it's likely something we can ignore for now but if we have this in here now we won't be testing the pdf download properly when the flag goes live.

PR #3400, opened 2025-09-25
UML-4022 - decommission old network vpc endpoints etc

andrewpearce-digital:
# Purpose

Remove old network resources and logic for network migration

Fixes UML-4022

## Approach

account
- delete old network resources (default vpc may need to be manually removed from state or deleted in the console)
- delete vpc flow logs for old network and associated iam resources 
- delete elsaticache subnet group for old network
- delete network migration logic from elasticache

environment
- remove network migration logic from ecs and load balancer resources
- remove data sources for for old network


andrewpearce-digital:
> We should keep the VPC flow logs for a while after migration rather than just delete.

good point, I'll remove them from state

PR #3680, opened 2026-04-14
Dockerise Psalm steps in CI

gregtyler:
# Purpose

We want to dockerise our workflows to ensure version consistency, and to avoid running PHP on the GHA runners directly, whilst ensuring that failures will fail the pipeline and annotate the PR.

<details>
<summary>Example output</summary>

image

I'm yet to figure out why these show up twice: it's an issue I've got on other repos too.

</details>

Fixes UML-4343 #patch

## Approach

Save docker dev images as artifacts in Docker stages, then load in a Psalm workflow and run Psalm inside the image.

Call the psalm_analysis workflow in the main PR and path-to-live workflows after Docker images have been built.

I had to stop deleting the vendor folder because otherwise the image isn't usable as a stand-alone.

* [ ] I have added relevant and appropriately leveled logging, **without PII**, to my code
* [ ] I have updated documentation (Confluence/GitHub wiki/tech debt doc)


service-api/docker/web/Dockerfile
```diff
@@ -3,7 +3,7 @@ USER root
 
 COPY service-api/docker/web/default.conf.template /etc/nginx/templates/default.conf.template
 
-RUN apk --no-cache add libcap=2.77-r0 && \
+RUN apk --no-cache add libcap=2.78-r0 && \
```
gregtyler, line 6: I couldn't build without this change: I believe the Alpine has updated which versions of libcap are available in apk.

service-api/docker/app/Dockerfile
```diff
@@ -56,8 +61,7 @@ RUN set -xe \
   && install-php-extensions xdebug-stable ffi \
   && echo "ffi.enable=true" >> /usr/local/etc/php/conf.d/docker-php-ext-ffi.ini
 
-# Development images are only ever run using compose so will have this mounted in as a volume
-RUN rm -rf /app/vendor
+RUN composer check-platform-reqs
```
cooperaj: Just wondering why this little dance when you could just run the composer install instead of the check-platform-reqs?

service-api/docker/web/Dockerfile
cooperaj: Yup, downside of pinning dependencies on a container OS that only ever has one version in it's repos.

service-api/docker/app/Dockerfile
cooperaj: The snippet doesn't quite cover what I selected–which was all of it. I don't think there is a need for a "dependency-dev" step if it's then only used in the development image as a copy operation.

gregtyler: Yep, excellent point, I went through a few iterations and lost sight of how much easier that would be 😁 I've simplified those down now.

.github/workflows/psalm-static-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Run psalm
+        run: |
+          mkdir -m 0777 ./output
+          docker run --rm --volume ./output:/tmp/output $IMAGE:latest \
+          vendor/bin/psalm --report=/tmp/output/results.sarif --output-format=github --long-progress
+
+      - name: Run psalm taint analysis
+        run: |
+          docker run --rm --volume ./output:/tmp/output $IMAGE:latest \
+          vendor/bin/psalm --output-format=github --taint-analysis --report=/tmp/output/results.sarif --long-progress --ignore-baseline
```
cooperaj: Will these two output operations result in a merge of the results.sarif file or does the second overwrite?

service-front/docker/app/Dockerfile
```diff
@@ -53,8 +53,8 @@ COPY service-front/docker/app/app-php-development.ini /usr/local/etc/php/conf.d/
 RUN set -xe \
   && install-php-extensions xdebug-stable
 
-# Development images are only ever run using compose so will have this mounted in as a volume
-RUN rm -rf /app/vendor
+RUN composer install --prefer-dist --no-interaction --no-scripts --optimize-autoloader --ignore-platform-reqs && \
```
cooperaj, line 56: Why is the --ignore-platform-reqs passed and then an immediate check-platform-reqs run?

TBH I don't know why we run the check-platform-reqs command in the dependency step as the install would fail if those hadn't been met (no ignore flag)

.github/workflows/psalm-static-analysis.yml
gregtyler: Good catch, I've matrix-ified these so they can be run in parallel and produce/upload their individual sarif files.

service-front/docker/app/Dockerfile
gregtyler: `composer install` always uses what's in the config.platform section of composer.json ([example](https://github.com/ministryofjustice/opg-use-an-lpa/blob/27963156d143930f59fa304f90a102813cd10ede/service-api/app/composer.json#L15-L17)), to allow you to install dependencies in a different environment to your image (i.e. it doesn't care what `php -v` is, or even if PHP is installed). So it's necessary to run `check-platform-reqs` on the runtime image to make sure it has the right version of PHP and all the necessary extensions.

This could be avoided by removing `config.platform`, but then other tools (like Dependabot/Renovate) will always assume you're running on latest and try to install packages that require later versions of PHP.

cooperaj: ah ok. so why do we ignore the req for dev but not production?

cooperaj (approved):

JoshMurray99 (approved):

PR #3681, opened 2026-04-15
Fix end of workflow logic

JoshMurray99:


gregtyler (dismissed):

.github/workflows/pull-request-path.yml
```diff
@@ -251,7 +251,6 @@ jobs:
       - resolve_docker_tag
       - terraform_lint
       - workflow_variables
-      - psalm_analysis
```
gregtyler, line 254: Does removing this mean that it will try to apply the environment even if there are Psalm errors?

JoshMurray99: yeah the reason I removed it was that if there were no app code changes, the psalm job wouldn't run, so tf apply environment wouldn't run and would fail the workflow. I can put a check like this instead : (needs.psalm_analysis.result == 'success' || needs.workflow_variables.outputs.detect_app_code_changes == 'false')

gregtyler (approved):

PR #3696, opened 2026-04-20
UML-3336 Show different page when activation key already made today

hawx:


.github/workflows/php-linting.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       if: env.CHANGED_PHP_FILES != ''
       run: |
         docker run --rm --volume ./output:/tmp/output $IMAGE:latest \
-        vendor/bin/phpcs --standard=phpcs.xml.dist --report=full --report-xml=/tmp/output/phpcs-report.xml \
+        vendor/bin/phpcs -n --standard=phpcs.xml.dist --report=full --report-xml=/tmp/output/phpcs-report.xml \
```
hawx, line 62: warnings are ignored on the pre-commit hook, so ignoring here too means no surprises

cooperaj: I'd argue that this is intended. We don't want to stop developer commits for work in progress things, but we should be stopping that code from going live if it doesn't meet the psalm standards we have in place.

PHPStorm would have been highlighting all these issues during development as well. I'm guessing it should be possible to setup the psalm language server in emacs to do similar.

cooperaj (approved):

.github/workflows/php-linting.yml
hawx: it should be possible for me to set that up, but I don't really know if I care about a line of 130 chars or a function with complexity of 12. I do care about a key not being defined in an array. If we do want to enforce those things then I'd rather they go from being warnings to being errors

jay-whitwell (approved):

PR #3701, opened 2026-04-24
UML-2852 Normalise names when checking equality

hawx:
Other options I could probably be convinced to swap to:
- make these non-static so the class can be injected (I didn't because I am currently in a "mocked unit testing is just writing code twice" phase)
- create separate classes for each thing as a value type with an `equals` method to use (I didn't because I thought it wouldn't be as obvious in the code; also there is no way to overload `==`, right?)
- remove the class and use with `use function` (I didn't because then I'd have to figure out what to do with the helper functions that are currently private)
- something else I haven't thought of...

### DIFF
--- a/service-api/app/src/App/src/Service/Equals.php
+++ b/service-api/app/src/App/src/Service/Equals.php
@@ -1,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Service;
+
+final class Equals
+{
+    public static function firstNames(string $a, string $b): bool
+    {
+        return self::normaliseFirstNames($a) === self::normaliseFirstNames($b);
+    }
+
+    private static function normaliseFirstNames(string $s): string
+    {
+        // only take the first of the firstnames for comparison
+        return self::turnUnicodeCharToAscii(strtolower(explode(' ', trim($s))[0]));
+    }
+
+    public static function lastName(string $a, string $b): bool
+    {
+        return self::normaliseLastName($a) === self::normaliseLastName($b);
+    }
+
+    private static function normaliseLastName(string $s): string
+    {
+        return self::turnUnicodeCharToAscii(preg_replace('/\s+/', ' ', strtolower(trim($s))));
+    }
+
+    public static function postcode(string $a, string $b): bool
+    {
+        return self::normalisePostcode($a) === self::normalisePostcode($b);
+    }
+
+    private static function normalisePostcode(string $s): string
+    {
+        return strtolower(str_replace(' ', '', $s));
+    }
+
+    private static function turnUnicodeCharToAscii(string $s): string
+    {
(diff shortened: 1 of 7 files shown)


service-api/app/src/App/src/Service/Equals.php
```diff
@@ -0,0 +1,46 @@
...
+
+    private static function normalisePostcode(string $s): string
+    {
+        return strtolower(str_replace(' ', '', $s));
+    }
+
+    private static function turnUnicodeCharToAscii(string $s): string
+    {
+        $s = str_replace(['‘', '’'], '\'', $s);
+        $s = str_replace(['‐', '‑', '‑', '‑', '‑'], '-', $s);
```
cooperaj: Probably just the way that github renders these but they all look the same?

hawx: they do, I was tempted to use something like `mb_chr` so they'd be more obvious. But then you have to look up the numbers to see what is what 🤷

hawx: actually, why don't I just write a comment next to them

service-api/app/src/App/src/Service/Equals.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return self::normalisePostcode($a) === self::normalisePostcode($b);
+    }
+
+    private static function normalisePostcode(string $s): string
+    {
+        return strtolower(str_replace(' ', '', $s));
+    }
+
+    private static function turnUnicodeCharToAscii(string $s): string
```
cooperaj, line 40: Potentially there is https://www.php.net/manual/en/class.transliterator.php to aid in this. This code is simple though which has that going for it.

cooperaj (dismissed):

service-api/app/src/App/src/Service/Equals.php
cooperaj: The use of a transliterator in names would also have the (ad|dis)advantage of removing all diacritics or accents as well. Might be worth asking Sarah/Jude about that direction though.

service-api/app/src/App/src/Service/Equals.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $s = str_replace(['‘', '’'], '\'', $s);
+        $s = str_replace([
```
cooperaj, line 43: Yeah I like this.

cooperaj (dismissed):

service-api/app/src/App/src/Service/Equals.php
cooperaj: i.e. If the Sirius scanning process is in the habit of removing or mangling them then maybe it's a good idea to just discount them when matching.

cooperaj (approved):

PR #3707, opened 2026-05-01
UML-4386: Request for activation key journey logic slightly re-ordered to meet the already have sent activation key criteria

SeemaMenon:
# Purpose

Fixes UML-4386

## Approach

## Learning

* [ ] I have added relevant and appropriately leveled logging, **without PII**, to my code
* [ ] I have updated documentation (Confluence/GitHub wiki/tech debt doc)


service-api/app/src/App/src/Service/Lpa/AccessForAll/AddAccessForAllLpa.php
```diff
@@ -70,6 +70,28 @@ private function processActivationCode(
         AccessForAllValidation $validationData,
         ?array $lpaAddedData,
     ): void {
+        $hasActivationCode = $this->accessForAllLpaService->hasActivationCode(
```
hawx, line 73: Does this mean the other "already requested activation key today" and "already requested activation key" pages won't show? If that is what we want maybe we should clean those up

hawx: I feel like this was all originally done under the assumption that sending the letter might take some time from the original request, but actually it happens (almost) instantly

SeemaMenon: Nope..... Those pages are still needed.

So first it would check for if an activation key has already been sent and is unused [by checking has code and created date]. And the page shows 'An activation key' has already been sent'. Irrespective of that the user could still request for a key and then the 'Activation key has been sent' page will be shown and if they continue to request a key again today or in near future then the  other "already requested activation key today" and "already requested activation key" would shown respectively.

So its just that the order of check had to shuffled to meet the below criteria
Screenshot 2026-04-30 at 00 06 18

SeemaMenon:
Screenshot 2025-01-13 at 13 19 35

service-api/app/src/App/src/Service/Lpa/AccessForAll/AddAccessForAllLpa.php
cooperaj: Yeah, I think this needs a rethink.

Maybe this whole process.

The original flow checked our local db first to see if a key had been requested using the service - and if it had provides the user with an estimated delivery date. If there was not a local record then it checked the code service and if it found a key there this indicates the user was given (or requested) a key *outside* the service - like in the initial LPA registration or from the call centre. Notice this does not attach the date and so the frontend would tell the user they already had a key but could request one.

By changing the order of the operations here the user would only see the first page *if* the key that had been requested had not yet been processed by a case worker - and since the majority of our key requests will succeed automatically it means they would never see the page with an estimated delivery date, they would always be told that they asked for a key, not that they had been sent one.

We can probably simplify all of this but will need to get Jude and Sarah together to plan a better flow.

SeemaMenon:
THERE IS A FOLLOW UP TICKET ON THIS ISSUE TO REVIEW THE WHOLE JOURNEY.

PR #3718, opened 2026-05-11
UML-4366 Don't wrap lines in .po files

hawx:
Poedit disagrees with Weblate on how to wrap lines. So just going to not wrap for simplicity. This also adds in the `en_GB` language to show that it works.

image


cooperaj (approved):
Guessing thats weblate not following the normal way of doing things. Sigh

PR #3720, opened 2026-05-12
Update urllib3 to 2.7.0

gregtyler:
# Purpose

v2.6.3 has a high-severity vulnerability

## Approach

By running `uv pip compile (...) --upgrade-package urllib3`

## Learning

Dependabot runs uv inside its directory, so changes all the "generated by command" comments. This doesn't really matter, but I chose a manual approach to avoid creating loads of separate PRs.


hawx:
> Dependabot runs uv inside its directory, so changes all the "generated by command" comments. 

My fault for using a naive bash loop to do the initial creation.

hawx (approved):

PR #3776, opened 2026-06-22
UML-4447 - Give data access role kms decrypt permissions on dynamodb …

JoshMurray99:


terraform/account/kms.tf
```diff
@@ -242,7 +242,8 @@ module "dynamodb_encryption_key" {
       "-opg-use-an-lpa-ci",
       "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/breakglass",
       "-admin-task-role",
-      "arn:aws:iam::${data.aws_caller_identity.backup.account_id}:role/aws-service-role/backup.amazonaws.com/AWSServiceRoleForBackup"
+      "arn:aws:iam::${data.aws_caller_identity.backup.account_id}:role/aws-service-role/backup.amazonaws.com/AWSServiceRoleForBackup",
+      "data-access"
```
andrewpearce-digital: this can be the full arn for the role. The other entries have the environment name as part of the arn so we only use part of the name.

andrewpearce-digital (approved):

PR #3787, opened 2026-06-26
UML-4460 - Make UV lock files for the remaining standalone scripts

JoshMurray99:


mock-integrations/opg-data-lpa/generate_examples/make_examples.sh
```diff
@@ -5,11 +5,15 @@ rm -f temp.yaml
 touch temp.yaml
 cat examples-template.yaml >> temp.yaml
 cp nginx-template.conf nginx.conf
+
+# Setup dependencies
+uv sync --locked --directory ../../../scripts/call-api-gateway
```
hawx: `call-api-gateway` is only used here, so you could move it to make this less `..`y

hawx (dismissed):

hawx (approved):

PR #3788, opened 2026-06-26
Remove extra empty log group created by lambda module

hawx:
The lambda module will currently create two log groups:

- `{env}-{lambda_name}` - this one is unused and empty
- `{lambda_name}-{env}` - this one is used

This PR stops creating the unused one.

I imagine this will complain when merged to production, and these log groups will have to be removed manually...


hawx:
looks like we want to keep this log group, and instead need to log to it instead of the default AWS one that is provided

PR #3829, opened 2026-07-23
UML-4489 - Security Hub- SSM.7 - SSM documents should have the block …

JoshMurray99:
…public sharing setting enabled

### DIFF
--- a/terraform/account/region/securityhub.tf
+++ b/terraform/account/region/securityhub.tf
@@ -1,3 +1,12 @@
 resource "aws_ebs_snapshot_block_public_access" "this" {
   state = "block-all-sharing"
+
+  provider = aws.region
+}
+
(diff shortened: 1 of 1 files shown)


terraform/account/region/ssm.tf
```diff
@@ -0,0 +1,4 @@
+resource "aws_ssm_service_setting" "public_sharing_permission" {
+  setting_id    = "arn:aws:ssm:${data.aws_region.current.region}:${var.account.account_id}:servicesetting/ssm/documents/console/public-sharing-permission"
+  setting_value = "Disable"
+}
```
andrewpearce-digital: missing a provider. This will only show as a problem on apply

andrewpearce-digital (approved):

PR #3834, opened 2026-07-24
Add a new slack output for translations

cooperaj:
# Purpose

Add a new type of slack output for translations changes and show it instead of the normal production output if they're detected.

* I have added relevant and appropriately leveled logging, **without PII**, to my code
* I have updated documentation (Confluence/GitHub wiki/tech debt doc)

### DIFF
--- a/.github/workflows/path-to-live.yml
+++ b/.github/workflows/path-to-live.yml
@@ -281,20 +281,6 @@ jobs:
       webhook: ${{ secrets.PROD_SLACK_WEB_HOOK }}
     if: needs.workflow_variables.outputs.specific_path == 'docs'
 
-  slack_notify_translation_update:
-    name: notify translations update
-    uses: ./.github/workflows/_slack-notification.yml
-    needs:
-      - production_health_check
-      - workflow_variables
-    with:
-      template: translations_update.txt
-      translations_update: ${{ needs.workflow_variables.outputs.translations_update }}
-      workspace: production
-    secrets:
-      webhook: ${{ secrets.PROD_SLACK_WEB_HOOK }}
-    if: needs.workflow_variables.outputs.translations_update == 'true'
-
   slack_notify:
     name: notify of result
     uses: ./.github/workflows/_slack-notification.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/path-to-live.yml
```diff
@@ -57,6 +58,17 @@ jobs:
...
+      - name: get changed translation files in any folder
+        id: changed-files-translations
+        uses: ministryofjustice/opg-github-actions/actions/detect-changes@2ad72748485e27f590599a7e9768b9fc0d5b8e03 # v4.8.7
+        with:
+          paths: |
+            **/*.po
+            **/*.mo
+      - name: extract variables for workflow
+        id: translations
+        run: echo "updated=${{ steps.changed-files-translations.outputs.has_changes }}" >> $GITHUB_OUTPUT
```
JoshMurray99: You can remove this step and just put this for the output on line 38: `translations_update: ${{ steps.changed-files-translations.outputs.has_changes }}`

.github/workflows/path-to-live.yml
```diff
@@ -275,8 +287,14 @@ jobs:
...
       - slack_notify_documentation_update
+      - slack_notify_translation_update
+      - workflow_variables
     with:
-      template: production_release.txt
+      template: |-
+        ${{ case(
+          needs.workflow_variables.outputs.translations_update == 'true', 'translations_update.txt',
+          'production_release.txt'
+        ) }}
```
JoshMurray99, line 293: I don't think case is valid syntax, something like this would work: 
`template: ${{ needs.workflow_variables.outputs.translations_update == 'true' && 'translations_update.txt' || 'production_release.txt' }}`

.github/workflows/path-to-live.yml
```diff
@@ -275,8 +287,14 @@ jobs:
     needs:
       - production_health_check
       - slack_notify_documentation_update
+      - slack_notify_translation_update
```
JoshMurray99: I don't think this job exists?

.github/workflows/path-to-live.yml
cooperaj: Took it straight from the docs so should be?

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case

JoshMurray99: oh nice! Haven't seen that before

.github/workflows/path-to-live.yml
cooperaj: Yeah. I'd done it differently first before doing this much simpler approach

.github/workflows/path-to-live.yml
cooperaj: Seems like a nicer way to do things then ternaries - especially if the list might grow.

JoshMurray99 (approved):

PR #3848, opened 2026-07-29
UML-4459 - Add new service-area tag

JoshMurray99:
Add the new POAS service-area tag.
Terraform wanted to destroy and recreate a lot of resources due to unknown values at apply time on certain data sources, so using variables instead for these data sources including in the firewalled network module.

### DIFF
--- a/terraform/account/region/modules/dns_firewall/variables.tf
+++ b/terraform/account/region/modules/dns_firewall/variables.tf
@@ -3,6 +3,11 @@ variable "kms_key_arn" {
   type        = string
 }
 
+variable "default_tags" {
+  description = "Default tags to apply to all resources"
+  type        = map(string)
+}
+
 variable "domains_allowed" {
   description = "The domains allowed to be resolved by the DNS server."
   type        = list(string)
@@ -21,5 +26,10 @@ variable "enable_block" {
 
 variable "brute_force_cache_primary_endpoint_address" {
   description = "The primary endpoint address of the Brute Force Cache."
+  type        = string
+}
+
+variable "region" {
+  description = "The aws region"
   type        = string
 }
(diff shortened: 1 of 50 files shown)


terraform/environment/region/variables.tf
```diff
@@ -287,3 +287,13 @@ variable "lpa_store_account_id" {
   description = "The AWS ID of the LPA store account"
   type        = string
 }
+
+variable "region" {
```
andrewpearce-digital: be careful not to call this region as there is the region attribute tht controls where the resource is created as of v6 of the provider

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/enhanced-region-support

terraform/account/region/modules/dns_firewall/variables.tf
```diff
@@ -3,6 +3,23 @@ variable "kms_key_arn" {
   type        = string
 }
 
+# variable "default_tags" {
```
andrewpearce-digital: needs cleanup

terraform/environment/locals.tf
```diff
@@ -3,13 +3,15 @@ locals {
   environment       = contains(keys(var.environments), local.environment_name) ? var.environments[local.environment_name] : var.environments["default"]
   dns_namespace_env = local.environment.account_name == "production" ? "" : "${local.environment_name}."
   capacity_provider = local.environment.fargate_spot ? "FARGATE_SPOT" : "FARGATE"
+  region            = data.aws_region.current.region
```
andrewpearce-digital, line 6: is this local.region still needed?

JoshMurray99: Yeah it's used in backups.tf and lambda.tf

PR #3871, opened 2026-08-07
Uml 4438 Fix up CSP and ensure we can move forward with enforcement

cooperaj:
# Purpose

Fix up reported issues in the logs so that we can turn on the enforcement mode of CSP

Fixes UML-4438

## Approach

Had to completely rework how we were doing it as it's not possible to work with form-action in Chrome in a OIDC world without making the CSP at least a little dynamic. That means we couldn't do it as a static nginx template and need to calculate the correct header based on what we're going to allow the user to do.

Implement that as a configurable middleware.

* I have added relevant and appropriately leveled logging, **without PII**, to my code
* I have updated documentation (Confluence/GitHub wiki/tech debt doc)


service-front/app/src/Common/src/Middleware/Security/CSPMiddleware.php
```diff
@@ -0,0 +1,65 @@
...
+        $csp .= "style-src    'self' 'nonce-" . $this->nonce . "';";
+        $csp .= "font-src     'self';";
+        $csp .= "manifest-src 'self';";
+        $csp .= "connect-src  'self' " .
+            'https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com;';
+
+        $iapSrc = '';
+        if (
+            $request->getAttribute(RouteResult::class)->getMatchedRouteName() === 'lpa.view'
+            || $request->getAttribute(RouteResult::class)->getMatchedRouteName() === 'view-lpa'
```
cooperaj, line 42: Rather than match on both route names this should be more intelligent about the application context (actor or viewer) and do the right thing.

service-front/app/src/Common/src/Middleware/Security/CSPMiddleware.php
```diff
@@ -0,0 +1,65 @@
...
+        ) {
+            $iapSrc = 'data: ' . $this->iapDomain;
+        }
+        $csp .= "img-src 'self' https://*.google-analytics.com https://*.googletagmanager.com " . $iapSrc . ';';
+
+        // instead of setting form-action to null to work around inconsistant Chrome/Firefox implementations
+        // lets just work out if we need to add the OIDC domain for this request.
+        $authenticationSrc = '';
+        if ($request->getAttribute(RouteResult::class)->getMatchedRouteName() === 'home') {
```
cooperaj, line 51: By the same token we don't want to really attach a form-action value for the 'home' route of viewer as it's not necessary.

gregtyler (approved):

